PR #29, opened 2021-01-25
[Annotation] Automatically add layer and source while changing style.

Chaoba:
QQ20210125-205122


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -94,6 +95,12 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     gesturesPlugin.addOnMapClickListener(mapClickResolver)
     gesturesPlugin.addOnMapLongClickListener(mapLongClickResolver)
     gesturesPlugin.addOnMoveListener(mapMoveResolver)
+    delegateProvider.mapListenerDelegate.addOnDidFinishRenderingMapListener {
```
pengdev, line 98: @Chaoba Instead of OnDidFinishRenderingMapListener, we could use MapStyleObserverPlugin on the AnnotationPlugin, and propagate the style changed event to the underlying annotation managers to re-add the layers/sources.

pengdev: ticketed at https://github.com/mapbox/mapbox-maps-android/issues/146

PR #45, opened 2021-01-27
Add 3D terrain with animated directions route 

tobrun:
This PR adds an example of an animated route on 3D terrain:
 - route geometry is fetched from the Mapbox directions API and visualize with a LineLayer
 - Animate a LocationLayer to visualize the end user location
 - Animate a FreeCameraOptions across the route 

ezgif com-video-to-gif(24)


app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+        )
+        +locationIndicatorLayer(LOCATION_LAYER_ID) {
+          topImage(FOREGROUND_ICON)
+          bearingImage(BACKGROUND_ICON)
+          imagePitchDisplacement(5.0)
+          topImageSize(1.5)
+          bearingImageSize(1.5)
+        }
+        +image(FOREGROUND_ICON) {
+          bitmap(BitmapFactory.decodeResource(resources, R.drawable.mapbox_mylocation_icon_default))
```
pengdev, line 89: nit: could use the generateBitmap function as well.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+      .accessToken(getString(R.string.mapbox_access_token))
+      .build()
+    client.enqueueCall(object : Callback<DirectionsResponse> {
+      override fun onResponse(
+        call: Call<DirectionsResponse>,
+        response: Response<DirectionsResponse>
+      ) {
+        animateRoute(
+          LineString.fromPolyline(
+            response.body()!!.routes()[0].geometry()!!,
```
pengdev, line 140: nit: we could avoid using `!!` to handle errors more gracefully.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+      LOCATION_LAYER_ID
+    )
+
+    // cache the camera
+    val camera = mapboxMap.getFreeCameraOptions()
+
+    // get a reference to elevations of the animation
+    // note: to call getElevation the coordinate needs to be in the viewport
+    val startElevation = mapboxMap.getElevation(POINT_START)!!.toDouble()
+    val endElevation = mapboxMap.getElevation(POINT_END)!!.toDouble()
```
pengdev, line 181: nit: could avoid using `!!` and handle the `null` gracefully by providing fallback values.

pengdev (commented):
Great example! LGTM with a few nits.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+
+    // load satellite style and add terrain with a line layer to visualize the route
+    mapboxMap.loadStyle(
+      style(styleUri = Style.SATELLITE_STREETS) {
+        +rasterDemSource(SOURCE) {
+          url(TERRAIN_URL_TILE_RESOURCE)
+          tileSize(512)
+        }
+        +terrain(SOURCE) {
+          exaggeration(1.7)
```
kiryldz, line 72: Do we actually need it?
If yes - move to const.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+      }
+
+      // use phase that trails behind to be able to create a tilted camera
+      val cameraLocation = TurfMeasurement.along(
+        lineString, routeDistance * cameraPhase,
+        TurfConstants.UNIT_KILOMETERS
+      )
+
+      // calculate the elevation of the user location
+      val elevation = minOf(maxElevation, (maxElevation * phase) + minElevation)
```
kiryldz, line 217: It seems that `elevation` will grow linearly from `minElevation` to `maxElevation` for some time phase starting from `phase == 0` and starting from some concrete phase will become `maxElevation` till `phase == 1.0`. 
Could you please clarify logic here?

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
```diff
@@ -0,0 +1,311 @@
...
+      locationLayer.location(
+        listOf(
+          cameraLookingAt.latitude(),
+          cameraLookingAt.longitude(),
+          elevation
+        )
+      )
+
+      // calculate an elevation difference between the user position and the camera
+      val elevationDifference = BASE_CAMERA_ELEVATION + elevation
```
kiryldz, line 229: Is it indeed difference? Looks more like a sum to me 😄

kiryldz (approved):
Spectacular example!
Asked some questions to better understand the underline logic but in general LGTM

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
tobrun: I will optimize for it but I disagree that we should always replace `!!` with graceful error handling. In the SDK, we should never use `!!`. In the test app we are allowed to use it. My personal believe is that it's better to fail fast and you at least know something is wrong. Versus silently ignoring and printing out logcat message.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
tobrun: need it, no but it looks cooler

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
tobrun: for context I'm replacing it with this code:

```
        response.body()?.let { body ->
          body.routes()[0].geometry()?.let {
            animateRoute(
              LineString.fromPolyline(it, PRECISION_6)
            )
            return
          }
        }
        throw RuntimeException("Not able to retrieve a directions route")
```

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
tobrun: correct

app/src/main/java/com/mapbox/maps/testapp/examples/terrain/SantaCatalinaActivity.kt
tobrun: updated the comments

pengdev (approved):

PR #56, opened 2021-01-28
[location-component] Add OnIndicatorPositionChangedListener and OnIndicatorBearingChangedListener.

pengdev:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add OnIndicatorPositionChangedListener and OnIndicatorBearingChangedListener.</changelog>`.

### Summary of changes

This PR adds the OnIndicatorPositionChangedListener and OnIndicatorBearingChangedListener to the location component plugin, these two callbacks will be called on each position and bearing change of the location indicator, including each animation frame.

This PR also updates the LocationComponentActivity to utilise the OnIndicatorPositionChangedListener and always have the location indicator centered.  And LocationTrackingActivity is updated to use LocationComponentPlugin instead of deprecated LocationPlugin.

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
@@ -3,15 +3,18 @@ package com.mapbox.maps.testapp.examples
 import android.os.Bundle
 import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
-import com.mapbox.android.core.location.LocationEngineRequest
+import androidx.appcompat.content.res.AppCompatResources
+import com.mapbox.android.gestures.MoveGestureDetector
 import com.mapbox.maps.CameraOptions
-import com.mapbox.maps.MapboxMap
 import com.mapbox.maps.Style
-import com.mapbox.maps.plugin.location.LocationComponentActivationOptions
-import com.mapbox.maps.plugin.location.getLocationPlugin
-import com.mapbox.maps.plugin.location.listeneres.OnCameraTrackingChangedListener
-import com.mapbox.maps.plugin.location.modes.CameraMode
-import com.mapbox.maps.plugin.location.modes.RenderMode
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
```diff
@@ -3,77 +3,115 @@ package com.mapbox.maps.testapp.examples
...
+  private fun setupGesturesListener() {
+    mapView.getGesturesPlugin().addOnMoveListener(onMoveListener)
   }
 
-  override fun onCameraTrackingDismissed() {
-    Toast.makeText(this, "onCameraTrackingDismissed", Toast.LENGTH_SHORT).show()
+  private fun initLocationComponent() {
+    val locationComponentPlugin = mapView.getLocationComponentPlugin()
+    locationComponentPlugin.updateSettings {
+      this.enabled = true
```
kiryldz, line 80: So do I get it correctly that if we miss that line - nothing will work? 😄

pengdev: right, we have to enable the location component plugin, we can either do it here in the code, or through adding the xml attribute.

kiryldz (approved):
@pengdev - great job!
As discussed personally - I will add tests later for that new functionality in separate PR.

app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
kiryldz: Can we try enabling plugin (if it's not enabled) if any function (like `updateSettings` in this case) invoked?

pengdev: I think the setting properties themselves are pretty clear, we should avoid adding too much logic and confuse users.

Chaoba (approved):

PR #67, opened 2021-02-02
[Annotation plugin] Add bitmap API for Symbol

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[Annotation plugin] Add bitmap API for Symbol</changelog>`.

### Summary of changes
This pr add the bitmap API for Symbol to enable users to add a bitmap to Symbol directly instead of adding it to style.
QQ20210203-144935

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolActivity.kt
@@ -86,16 +86,24 @@ class SymbolActivity : AppCompatActivity() {
             .withDraggable(true)
           symbol = create(symbolOptions)
         }
-        // create nearby symbols
-        val nearbyOptions: SymbolOptions = SymbolOptions()
-          .withPoint(Point.fromLngLat(NEARBY_LONGITUDE, NEARBY_LATITUDE))
-          .withIconImage(MAKI_ICON_CIRCLE)
-          .withIconColor(ColorUtils.colorToRgbaString(Color.YELLOW))
-          .withIconSize(2.5)
-          .withSymbolSortKey(5.0)
-          .withDraggable(true)
-        create(nearbyOptions)
 
+        BitmapUtils.getBitmapFromDrawable(
+          ResourcesCompat.getDrawable(
+            resources,
+            R.drawable.custom_user_arrow,
+            this@SymbolActivity.theme
+          )
+        )?.let {
+          // create nearby symbols
+          val nearbyOptions: SymbolOptions = SymbolOptions()
+            .withPoint(Point.fromLngLat(NEARBY_LONGITUDE, NEARBY_LATITUDE))
+            .withIconImage(it)
+            .withIconColor(ColorUtils.colorToRgbaString(Color.YELLOW))
+            .withIconSize(2.5)
+            .withSymbolSortKey(5.0)
+            .withDraggable(true)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolActivity.kt
```diff
@@ -80,10 +66,17 @@ class SymbolActivity : AppCompatActivity() {
...
+            resources,
+            R.drawable.ic_airplanemode_active_black_24dp,
+            this@SymbolActivity.theme
+          )
+        )
         // create a symbol
         val symbolOptions: SymbolOptions = SymbolOptions()
           .withPoint(Point.fromLngLat(AIRPORT_LONGITUDE, AIRPORT_LATITUDE))
-          .withIconImage(ID_ICON_AIRPORT)
+          .withIconImage(airportImage!!)
```
kiryldz, line 79: Let's not use `!!` here - we could check for null while preparing bitmap above and throw some exception in case `bitmap == null`

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -215,18 +218,28 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+    source?.let { geoJsonSource ->
+      if (!style.styleSourceExists(geoJsonSource.sourceId)) {
         Logger.e(TAG, "Can't update source: source has not been added to style.")
         return
       }
+      annotations
+        .filter { it.value.getType() == AnnotationType.Symbol }
+        .forEach {
+          (it.value as Symbol).iconImageBitmap?.let {
+            val imagePlugin = image(Symbol.ICON_DEFAULT_NAME) {
```
tobrun, line 230: We are creating a style image with a fixed id. What will happen if we add multiple symbols using the bitmap api? They would share the same style image id and would be overriding eachother?

tobrun (commented):
One that this this issue isn't resolving yet is for the use-case of style loading. What will happen if I use the bitmap API and load a new style? The symbol is supposed to be shown again but the style image is not part of the style spritesheet. We need to take this use-case into account and readd style images when needed.

Chaoba:
> One that this this issue isn't resolving yet is for the use-case of style loading. What will happen if I use the bitmap API and load a new style? The symbol is supposed to be shown again but the style image is not part of the style spritesheet. We need to take this use-case into account and readd style images when needed.

Every time we change to a new style, the `updateSource` will be invoked and the bitmap will be added to the new style, so no worry about this scenario 😄 , I will attach a gif for it.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: Good catch

tobrun (approved):
I think this is good for first iteration. I do think we can improve one specific use-case below but we can do that as tailwork:
 - what happens if a user adds 2 symbols using the same iconImageBitmap? It will currently add 2 bitmaps with 2 separate ids to the style. Ideally we only add one and reuse them.

PR #70, opened 2021-02-02
[camera-animation] Fix using camera animation extension functions

### LINKED ISSUE #63: Camera animations plugin doesn't work properly with multiple mapView instances.
## Environment

- Android OS version: 10
- Devices affected: OnePlus 7 Pro
- Maps SDK Version: v10.0.0-beta.12

## Observed behavior and steps to reproduce

When the application uses multiple MapViews, the high level animation APIs like `easeTo`, `flyTo`,`pitchBy` etc. will not work properly for all the MapViews.  These camera animations will only be applied to the last created MapView.

## Expected behavior

The high level animation APIs should work property for every MapView created in the application.

## Notes / preliminary analysis

Currently in the camera animation plugin, the high level animation APIs, such as `easeTo`, `flyTo`,`pitchBy` etc, are extension functions of the `MapboxMap`, but they are not bound to the specific `MapboxMap` instance. These  extension functions use the WeakReference of the animations plugin inside the `CameraAnimationsPluginImpl`'s companion object. As a result, animation plugin created together with the new MapView will override the old weak reference.

kiryldz:
Resolves #63 
`<changelog>[camera-animation] Fix using camera animation extension functions with multiple MapView's</changelog>`

## Pull request checklist:

### Summary of changes

Fix using camera animation extension functions with multiple MapView's

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -130,7 +130,7 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
    * Called when the plugin is first added to the map.
    */
   override fun initialize() {
-    animationsPluginWeakRef = WeakReference(this)
+    pluginRegistry[mapTransformDelegate] = WeakReference(this)
   }
 
   /**
@@ -138,7 +138,7 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
    * Cancel all running animations and cleanup all resources (registered animations, listeners).
    */
   override fun cleanup() {
-    unregisterAnimators(*animators.map { it as ValueAnimator }.toTypedArray())
+    unregisterAnimators(*animators.toTypedArray())
     cancelAnimatorSet()
     centerListeners.clear()
     zoomListeners.clear()
@@ -148,8 +148,9 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     paddingListeners.clear()
     lifecycleListener.clear()
     animators.clear()
-    animationsPluginWeakRef?.clear()
-    animationsPluginWeakRef = null
+    pluginRegistry.remove(mapTransformDelegate).also { weakRef ->
+      weakRef?.clear()
+    }
   }
 
   private fun performMapJump(cameraOptions: CameraOptions) {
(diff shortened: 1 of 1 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -816,7 +817,8 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
    */
   companion object {
     private const val TAG = "Mbgl-CameraManager"
-    internal var animationsPluginWeakRef: WeakReference<CameraAnimationsPluginImpl>? = null
+    internal var pluginRegistry: HashMap<MapTransformDelegate, WeakReference<CameraAnimationsPluginImpl>> =
```
tobrun, line 820: Is there a possibility to remove this dependency on companion object fully? Why do we need to have static configuration at all?

kiryldz: @tobrun - I can't see any other options to make extension functions work other than have some static registry. We need some kind of link (preferably weak one) to `CameraAnimationsPluginImpl` to declare e.g.:
```
fun MapTransformDelegate.easeTo(
  cameraOptions: CameraOptions,
  animationOptions: MapAnimationOptions? = null
) = CameraAnimationsPluginImpl.pluginRegistry[this]?.get()?.easeTo(cameraOptions, animationOptions)
```
I do not like that approach as well but can't come up with another one, if you have any ideas - would like to hear them!

pengdev: How about add a plugin provider interface into the MapboxMap? so that we can get the plugin instance from the MapboxMap and the extension function should work fine with the MapboxMap instance.

kiryldz: @pengdev camera plugin knows actually nothing about `MapboxMap`.
We could enhance `interface MapTransformDelegate` to make it work...
Technically we could add introduce new interface to `sdk-base/src/main/java/com/mapbox/maps/plugin/delegates` and try make it correct way.
Let's track it as separate task.

pengdev (approved):
The fix LGTM, let's ticket out a more graceful way to handle this situation and refactor the whole codebase regarding the extension functions with multiple maps. We might have similar issue with GesturesPlugin etc.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: Tracking tailwork here https://github.com/mapbox/mapbox-maps-android/issues/73

Issue #73, opened 2021-02-03
Revisit MapboxMap plugin extension functions


kiryldz:
## Environment

- Android OS version: any
- Devices affected: any
- Maps SDK Version: v10.0.0-beta.12

Tailwork for https://github.com/mapbox/mapbox-maps-android/issues/63
We need to revisit approach in general of using some plugin functionality as extension functions for `MapboxMap`.

Current examples are `plugin-animation`, `plugin-gestures` that expose some of their functions as extension functions (e.g. `MapboxMap.flyTo`, `MapboxMap.setGesturesManager`).

Camera plugin was fixed in https://github.com/mapbox/mapbox-maps-android/issues/63 but still we're using weak references stored in static `HashMap` which is obviously not perfect. Gestures plugin will work incorrectly now if using several `MapboxMap` instances.

Aim of this task is to introduce complete new approach to connect plugins with `MapboxMap` and make it easy to add extension functions that will work properly in case of several `MapboxMap` objects.

PR #75, opened 2021-02-03
Revisit MapboxMap plugin related extension functions to not use singletons

kiryldz:
Fixes: #73 
`<changelog>Revisit MapboxMap plugin related extension functions to not use singletons</changelog>`

## Pull request checklist:

### Summary of changes

Fix extension function usage in case of several instances of `MapboxMap` objects if using plugins.


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -32,13 +35,17 @@ class MapboxMap internal constructor(
...
+  MapPluginExtensionsDelegate,
   MapStyleStateDelegate {
 
   private val nativeMapWeakRef = WeakReference(nativeMap)
   internal lateinit var style: Style
 
   private var terrainEnabled = false
 
+  private var cameraAnimationsPlugin: WeakReference<CameraAnimationsPlugin>? = null
+  private var gesturesPlugin: WeakReference<GesturesPlugin>? = null
```
pengdev, line 47: Since we can get the CameraAnimationsPlugin and GesturesPlugin from the MapView, would it possible to get a reference of the MapView inside the MapboxMap? and then we don't need such complex setup, instead, we could use an extension like:

```
interface MapViewProviderDelegate {
  fun getMapView() : MapView
}
```

`MapboxMap : MapViewProviderDelegate {
  ...
}`

And in the CameraAnimationsPlugin, we could have:

```
fun MapViewProviderDelegate.flyTo(point: Point) {
  this.getMapView().getCameraAnimationPlugin().flyTo(point)
}
```

wdyt?

kiryldz: The main problem at all is that technically `MapView` contains `MapboxMap` : `MapView -> MapController -> MapboxMap`. If we add direct reverse link we technically get clear circular dependency.

Current approach seems lesser evil to me.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1057,4 +1064,28 @@ class MapboxMap internal constructor(
...
+    }
+  }
+
+  /**
+   * Get an instance of [CameraAnimationsPlugin].
+   * In most cases must not be called directly.
+   */
+  override fun getCameraAnimationPlugin() = cameraAnimationsPlugin?.get()
+
+  internal fun setGesturesAnimationPlugin(gesturesPlugin: GesturesPlugin?) {
```
Chaoba, line 1080: In `MapPluginRegistry` we are using `CopyOnWriteArrayList` to cache gesture and animation plugin instances, that said there might be multi gesture and animation plugins. In that case, it will overwrite the former ones.
How about add a API `addMapPluginRegistry` for `MapboxMap`?

tobrun (commented):
I appreciate the thinking outside the box. Moving the extension functions to SDK-base makes them always part of the SDK public API. We will conditionally execute APIs if the plugin dependency has been met. If we want to moe forward to such a setup, we can also replace the extension functions from sdk-base and add them directly on MapboxMap.

Note how this is conceptually different as the initial idea behind it. We only wanted to populate the MapboxMap API with new APIs if a plugin dependency was met. If you add animations plugin dependency, "magically" MapboxMap was able to use flyTo etc. without it, it wasn't.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
tobrun: I follow @pengdev suggestion of adding another interface, another abstraction layer, to work around the current issue. MapView might not be the best object as it's part of the SDK module not base. Abstractions need to occur in SDK base and be implemented in the SDK module to have clear separations of concerns. 

> circular dependency

In JVM languages, circular dependencies aren't as a big issue as with others languages. Don't get me wrong, it's a not a great practice but afaik the downsides are that they make the code hard to understand and make it difficult for the garbage collector to do his work. However, there is always a way to untangle these constructs by introducing a third party inside that circular relation.

kiryldz:
> I appreciate the thinking outside the box. Moving the extension functions to SDK-base makes them always part of the SDK public API. We will conditionally execute APIs if the plugin dependency has been met. If we want to moe forward to such a setup, we can also replace the extension functions from sdk-base and add them directly on MapboxMap.
> 
> Note how this is conceptually different as the initial idea behind it. We only wanted to populate the MapboxMap API with new APIs if a plugin dependency was met. If you add animations plugin dependency, "magically" MapboxMap was able to use flyTo etc. without it, it wasn't.

@tobrun - thanks a lot for such detailed write-up! I actually did not realize that bringing extension functions to `sdk-base` actually ruin all that "magic" as you described. I guess then indeed the best option is do it in a way @pengdev suggested.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: Good catch - now we apply extension functions to one concrete plugin.

tobrun (approved):
Super clean, I like it

Chaoba (approved):
LGTM

pengdev:
Great work! can we add tests to these functions?

kiryldz:
> Great work! can we add tests to these functions?

Yep, working on that right now 👍

PR #77, opened 2021-02-04
[sdk] Use cached resourceOptions if it's initialized.

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[sdk] Use cached resourceOptions if it's initialized.</changelog>`.

### Summary of changes
This pr add the function that checks and use the cached `resourceOptions` if it's initialized instead of recreate it from ResourcesOptions

### DIFF
--- a/sdk/src/androidTest/java/com/mapbox/maps/MapboxMapIntegrationTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/MapboxMapIntegrationTest.kt
@@ -36,12 +36,7 @@ class MapboxMapIntegrationTest {
   fun testGetResourceOptions() {
     val defaultOptions = MapboxOptions.getDefaultResourceOptions(mapView.context)
     val currentOptions = mapboxMap.getResourceOptions()
-    // TODO use assertEquals(defaultOptions, currentOptions) when equals() is supported on records
-    assertEquals(defaultOptions.accessToken, currentOptions.accessToken)
-    assertEquals(defaultOptions.baseURL, currentOptions.baseURL)
-    assertEquals(defaultOptions.assetPath, currentOptions.assetPath)
-    assertEquals(defaultOptions.cachePath, currentOptions.cachePath)
-    assertEquals(defaultOptions.cacheSize, currentOptions.cacheSize)
+    assertEquals(defaultOptions, currentOptions)
   }
 
   @UiThreadTest
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMapOptions.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMapOptions.kt
@@ -61,7 +61,11 @@ data class MapboxMapOptions constructor(
   init {
     val typedArray = context.obtainStyledAttributes(attrs, R.styleable.mapbox_MapView, 0, 0)
     try {
-      internalResourceOptions = ResourcesAttributeParser.parseResourcesOptions(context, typedArray)
+      internalResourceOptions = if (MapboxOptions.isInitialized()) {
+        MapboxOptions.getDefaultResourceOptions(context)
+      } else {
+        ResourcesAttributeParser.parseResourcesOptions(context, typedArray)
+      }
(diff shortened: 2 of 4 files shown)


kiryldz (commented):
@Chaoba let's imagine we have 2 `MapView`s in one layout and we want to set different tokens to each of it programatically (and also different from `<string name="mapbox_access_token"></string>`) . Can it be done with current setup? If yes - guess we should provide an example in `app` module.

sdk/src/androidTest/java/com/mapbox/maps/MapboxMapIntegrationTest.kt
```diff
@@ -36,12 +36,7 @@ class MapboxMapIntegrationTest {
   fun testGetResourceOptions() {
     val defaultOptions = MapboxOptions.getDefaultResourceOptions(mapView.context)
     val currentOptions = mapboxMap.getResourceOptions()
-    // TODO use assertEquals(defaultOptions, currentOptions) when equals() is supported on records
-    assertEquals(defaultOptions.accessToken, currentOptions.accessToken)
-    assertEquals(defaultOptions.baseURL, currentOptions.baseURL)
-    assertEquals(defaultOptions.assetPath, currentOptions.assetPath)
-    assertEquals(defaultOptions.cachePath, currentOptions.cachePath)
-    assertEquals(defaultOptions.cacheSize, currentOptions.cacheSize)
+    assertEquals(defaultOptions, currentOptions)
```
pengdev, line 39: nice catch!

pengdev (approved):

pengdev:
> @Chaoba let's imagine we have 2 `MapView`s in one layout and we want to set different tokens to each of it programatically (and also different from `<string name="mapbox_access_token"></string>`) . Can it be done with current setup? If yes - guess we should provide an example in `app` module.

I guess it's hard to showcase this example without exposing the access token in the code.

Chaoba:
> @Chaoba let's imagine we have 2 `MapView`s in one layout and we want to set different tokens to each of it programatically (and also different from `<string name="mapbox_access_token"></string>`) . Can it be done with current setup? If yes - guess we should provide an example in `app` module.

@kiryldz Even without the changes in this pr, set different tokens to 2 MapView is available. This pr resolves the scenario that if users use interface `MapboxOptions#setDefaultResourceOptions` updated the ResourceOptions and create a new MapView, the updated ResourceOptions will not work on the new MapView.

kiryldz:
> > @Chaoba let's imagine we have 2 `MapView`s in one layout and we want to set different tokens to each of it programatically (and also different from `<string name="mapbox_access_token"></string>`) . Can it be done with current setup? If yes - guess we should provide an example in `app` module.
> 
> I guess it's hard to showcase this example without exposing the access token in the code.

We could use any string and not make `MapView` show data (because token is invalid).
However that use-case is pretty real, I faced it myself in our Preview App, when valid token comes through web and gets saved to shared preferences and then must be used with `MapView` inflated from xml.

kiryldz:
> > @Chaoba let's imagine we have 2 `MapView`s in one layout and we want to set different tokens to each of it programatically (and also different from `<string name="mapbox_access_token"></string>`) . Can it be done with current setup? If yes - guess we should provide an example in `app` module.
> 
> @kiryldz Even without the changes in this pr, set different tokens to 2 MapView is available. This pr resolves the scenario that if users use interface `MapboxOptions#setDefaultResourceOptions` updated the ResourceOptions and create a new MapView, the updated ResourceOptions will not work on the new MapView.

My concern here is that our current logic will work fine with 1 `MapView` in xml - then we could `MapboxOptions#setDefaultResourceOptions` before `setContentView` in activity. But if we have 2 `MapView`s in the layout and each of it should have it's own token - that will not work I guess.
Generally such scenario should not be pretty common but anyway.

kiryldz (approved):
LGTM, created a task to track revisiting approach in general.

PR #79, opened 2021-02-05
[Annotation plugin] Add GeoJsonOptions to allow to configure the underlying used GeoJsonSource

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[Annotation plugin] Add GeoJsonOptions to allow to configure the underlying used GeoJsonSource</changelog>`.

### Summary of changes
This pr add a config class `GeoJsonOptions` to allow to configure the underlying used GeoJsonSource and add `toString` method for `Layer`

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/CircleManager.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/CircleManager.kt
@@ -232,16 +232,7 @@ class CircleManager(
    * @return the layer created
    */
   override fun createLayer(): CircleLayer {
-    return circleLayer(layerId, sourceId) {
-      annotationConfig?.geoJsonOptions?.let { geoJsonOptions ->
-        geoJsonOptions.minZoom?.let {
-          minZoom(it)
-        }
-        geoJsonOptions.maxZoom?.let {
-          maxZoom(it)
-        }
-      }
-    }
+    return circleLayer(layerId, sourceId) {}
   }
 
   /**
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/CircleManager.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/CircleManager.kt
@@ -7,9 +7,8 @@ import com.mapbox.geojson.*
 import com.mapbox.maps.extension.style.expressions.generated.Expression
 import com.mapbox.maps.extension.style.expressions.generated.Expression.Companion.get
 import com.mapbox.maps.extension.style.layers.generated.CircleLayer
+import com.mapbox.maps.extension.style.layers.generated.circleLayer
 import com.mapbox.maps.extension.style.layers.properties.generated.*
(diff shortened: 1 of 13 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/CircleManager.kt
```diff
@@ -229,22 +226,22 @@ class CircleManager(
...
+    return circleLayer(layerId, sourceId) {
+      annotationConfig?.geoJsonOptions?.let { geoJsonOptions ->
+        geoJsonOptions.minZoom?.let {
+          minZoom(it)
+        }
+        geoJsonOptions.maxZoom?.let {
+          maxZoom(it)
+        }
+      }
+    }
```
tobrun, line 244: Both source as layer have the concept of minZoom and maxZoom (they do different things though) but conceptually, the minZoom/maxZoom provided to the geoJsonOptions should only be applied to the source. I do think we can ticket a feature request for exposing this additionally to layer but as separate properties.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/GeoJsonOptions.kt
```diff
@@ -0,0 +1,68 @@
+package com.mapbox.maps.plugin.annotation
+
+/**
+ * Configure class for composing GeoJsonSource objects that included in Annotation
+ * @see <a href="https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson">The online documentation</a>
+ */
+data class GeoJsonOptions(
```
tobrun, line 7: Adding this class to SDK-base and making it not fully compatible with style extension will be confusing for users that are migrating from the old API. Since it's only used for annotations, what about we rename this class to `AnnotationSourceOptions` and move it to annotation plugin itself?

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/GeoJsonOptions.kt
```diff
@@ -0,0 +1,68 @@
...
+   * The minimum zoom level for the layer. At zoom levels less than the minzoom, the layer will be hidden.
+   * Range: [0, 24]
+   */
+  val minZoom: Double? = null,
+
+  /**
+   * The maximum zoom level for the layer. At zoom levels equal to or greater than the maxzoom, the layer will be hidden.
+   * Range: [0, 24]
+   */
+  val maxZoom: Double? = null,
```
tobrun, line 18: the configurations in this class should only impact the source configuration. We could broaden this but if so, we need to qualify what maxZoom/minZoom you are changing layer or source.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/GeoJsonOptions.kt
Chaoba: We will need to keep this class in sdk-base as it is part of the API, move it to annotation plugin will make it not available for users.

tobrun (approved):

PR #103, opened 2021-02-12
Change default gesture interpolator to LinearOutSlowInInterpolator

tobrun:
This PR changes the default interpolator from `Decelerate` to `LinearOutSlowInInterpolator`. Conceptually they do the same thing but the latter was introduced as part of material design spec as an improved version of the former. 

See https://material.io/design/motion/speed.html > decelerate easing. 

cc @kiryldz


pengdev (approved):
LGTM

PR #104, opened 2021-02-12
[map] remove renderStill API, this is an internal API

tobrun:
The renderStill API wasn't supposed to be exposed. Users can leverage snapshot() instead.


pengdev:
Probably need to also remove the unit test as well.

pengdev (approved):

Issue #82, opened 2021-02-05
v10 zoom and rotation do not use correct axis


sbma44:
## Environment
- Android OS version: 9
- Devices affected: replicated on Lenovo TB-8505FS
- Maps SDK Version: v10 beta 12

## Observed behavior and steps to reproduce
Same issue as reported for iOS in https://github.com/mapbox/mapbox-maps-ios/issues/65 with exception of fact that panning works while two-finger gesture is in progress, which is not true on iOS.

## Expected behavior
pinch to zoom and rotate center on the gesture rather than center of the map

pengdev:
Tried to reproduce the issue locally, the issue seems to be related to the animations plugin. 
If I change https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt#L568 from `easeTo` to `jumpTo`, the issue seems to be fixed.


cc @mapbox/maps-android

kiryldz:
> Tried to reproduce the issue locally, the issue seems to be related to the animations plugin.
> If I change https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt#L568 from `easeTo` to `jumpTo`, the issue seems to be fixed.
> 
> Since all the easeTo functions uses the same cameraAnimation options:
> 
> ```
> mapAnimationOptions {
>       duration = 0
>       owner = MapAnimationOwnerRegistry.GESTURES
>     }
> ```
> 
> Shall we remove the dependency of AnimationsPlugin and use jumpTo instead? would this cause issue to the gestures ownership management?
> 
> cc @mapbox/maps-android @kiryldz

That was the main idea to use `easeTo` with zero duration to perform immediate jump but do it via animations plugin. I'll take a look at it - `easeTo` with `duration = 0` should actually do a single `jumpTo` under the hood, seems to be a bug.

PR #109, opened 2021-02-16
[animations] Process anchor animators correctly

kiryldz:
Fixes: #82 
`<changelog>[animations] Cache map anchor in animation plugin, fix anchor camera animators</changelog>`

## Pull request checklist:

### Summary of changes

Treat `CameraOptions.anchor(ScreenCoordinate)` as all other camera options (zoom, bearing, etc). It means that if some animator did change an anchor - it will be saved and cached in `CameraAnimationPlugin` and used as initial start value for all upcoming animators if start value not specified explicitly.
This change fixes using anchor animators in gesture plugin.
`CameraAnimationsPlugin.anchor` override var is added to get / set anchor values.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -153,7 +153,8 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     val startValue = cameraAnimator.startValue ?: when (cameraAnimator.type) {
       CameraAnimatorType.CENTER -> mapCameraDelegate.getCameraOptions().center
       CameraAnimatorType.ZOOM -> mapCameraDelegate.getCameraOptions().zoom
-      CameraAnimatorType.ANCHOR -> anchor
+      // TODO revisit after https://github.com/mapbox/mapbox-maps-android/issues/119
+      CameraAnimatorType.ANCHOR -> anchor ?: ScreenCoordinate(0.0, 0.0)
       CameraAnimatorType.PADDING -> mapCameraDelegate.getCameraOptions().padding
       CameraAnimatorType.BEARING -> mapCameraDelegate.getCameraOptions().bearing
       CameraAnimatorType.PITCH -> mapCameraDelegate.getCameraOptions().pitch
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -43,7 +43,7 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
   private val zoomListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Double>>()
   private val paddingListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<EdgeInsets>>()
   private val anchorListeners =
-    CopyOnWriteArraySet<CameraAnimatorChangeListener<ScreenCoordinate>>()
+    CopyOnWriteArraySet<CameraAnimatorNullableChangeListener<ScreenCoordinate?>>()
   private val bearingListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Double>>()
   private val pitchListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Double>>()
 
@@ -73,11 +73,9 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     }
   }
 
-  private var anchor by Delegates.observable<ScreenCoordinate?>(null) { _, old, new ->
-    new?.let {
-      if (old != it) {
-        anchorListeners.forEach { listener -> listener.onChanged(it) }
-      }
+  override var anchor by Delegates.observable<ScreenCoordinate?>(null) { _, old, new ->
+    if (old != new) {
+      anchorListeners.forEach { listener -> listener.onChanged(new) }
     }
   }
 
@@ -155,7 +153,7 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     val startValue = cameraAnimator.startValue ?: when (cameraAnimator.type) {
       CameraAnimatorType.CENTER -> mapCameraDelegate.getCameraOptions().center
       CameraAnimatorType.ZOOM -> mapCameraDelegate.getCameraOptions().zoom
-      CameraAnimatorType.ANCHOR -> cameraAnimator.startValue
+      CameraAnimatorType.ANCHOR -> anchor
       CameraAnimatorType.PADDING -> mapCameraDelegate.getCameraOptions().padding
       CameraAnimatorType.BEARING -> mapCameraDelegate.getCameraOptions().bearing
       CameraAnimatorType.PITCH -> mapCameraDelegate.getCameraOptions().pitch
@@ -191,7 +189,6 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
   internal fun notifyListeners(cameraOptions: CameraOptions) {
     cameraOptions.also {
-      anchor = it.anchor
       bearing = it.bearing
       center = it.center
       padding = it.padding
@@ -278,7 +275,11 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
             unregisterAnimators(this, cancelAnimators = false)
           }
           if (runningAnimatorsQueue.isEmpty()) {
-            performMapJump(cameraOptionsBuilder.build())
+            // TODO revisit after https://github.com/mapbox/mapbox-maps-android/issues/119
+            if (animator.type == CameraAnimatorType.ANCHOR) {
+              anchor = animatedValue as ScreenCoordinate
+            }
+            performMapJump(cameraOptionsBuilder.anchor(anchor).build())
             mapTransformDelegate.setUserAnimationInProgress(false)
           }
         } ?: throw RuntimeException(
@@ -312,7 +313,12 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
           null
         }
       }
+      // TODO revisit after https://github.com/mapbox/mapbox-maps-android/issues/119
+      if (animator.type == CameraAnimatorType.ANCHOR) {
+        anchor = it.animatedValue as ScreenCoordinate
+      }
       cameraOptions?.let { camera ->
+        camera.anchor = anchor
         // move map camera
         performMapJump(camera)
         // reset values
@@ -471,20 +477,20 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
   }
 
   /**
-   * Add [CameraAnimatorChangeListener] to receive map anchor updates.
+   * Add [CameraAnimatorNullableChangeListener] to receive map anchor updates.
    *
-   * @param listener Instance of [CameraAnimatorChangeListener]
+   * @param listener Instance of [CameraAnimatorNullableChangeListener]
    */
-  override fun addCameraAnchorChangeListener(listener: CameraAnimatorChangeListener<ScreenCoordinate>) {
+  override fun addCameraAnchorChangeListener(listener: CameraAnimatorNullableChangeListener<ScreenCoordinate?>) {
     anchorListeners.add(listener)
   }
 
   /**
-   * Remove [CameraAnimatorChangeListener]. No updates will arrive after that.
+   * Remove [CameraAnimatorNullableChangeListener]. No updates will arrive after that.
    *
-   * @param listener Instance of [CameraAnimatorChangeListener]
+   * @param listener Instance of [CameraAnimatorNullableChangeListener]
    */
-  override fun removeCameraAnchorChangeListener(listener: CameraAnimatorChangeListener<ScreenCoordinate>) {
+  override fun removeCameraAnchorChangeListener(listener: CameraAnimatorNullableChangeListener<ScreenCoordinate?>) {
     anchorListeners.remove(listener)
   }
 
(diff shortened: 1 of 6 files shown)


tobrun (approved):
:shipit:

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -278,7 +284,11 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
             unregisterAnimators(this, cancelAnimators = false)
           }
           if (runningAnimatorsQueue.isEmpty()) {
-            performMapJump(cameraOptionsBuilder.build())
+            // TODO handle anchor update manually, to be fixed in core to obtain from CameraOptions
```
pengdev, line 287: can we have a ticket tracking it in this repo, so that it can be picked up when the fix in core is ready?

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -727,11 +730,12 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
       duration = animationTime
     }
 
-    bearingAnimator.addUpdateListener {
+    anchorAnimator.addUpdateListener {
```
pengdev, line 733: any reason removing the bearingAnimator's update listener.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -988,13 +992,14 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
       duration = animationTime
     }
 
-    zoomAnimator.addListener(object : AnimatorListenerAdapter() {
```
pengdev, line 991: any reason removing the zoomAnimator update listener?

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -988,13 +992,14 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
...
+    anchorAnimator.addListener(object : AnimatorListenerAdapter() {
 
       override fun onAnimationStart(animation: Animator) {}
 
       override fun onAnimationCancel(animation: Animator) {}
 
       override fun onAnimationEnd(animation: Animator) {
+        cameraAnimationsPlugin.resetAnchor()
```
LukasPaczos, line 1002: When the gesture finishes it shouldn't reset the anchor to map center but rather to the value that was there previously, right?

This is what is proposed:
1. User transitions anchor from X -> Y
2. Gesture starts and transitions the anchor from Y -> Z
3. Gesture ends and transitions the anchor from Z -> map_center

This looks incorrect to me, I think it should rather be:
1. User transitions anchor from X -> Y
2. Gesture starts and transitions the anchor from Y -> Z
3. Gesture ends and transitions the anchor from Z -> Y

kiryldz: Good catch, it means we should actually expose `cameraAnimationsPlugin.getAnchor()` as I did initially but then decided reset is enough 😄 
Will do!

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: yes, we actually create zoom + anchor animator and run them as a set.
so it's more logical to reset anchor when anchor animator ends (although it's technically same moment when zoom animator ends).

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: yes, we actually create bearing + anchor animator and run them as a set.
so it's more logical to reset anchor when anchor animator ends (although it's technically same moment when bearing animator ends).

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: https://github.com/mapbox/mapbox-maps-android/issues/119

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -778,6 +790,17 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     animatorSet.start()
   }
 
+  override fun onSizeChanged(width: Int, height: Int) {
+    mapCenter = ScreenCoordinate((width / 2).toDouble(), (height / 2).toDouble())
+    cameraAnchor = mapCenter
+  }
```
LukasPaczos, line 796: Would be great to document this relationship.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -778,6 +790,17 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+
+  override fun getAnchor() = cameraAnchor
+
+  override fun resetAnchor(anchor: ScreenCoordinate?) {
+    cameraAnchor = anchor ?: mapCenter
```
LukasPaczos, line 801:
```suggestion
    cameraAnchor = anchor ?: mapCenter
    map.jumpTo(cameraAnchor)
```

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -988,13 +997,14 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
...
 
-    zoomAnimator.addListener(object : AnimatorListenerAdapter() {
+    anchorAnimator.addListener(object : AnimatorListenerAdapter() {
 
       override fun onAnimationStart(animation: Animator) {}
 
       override fun onAnimationCancel(animation: Animator) {}
 
       override fun onAnimationEnd(animation: Animator) {
+        cameraAnimationsPlugin.resetAnchor(cachedAnchor)
```
LukasPaczos, line 1007: What if the zoom gesture velocity animation was canceled by a developer's anchor animation?

I guess the cancelation happens first, so we will first reset the anchor to the cached value, and then the developer's transition will proceed so the end result should be correct, right?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: @LukasPaczos jumping to `CameraOptions` with only `anchor` specified will actually result in nothing for user. I guess simply setting it here and performing `jumpTo` later on as part of some upcoming animation with some other camera options like zoom, bearing should work fine.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: @LukasPaczos yes, it will work like this - we would first cancel, it will also trigger an end callback, we will reset anchor and start user anchor animation with cached value that he was actually expecting.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: I remembered our discussion to send `null` as anchor to core - if I understand correctly it's equivalent to setting to map center as it is done now.
Will check it out, if it will work - it will simplify PR a bit, we will not need to override `onSizeChanged` at all.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -309,6 +309,7 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
 
           // Scale the map by the appropriate power of two factor
           val currentZoom = mapTransformDelegate.getCameraOptions(null).zoom
+          val cachedAnchor = cameraAnimationsPlugin.anchor
```
pengdev, line 312: Do we need to get and set `cameraAnimationsPlugin.anchor` before and after the `cameraAnimationsPlugin.easeTo`? 

This seems a bit too much extra work for end-developers. Is there any way we can improve the caching to be totally on the cameraAnimationsPlugin side?

kiryldz: It's specifics of working with anchor for gestures plugin only.
Treat it as let's say zoom - if some flyTo changes it - map will actually have that zoom as part of current camera options.

It looks complicated for gestures because we want to make sure we do not actually change anchor property in the end, it should be changed during gestures processing only. 
In regular case if user knows what he's doing - he might not need to cache start value and set it in the end, perhaps it's his business case: set anchor with some animator and apply same anchor for all upcoming animations.

pengdev: I tried to comment out these getter and setters, it seems the map gestures are still working as expected.

> Treat it as let's say zoom - if some flyTo changes it - map will actually have that zoom as part of current camera options.

I think in this case the new animation is overwriting the gestures plugin's animation, I guess the new value should overwrite the old value, right?

> It looks complicated for gestures because we want to make sure we do not actually change anchor property in the end, it should be changed during gestures processing only.

It seems this additional logic (resetting the anchor after the gestures are processed) doesn't match the v9 behaviour. Considering if the native `CameraOptions.getAnchor` is fixed, will we remove these logic?

kiryldz: @pengdev those getters / setters are for such case
https://github.com/mapbox/mapbox-maps-android/pull/109#discussion_r577498600

kiryldz: Again - the reason for doing so here is not to interfere with some custom user animator logic.
Changing anchor is not visible as let's say zoom and if gesture plugin would silently change anchor - it's no good. That's why we have cleaning up in terms of anchor for gestures.

kiryldz: If let's say anchor was changed by gestures and not reset afterwards then user animation like
```
  mapView.getMapboxMap().flyTo(
      CameraOptions.Builder().zoom(3.0).build(),
      mapAnimationOptions {
        duration(DURATION)
      }
    )
```
would use that anchor silently under the hood because it's not specified in this camera options explicitly and user did not keep in mind that anchor was changed by gestures beforehand.

pengdev: > would use that anchor silently under the hood because it's not specified in this camera options explicitly and user did not keep in mind that anchor was changed by gestures beforehand.

Would then reset the anchor to null after the animation finishes solve this issue?

kiryldz: > Would then reset the anchor to null after the animation finishes solve this issue?

Yes that's what we're doing here.
But we can't reset to NULL blindly because it may have been set by user to some non-null value before gesture.
 > A new animation without specifying the anchor should be treated as anchor being in the center?

No, anchor is treated as any other camera option so if user has set anchor X and then runs animation __without__ specifying anchor afterwards - that anchor X will be used as default and not screen center (also equivalent to `anchor = null`)

pengdev: Thanks for the clarification 👍

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPlugin.kt
```diff
@@ -11,6 +11,13 @@ import com.mapbox.maps.plugin.MapPlugin
  * Interface to interact with Camera Animations plugin
  */
 interface CameraAnimationsPlugin : MapPlugin {
+
+  /**
+   * Map camera anchor value.
+   * If equal to NULL then center of map view is treated as anchor.
+   */
+  var anchor: ScreenCoordinate?
```
LukasPaczos, line 19: Not sure whether this is the best way to expose the value publicly. As far as I can tell, the SDK does not set `NULL` as an anchor anywhere. Developers cannot set the value to `NULL` either, the animator's constructor values are non-nullable. So this will always be `ScreenCoordinate(0, 0)` beside the initial state. Is that worth introducing a new listener and nullable value type? Wondering whether we could just initialize with `ScreenCoordinate(0, 0)` and it would actually be easier to consume.

kiryldz: Discussed locally that `ScreenCoordinate(0.0, 0.0)` is not center but top-left corner so this approach will not work. Will be leaving current approach with some clarifying docs.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: Not relevant anymore.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: Not relevant anymore.

PR #120, opened 2021-02-22
[deps] update gl-native to beta.15 and common to beta.11, adopt break…

tobrun:
- Update gl-native to `v10.0.0-beta.15` and common `v10.0.0-beta.11`.
 - Update breaking change of signature of `Map#getElevation` from Float to Double. 
 - Update breaking change of `text-field` property in SymbolLayer and `data` property GeoJsonSource.


buildSrc/src/main/kotlin/Project.kt
```diff
@@ -66,8 +66,8 @@ object Versions {
   const val mapboxGestures = "0.7.0"
   const val mapboxJavaServices = "5.4.1"
   const val mapboxBase = "0.5.0"
-  const val mapboxGlNative = "10.0.0-beta.14"
-  const val mapboxCommon = "10.0.0-beta.9.1"
+  const val mapboxGlNative = "10.0.0-beta.15"
+  const val mapboxCommon = "10.0.0-beta.11"
```
kiryldz, line 70: Lost `10.0.0-beta.10` somewhere 😄

kiryldz (approved):

pengdev (approved):

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -131,7 +131,6 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
 
   private fun createSource(): GeoJsonSource {
     return geoJsonSource(sourceId) {
-      data("")
```
pengdev, line 134: This is a breaking change, previously we allow setting empty GeoJsonSource and update the GeoJson afterwards.

app/src/main/java/com/mapbox/maps/testapp/examples/RestrictBoundsActivity.kt
```diff
@@ -37,7 +38,7 @@ class RestrictBoundsActivity : AppCompatActivity() {
     mapboxMap.loadStyle(
       style(Style.SATELLITE_STREETS) {
         +geoJsonSource(BOUNDS_ID) {
-          data("")
+          featureCollection(FeatureCollection.fromFeatures(listOf()))
```
pengdev, line 41: If this works, let's move the code to the annotation plugin, but this will break some other examples as well. For example: RestrictBoundsActivity.

Chaoba (approved):

pengdev:
@tobrun Updated the generated style extension code and tests to accommodate the breaking changes of text-field property in SymbolLayer and data property GeoJsonSource. This PR should be unblocked after the CI passes.

PR #122, opened 2021-02-23
[Annotation plugin] Add symbol cluster support

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[Annotation plugin] Add symbol cluster support</changelog>`.

### Summary of changes
This pr add symbol cluster support as well as the test demo Activity.
QQ20210224-141326


app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+    mapboxMap = mapView.getMapboxMap()
+      .apply {
+        jumpTo(
+          CameraOptions.Builder()
+            .center(Point.fromLngLat(LONGITUDE, LATITUDE))
+            .zoom(10.0)
+            .build()
+        )
+        loadStyleUri(nextStyle) {
+          launch {
```
kiryldz, line 52: wow, coroutines! 👍

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+    if (options == null) {
+      AnnotationUtils.loadStringFromAssets(
+        this@SymbolClusterActivity,
+        "points.geojson"
+      )?.let {
+        FeatureCollection.fromJson(it).features()?.let { features ->
+          features.shuffle()
+          options = features.take(AMOUNT).map { feature ->
+            SymbolOptions()
+              .withGeometry((feature.geometry() as Point))
```
kiryldz, line 88: Is this `as` needed here?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -58,7 +69,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private val mapLongClickResolver = MapLongClick()
   private val mapMoveResolver = MapMove()
   private var draggedAnnotation: T? = null
-  private val annotationMap = mutableMapOf<Long, T>()
+  private val annotationMap = ConcurrentHashMap<Long, T>()
```
kiryldz, line 72: why do we concurrency here?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -183,10 +192,60 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+        }
+      }
+      val clusterTextLayer = createClusterTextLayer()
+      if (!style.styleLayerExists(clusterTextLayer.layerId)) {
+        style.addLayer(clusterTextLayer)
+      }
+    }
+  }
+
+  private fun createClusterLevelLayer(level: Int, colorLevels: List<Pair<Int, Int>>): CircleLayer {
```
kiryldz, line 217: Could simplify here to `private fun createClusterLevelLayer(level: Int, colorLevels: List<Pair<Int, Int>>) = circleLayer() { ... }`

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -183,10 +192,60 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+        ) else all(
+          has("point_count"),
+          gt(pointCount, literal(colorLevels[level].first.toLong())),
+          lt(pointCount, literal(colorLevels[level - 1].first.toLong()))
+        )
+      )
+    }
+  }
+
+  private fun createClusterTextLayer(): SymbolLayer {
```
kiryldz, line 237: Same as above for circle layer

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * If the data is a collection of point features, setting this to true clusters the points
+   * by radius into groups. Cluster groups become new `Point` features in the source with additional properties:
+   * - `cluster` Is `true` if the point is a cluster
+   * - `cluster_id` A unqiue id for the cluster to be used in conjunction with the
+   * [cluster inspection methods](https://www.mapbox.com/mapbox-gl-js/api/#geojsonsource#getclusterexpansionzoom)
+   * - `point_count` Number of original points grouped into this cluster
+   * - `point_count_abbreviated` An abbreviated point count
+   */
+  val cluster: Boolean = true,
```
kiryldz, line 23: Where do those defaults come from? Shouldn't it go from style spec?

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
Chaoba: Yes, the parameter must be Point

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: HashMap may be read and modified at the same time and cause `ConcurrentModificationException` error

kiryldz: Iterators must be used to handle it properly
https://stackoverflow.com/questions/602636/why-is-a-concurrentmodificationexception-thrown-and-how-to-debug-it

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
Chaoba: If users create and set the `ClusterOptions` instance, that means they want to use the cluster feature, so set the default value to true will be more user-friendly.

kiryldz: I am more asking about all other concrete values below.

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
```diff
@@ -0,0 +1,126 @@
...
+import kotlinx.android.synthetic.main.activity_add_marker_symbol.mapView
+import kotlinx.android.synthetic.main.activity_annotation.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import java.util.*
+
+/**
+ * Example showing how to add Symbol annotations
```
tobrun, line 26: Maybe include "clustering" in this title?

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
tobrun: There is no style-spec default way to style clusters themselves, we need to come up with our own values but might need to align on this with iOS

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
```diff
@@ -0,0 +1,69 @@
...
+import android.graphics.Color
+
+/**
+ * Options to show and configure symbol clustering with using SymbolManager.
+ * <p>
+ * It exposes a minimal of configuration options, a more advanced setup can be created manually with
+ * using CircleLayer and SymbolLayers directly.
+ * </p>
+ */
+data class ClusterOptions(
```
tobrun, line 12: The previous version of this file that worked with v9.0, we were able to use expressions to style. Would we be able to support this as well? https://github.com/mapbox/mapbox-plugins-android/blob/cb8ba99b74938508fe5a572b93dc75485e17afd0/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/ClusterOptions.java#L42-L49

Chaoba: This definition is in sdk-base and Expressions are in style extension, I guess we are unable to support this unless we could find a way to resolve this dependency

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: We don't have the same logic that removes items while iterating this map. It happens more like one thread is reading while one thread is modifying. Sadly I can't reproduce it anymore, use `ConcurrentHashMap` just to prevent this issue happens again.

kiryldz: @Chaoba - I checked impl and it seems it should be accessed from one thread only.
What's the potential situation when we have several threads working with this map? I guess it's pretty important because if we have concurrency in general - it's definitely not enough to just make one map concurrent, there may be other hard to catch bugs.

Chaoba: We can't prevent users from invoking different interfaces from outside at the same time unless we change all the interfaces to synchronous. For example, we [create annotations](https://github.com/mapbox/mapbox-maps-android/pull/122/files#diff-7fb27c886fce919b4fd72a2464f1b8f982933f1ec65355410969b4d7b4f8a7e1R96) from the worker thread while we may [delete annotations](https://github.com/mapbox/mapbox-maps-android/pull/122/files#diff-7fb27c886fce919b4fd72a2464f1b8f982933f1ec65355410969b4d7b4f8a7e1R72) from the UI thread.

kiryldz: I understand it all but mostly all our SDK is non thread-safe and we need to state it.
Of course we could easily call any our SDK methods from one thread and another method from another thread and most likely get some error.
My main concern here is that most likely you did not call your annotation methods from different threads 😄 Annotation plugin is designed to be non thread-safe so using `ConcurrentHashMap` seems more like a workaround for some issue.

Could that hash map be modified somehow in result callback of `queryRenderFeatures`? It uses another thread to deliver results. As I see - it's the only place where we could have another thread.

Chaoba: `queryRenderFeatures` will only be invoked with a gesture, but in the test example, we don't enable any gesture. 
I have reproduced the crash: firstly change it to normal HashMap and then click the deleteAll button quickly while opening the example.
However if change it to ConcurrentHashMap, the crash will not happen anymore. This verified ConcurrentHashMap does prevent such issues and makes the SDK more robust.

kiryldz: Let me check it by myself, I'm very curious about it.

kiryldz: @Chaoba checked your branch - this issue could be fixed if using iterators everywhere while it makes code a bit more messy or indeed use `ConcurrentHashMap` as you do. 
Let's go your way.
Another thing I mentioned - if open `SymbolClusterActivity` and quickly press delete button until nothing is loaded - all annotations will still appear on screen after they get calculated. I guess it's wrong behaviour.

pengdev:
Can we shrink the big geojson file attached with this PR( `app/src/main/assets/points.geojson `)? Could we fetch it from the web?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -183,10 +191,57 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+    }
+  }
+
+  private fun createClusterLevelLayer(level: Int, colorLevels: List<Pair<Int, Int>>) =
+    circleLayer("mapbox-android-cluster-circle-$level", sourceId) {
+      circleColor(colorLevels[level].second)
+      annotationConfig?.annotationSourceOptions?.clusterOptions?.circleRadius?.let {
+        circleRadius(it)
+      }
+      val pointCount = Expression.toNumber(get("point_count"))
```
pengdev, line 222: nit:  move to `private const val `.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -183,10 +191,57 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+          gte(pointCount, literal(colorLevels[level].first.toLong()))
+        ) else all(
+          has("point_count"),
+          gt(pointCount, literal(colorLevels[level].first.toLong())),
+          lt(pointCount, literal(colorLevels[level - 1].first.toLong()))
+        )
+      )
+    }
+
+  private fun createClusterTextLayer() = symbolLayer("mapbox-android-cluster-text", sourceId) {
```
pengdev, line 235: nit:  move to `private const val `.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
pengdev: we could expose them as expression json String. I have added a `Value.toJson()` extension function to convert the constructed Expression to json string in sdk-base.

Chaoba:
> Can we shrink the big geojson file attached with this PR( `app/src/main/assets/points.geojson `)? Could we fetch it from the web?

Maybe we could generate random points instead of reading from this file

app/src/main/res/values/example_descriptions.xml
```diff
@@ -66,6 +66,7 @@
     <string name="description_annotation_fill">Show fills on a map</string>
     <string name="description_annotation_circle">Show circles on a map</string>
     <string name="description_annotation_symbol">Show symbols on a map</string>
+    <string name="description_annotation_symbol_cluster">Show fire hydrants in Washington DC area in a cluster</string>
```
kiryldz, line 69: I guess now it shows just random icons and not true fire hydrants

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/ClusterOptions.kt
```diff
@@ -0,0 +1,70 @@
...
+   * - `point_count` Number of original points grouped into this cluster
+   * - `point_count_abbreviated` An abbreviated point count
+   */
+  val cluster: Boolean = true,
+
+  /**
+   * Radius of each cluster if clustering is enabled. A value of 512 indicates a radius equal
+   * to the width of a tile, 50 by default.
+   */
+  val clusterRadius: Long = 50,
```
kiryldz, line 29: Can we at least move all numeric constants to some file?

Chaoba: IMHO, it is more clear for users to get the default values here.

tobrun (commented):
Capturing from our talk together. Have you tried exposing expressions through the value API directly? Does that work?

app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.INTERNET" />
```
tobrun, line 9: this permission should be granted by default?

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+    symbolManager?.deleteAll()
+    runOnUiThread {
+      options?.let {
+        symbolManager?.create(it)
+      }
+      progress.visibility = View.GONE
+    }
+  }
```
tobrun, line 107: Some of this code seems to be have optimized to be calling loadData more than once but if I'm only seeing loadData being called once? Can this be simplified?

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/SymbolClusterActivity.kt
```diff
@@ -0,0 +1,160 @@
...
+      )
+    }
+    return points
+  }
+
+  private fun randomDouble(origin: Double, bound: Double): Double {
+    var r = random.nextDouble()
+    r = r * (bound - origin) + origin
+    return r
+  }
```
tobrun, line 146: is this used?

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/AnnotationUtils.kt
```diff
@@ -95,4 +95,68 @@ object AnnotationUtils {
...
+   * Load the string content from net
+   *
+   * @param context the context
+   * @param url the url of the file to load
+   */
+  fun loadStringFromNet(context: Context, url: String): String? {
+    val dir = context.cacheDir
+    val file = File("$dir/${url.hashCode()}")
+    if (file.exists()) {
+      return loadStringFromFile(file.absolutePath)
```
tobrun, line 138: I would suggest avoiding to build your own caching system from scratch.

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/AnnotationUtils.kt
```diff
@@ -95,4 +95,68 @@ object AnnotationUtils {
...
+    }
+    file.createNewFile()
+    val urlConnection: HttpURLConnection = URL(url).openConnection() as HttpURLConnection
```
tobrun, line 141: Can we use OkHttp instead?

tobrun (approved):

PR #128, opened 2021-02-24
Remove ModelLayer and ModelSource API

tobrun:
We won't be supporting this API through style extension, only though location integration.


Chaoba:
Assets files should be deleted too

pengdev:
I think the `race_car_model.gltf` is used in our test app, i.e. LocationModesActivity and LocationComponentActivity, maybe we consider keeping it in the repo.

pengdev (approved):

PR #149, opened 2021-03-04
[map] make public API entry points as MapView and Snapshotter

tobrun:
Based on mutliple user feedback, we want still want to allow same experience as with v9.0.0 of the product with having the ability to extend MapView and make a custom implementation of it.


Chaoba (approved):

kiryldz (approved):
Guess we need to add changelog entry for it.

srajala:
@tobrun This just opens the class for inheritance. None of the functions can be overridden since they are still final and not open. Was this intended?

PR #161, opened 2021-03-08
[Annotation plugin] Implement MapStyleObserverPlugin to listen style load event to reload layer and source

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[Annotation plugin] Implement MapStyleObserverPlugin to listen style load event to reload layer and source</changelog>`.

### Summary of changes
This pr implement MapStyleObserverPlugin listen in annotation plugin to prevent reload failed issue.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -8,7 +8,6 @@ import com.mapbox.geojson.Feature
 import com.mapbox.geojson.FeatureCollection
 import com.mapbox.geojson.Geometry
 import com.mapbox.geojson.Point
-import com.mapbox.maps.MapChange
 import com.mapbox.maps.RenderedQueryOptions
 import com.mapbox.maps.ScreenCoordinate
 import com.mapbox.maps.StyleManagerInterface
@@ -60,7 +59,6 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private val mapMoveResolver = MapMove()
   private var draggedAnnotation: T? = null
   private val annotationMap = mutableMapOf<Long, T>()
-  private var mapChangeListenerAdded = false
   protected var touchAreaShiftX: Int = mapView.scrollX
   protected var touchAreaShiftY: Int = mapView.scrollY
   protected abstract val layerId: String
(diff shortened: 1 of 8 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -162,7 +165,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     }
   }
 
-  protected fun initLayerAndSource() {
+  @Synchronized protected fun initLayerAndSource() {
```
kiryldz, line 168: Synchronized not needed, all those events will happen on main thread.
https://github.com/mapbox/mapbox-maps-android/blob/da2ea096429959c2159a955042d5bfc222763922/sdk/src/main/java/com/mapbox/maps/NativeMapObserver.kt#L31-L33

Chaoba: In the last commit, we got the following error on firebase, it seems we have sync issue in MapChange and getStyle callback. Maybe we could remove one to remove this `@Synchronized`, need to rethink it.

`java.lang.RuntimeException: Add source failed: Source mapbox-android-circle-source-1 already exists
	at com.mapbox.maps.extension.style.sources.Source.bindTo(Source.kt:59)
	at com.mapbox.maps.extension.style.sources.SourceKt.addSource(Source.kt:185)
	at com.mapbox.maps.plugin.annotation.AnnotationManagerImpl.initLayerAndSource(AnnotationManagerImpl.kt:177)
	at com.mapbox.maps.plugin.annotation.AnnotationManagerImpl$1$1.invoke(AnnotationManagerImpl.kt:114)
	at com.mapbox.maps.plugin.annotation.AnnotationManagerImpl$1$1.invoke(AnnotationManagerImpl.kt:42)
	at com.mapbox.maps.plugin.MapDelegateProviderImpl$getStyle$1.onStyleLoaded(MapDelegateProviderImpl.kt:20)
	at com.mapbox.maps.MapboxMap.getStyle(MapboxMap.kt:226)`

kiryldz: Would be great to reproduce the error and find the root cause but for me it seems that `synchronized` is not needed here (if looking at code) and may just find the root cause for that error even harder to detect in future.

Chaoba: Yeah, find a new solution to remove `synchronized`, and works well.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -185,6 +181,19 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     }
 
     updateSource()
+
+    // Listen to map change event to reload layer and source in the first init
+    if (!mapChangeListenerAdded) {
+      mapChangeListenerAdded = true
+      delegateProvider.mapListenerDelegate.addOnMapChangedListener {
```
kiryldz, line 188: Should this listener be cleared as others in `onDestroy`? If yes - then flag should be reset in `onDestroy` as well.

pengdev:
@Chaoba Regarding the approach, instead of adding a new map event listener, I think we could add `MapStyleObserverPlugin` interface to the AnnotationPlugin, and propagate the style changed event to the underlying annotation managers to re-add the layers/sources.  This delegate is also used to update the UI in LocationComponentPlugin.

pengdev (approved):
LGTM, please test locally as well 👍

kiryldz (approved):

PR #162, opened 2021-03-08
Use String protocol for passing GeoJSON data 

tobrun:
This PR changes how we marshal GeoJSON data across the JNI boundary. Instead of converting each Feature/Geometry to nested value types. We are taking the String representation of the GeoJSON data and wrapping that into a Value instead. Benchmark upstream has shown that this can result in a +3x improvement for ~10000 GeoJSON points. 

#### Todo: 
 - [x] back up this change with own benchmarks and track this as a metric overtime. 

cc @alexshalamov, thank you for leading out this improvement and stopgap solution.

### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
@@ -2,7 +2,6 @@ package com.mapbox.maps.extension.style.utils
 
 import com.google.gson.JsonPrimitive
 import com.mapbox.bindgen.Value
-import com.mapbox.common.ValueConverter
 import com.mapbox.geojson.Feature
 import com.mapbox.geojson.FeatureCollection
 import com.mapbox.geojson.Geometry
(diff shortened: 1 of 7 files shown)


kiryldz (approved):
Good to go after fixing tests.
Also would really value concrete numbers for concrete data how much we gain from this change.

pengdev (approved):
Agree with @kiryldz here, probably we could add some benchmark to measure this.

tobrun:
Before:
Execution time: geojson_parse_to_value: 3794

After:
Execution time: geojson_parse_to_value: 793

Using:
https://opendata.dc.gov/datasets/crime-incidents-in-2020

extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
```diff
@@ -297,31 +296,19 @@ inline fun <reified T> StylePropertyValue.silentUnwrap(): T? {
...
 /**
  * Extension function for [FeatureCollection] to convert [FeatureCollection] to [Value].
  */
 fun FeatureCollection.toValue(): Value {
-  val expected = ValueConverter.fromJson(this.toJson())
-  expected.value?.let {
-    return it
-  }
-  throw RuntimeException(expected.error)
+  return TypeUtils.wrapToValue(this.toJson())
```
alexshalamov, line 306: @tobrun I think it would be better to move this temporary optimization to `GeoJsonSource`'s feature, featureCollection, geometry methods. The `toValue` methods could be useful in some cases.

tobrun: TypeUtils is an internal class and adopted functions are only used by GeoJsonSource

Chaoba (approved):

zugaldia:
@tobrun @kiryldz When do we expect this PR to be merged? This has been reported as a P0 by a customer.

/cc: @knov

tobrun:
> @tobrun @kiryldz When do we expect this PR to be merged? This has been reported as a P0 by a customer.

When CI passes :) asap

Issue #49, opened 2021-01-28
Refactor Event API into new Observers


tobrun:
Upstream we have been refactoring the event API through the Observable interface to make the event system untyped. This will allow us long term to be more reactive to changes by expoing experimental events while not being tied down by semver. We do still need to refactor the code to remove the old system and put in place a replacement. 

For now, we will wrap the untyped system with the existing callbacks but there will be cases we're will replace specific implementations (Eg. decoupling MapChangeListener into specific interfaces), This will lead to some breaking changes.

PR #166, opened 2021-03-09
Refactor Event API into new Observers.

pengdev:
Fixes: https://github.com/mapbox/mapbox-maps-android/issues/49

<changelog>Refactor Event API into new Observers.</changelog>

## Pull request checklist:

### Summary of changes

This PR refactors events API to use the new observable API, and wrapped the untyped system with the existing callbacks.

However there are cases that we replace specific implementations (Eg. decoupling MapChangeListener into specific interfaces), This will lead to some breaking changes.

This PR also removes the extension-observable module from the project, its functionalities are moved to the sdk itself.


plugin-location/src/main/java/com/mapbox/maps/plugin/location/LocationPluginImpl.kt
```diff
@@ -1632,14 +1632,14 @@ class LocationPluginImpl : LocationPlugin {
    */
   override fun onStyleLoading() {
     Logger.d(TAG, "onStartLoadingMap")
-    onStartLoadingMap()
   }
```
tobrun, line 1635: was this an expected change?

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/listeners/OnMapLoadingFinishedListener.kt
```diff
@@ -0,0 +1,13 @@
+package com.mapbox.maps.plugin.delegates.listeners
+
+/**
+ * Definition for listener invoked when the map loading finishes.
+ */
+fun interface OnMapLoadingFinishedListener {
+
+  /**
+   * Invoked when the Map's style has been fully loaded, and the Map has rendered all visible tiles.
+
```
tobrun, line 10: nit whiteline

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -80,32 +75,33 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
   override fun getMapboxMap(): MapboxMap {
     return mapboxMap
   }
 
   override fun onStart() {
-    mapObserver.addOnCameraChangeListener(cameraChangeListener)
-    mapObserver.addOnMapChangedListener(mapChangeListener)
+    this.nativeObserver.onStart()
+    this.nativeObserver.addOnCameraChangeListener(this.onCameraChangedListener)
+    this.nativeObserver.addOnStyleLoadingFinishedListener(this.onStyleLoadingFinishedListener)
```
tobrun, line 104: should we fold this logic into onStart itself?

tobrun: if not, maybe use apply block for readability?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -115,14 +111,15 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+    this.nativeObserver.onStop()
+    this.nativeObserver.removeOnCameraChangeListener(this.onCameraChangedListener)
+    this.nativeObserver.removeOnStyleLoadingFinishedListener(this.onStyleLoadingFinishedListener)
     renderer.onStop()
     pluginRegistry.onStop()
   }
 
   override fun onDestroy() {
-    mapObserver.clearListeners()
+    this.nativeObserver.clearListeners()
```
tobrun, line 122: why explicitly using`this` in the class?

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
```diff
@@ -0,0 +1,441 @@
+package com.mapbox.maps
+
+import android.os.Handler
+import com.mapbox.maps.extension.observable.*
+import com.mapbox.maps.plugin.delegates.listeners.*
+import java.lang.ref.WeakReference
+import java.util.concurrent.CopyOnWriteArrayList
+
+internal class NativeObserver(
```
tobrun, line 9: When registering or unregistering, we are using a handler to post to make sure that happens on the main thread. Imo calls on the NativeObserver should be default be on the main thread. (that callbacks get invoked on another, is independent of this).

plugin-location/src/main/java/com/mapbox/maps/plugin/location/LocationPluginImpl.kt
pengdev: with the new observable API, we don't have the `MapChange.WILL_START_LOADING_MAP` event any more.
Since it is only used in the deprecated Location Plugin, we might just walk around it.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: introduced during the refactor, will roll back.

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
pengdev: This is added since we introduced the on demand subscription.
Since the events will be emitted on the same thread the thread that subscribe is called, I'm thinking there would be a case that user firstly adds a listener on a non-main thread and then in some other place adds a second listener on the main thread and expect the events is emitted on the main thread, but then that is not the case, since the event is still emitted to the thread that the first subscriber's thread during subscription.
To be consistent I think it would make sense to always call the subscribe on the main thread, wdyt?

plugin-location/src/main/java/com/mapbox/maps/plugin/location/LocationPluginImpl.kt
pengdev: pushed a new commit to remove the onStyleLoading callback, and wired LocationPlugin to use onStyleLoaded.

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
```diff
@@ -0,0 +1,441 @@
...
+      handler.post {
+        observable.get()?.subscribe(this, listOf(MapEvents.STYLE_IMAGE_REMOVE_UNUSED))
+      }
+      observedEvents.add(MapEvents.STYLE_IMAGE_REMOVE_UNUSED)
+    }
+    onStyleImageUnusedListeners.add(onStyleImageUnusedListener)
+  }
+
+  fun removeOnStyleImageUnusedListener(onStyleImageUnusedListener: OnStyleImageUnusedListener) {
+    onStyleImageUnusedListeners.remove(onStyleImageUnusedListener)
```
Chaoba, line 368: There are so many similar codes, maybe we could extract them into one method and use `MapEvents` as a param. For example `internalRemoveListener(onStyleImageUnusedListener, MapEvents.STYLE_IMAGE_REMOVE_UNUSED)`

And for add interface  `internalAddListener(onStyleImageUnusedListener,MapEvents.STYLE_IMAGE_REMOVE_UNUSE)` . 

We could also put all listeners into a map with `MapEvents` as the key to quick access them by key. BTW, it will only need a `forEach` in `clearListeners()`

pengdev: Moved the `subscribeNewEvent` and `unsubscribeUnusedEvent` logic out, so that the redundant code is reduced, however to avoid force type casting, I prefer to leave out the hashmap approach.

Chaoba (approved):
LGTM

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
tobrun: :+1: on only main thread requirement, @alexshalamov confirmed this.

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapListenerDelegate.kt
```diff
@@ -6,49 +6,91 @@ import com.mapbox.maps.plugin.delegates.listeners.*
...
+  // Map events
   /**
-   * Remove the map change listener.
+   * Add a listener that's going to be invoked whenever map has entered the idle state.
+   *
+   * The Map is in the idle state when there are no ongoing transitions and the Map has rendered all
+   * available tiles.
    */
-  fun removeOnMapChangedListener(onMapChangedListener: OnMapChangedListener)
+  fun addOnMapIdleListener(onMapIdleListener: OnMapIdleListener)
```
kiryldz, line 27: @tobrun do I get it correctly that for majority of benchmarks now it's better to use that listener?

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -9,22 +9,20 @@ internal object MapProvider {
...
   )
 
   fun getMapboxMap(
     nativeMap: MapInterface,
-    mapObserver: NativeMapObserver,
+    nativeObserver: NativeObserver,
     pixelRatio: Float
   ) =
-    MapboxMap(nativeMap, mapObserver, pixelRatio)
+    MapboxMap(nativeMap, nativeObserver, pixelRatio)
```
kiryldz, line 25: @pengdev could we remove `MapProvider` class at all? It's actually useless as it incapsulates nothing, just hides ctor invocations while each of that functions is used only once in project and once in test.

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
kiryldz: @pengdev just to make sure I get it all right - we use handler post for only registering / unregistering listeners while `notify` is always called from main thread and concrete listeners get notified immediately on main thread (and do not post that updates on main thread handler).

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapListenerDelegate.kt
tobrun: If your benchmark needs to include symbol placement to be fully ready, yes you can use that dedicated listener.
if that doesn't need to be tracked, there are callbacks that called earlier.

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/listeners/eventdata/MapLoadError.kt
```diff
@@ -0,0 +1,22 @@
+package com.mapbox.maps.plugin.delegates.listeners.eventdata
+
+import com.google.gson.annotations.SerializedName
+
+/** Describes an error while loading the map.  */
+enum class MapLoadError {
+  /** An error while interpreting the contents of the style.  */
+  @SerializedName("style-parse")
```
kiryldz, line 8: Do we need serialized here?

kiryldz:
@pengdev - I did not see `README` being updated for observable extension.
Also would be nice to have several examples in code of using that extension in actual examples.

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapListenerDelegate.kt
kiryldz: I see docs stating
```
  /**
   * Invoked when the Map has entered the idle state. The Map is in the idle state when there are no ongoing transitions
   * and the Map has rendered all available tiles.
   */
```
For me it seems like a classic convenient IDLE meaning that map has loaded everything (I see nothing here about symbol placement etc), no rendering will happen now and we could assume that benchmark could be ended. Thus it could be used literally everywhere as an exit point for a benchmark.
Am I wrong?

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/listeners/eventdata/MapLoadError.kt
tobrun: Yes, the data returned from gl-native is in a json format that needs to be serialized to a kotlin compatible format

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
pengdev: @kiryldz right, that is how it is expected to work.

pengdev:
> @pengdev - I did not see `README` being updated for observable extension.
> Also would be nice to have several examples in code of using that extension in actual examples.

Regarding the observable extension, I'm thinking to remove this module, since I have migrated this piece of code to sdk module, and added more observable extension functions for other events. wdyt? cc @tobrun

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
pengdev: I guess it's meant to help with testing, correct me if I'm wrong @tobrun

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapListenerDelegate.kt
pengdev: @kiryldz The events is revised again in the gl-native beta.17 release, it will include more granular control over the loading status. I will integrate the changes to this PR when the gl-native release lands.

tobrun:
> Regarding the observable extension, I'm thinking to remove this module, since I have migrated this piece of code to sdk module, and added more observable extension functions for other events in the sdk module. wdyt?

This is the right action for now, would be great to know about the binary impact it has (the model classes) but we can see that after the PR has landed. Long term we could still opt to decouple the events from the main SDK but not a priority atm.

tobrun (approved):
amazing lift @pengdev. great work!

app/src/main/java/com/mapbox/maps/testapp/examples/DebugModeActivity.kt
```diff
@@ -83,6 +85,63 @@ class DebugModeActivity : AppCompatActivity() {
...
+        TAG,
+        "OnSourceChangeListener: $it"
+      )
+    }
+    mapboxMap.addOnSourceRemovedListener {
+      Logger.i(
+        TAG,
+        "OnSourceRemovedListener: $it"
+      )
+    }
```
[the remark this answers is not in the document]
pengdev: I have verified the actions from `RuntimeStylingActivity`, but these listeners wouldn't be a good fit in that activity, so only added in the `DebugModeActivity` to showcase the usage. Long term we should add the options to trigger these events 👍

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapListenerDelegate.kt
pengdev: I think I will move forward and merge this PR first, and adapt breaking changes again together with bumping gl-native in a separate PR. This PR is getting too big for review again.

Issue #169, opened 2021-03-11
Too many data-driven properties are triggered in Annotation Plugin


Chaoba:
In the 4 annotation Option classes, most of the properties have a default value and this default value will trigger the related data-driven property which is not what we want.

The solution for this is to change the default value of these properties to null and only enable data-driven properties that are set by users.

PR #173, opened 2021-03-16
[Annotation plugin] Set default values for annotation option properties to null.

Chaoba:
Fixes: #169 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[Annotation plugin] Set default values for annotation option properties to null.</changelog>`.

### Summary of changes
This pr change the default value to null for properties in annotation options to avoid triggering to many data-driven properties.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/Circle.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/Circle.kt
@@ -66,8 +66,8 @@ class Circle(
      */
     get() {
       val value = jsonObject.get(CircleOptions.PROPERTY_CIRCLE_SORT_KEY)
-      if (!value.isJsonNull) {
-        return value.asString.toDouble()
+      value?.let {
+        return it.asString.toDouble()
       }
       return null
     }
@@ -99,8 +99,8 @@ class Circle(
      */
     get() {
       val value = jsonObject.get(CircleOptions.PROPERTY_CIRCLE_BLUR)
-      if (!value.isJsonNull) {
-        return value.asString.toDouble()
+      value?.let {
+        return it.asString.toDouble()
       }
       return null
     }
@@ -132,8 +132,8 @@ class Circle(
     @ColorInt
     get() {
       val value = jsonObject.get(CircleOptions.PROPERTY_CIRCLE_COLOR)
-      if (!value.isJsonNull) {
-        ColorUtils.rgbaToColor(value.asString)?.let {
+      value?.let {
+        ColorUtils.rgbaToColor(it.asString)?.let {
           return it
         }
       }
@@ -168,8 +168,8 @@ class Circle(
      */
     get() {
       val value = jsonObject.get(CircleOptions.PROPERTY_CIRCLE_COLOR)
-      if (!value.isJsonNull) {
-        return value.asString.toString()
+      value?.let {
+        return it.asString.toString()
       }
       return null
     }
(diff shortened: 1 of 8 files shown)


kiryldz:
@Chaoba are there any numbers how does this change affect performance in a better way?

Chaoba:
> @Chaoba are there any numbers how does this change affect performance in a better way?

In the past, since we have a default value for most properties, it will trigger too many data-driven properties. In this pr, it will only trigger data-driven properties that users set. So theoretically, this will improve the performance.

kiryldz:
> > @Chaoba are there any numbers how does this change affect performance in a better way?
> 
> In the past, since we have a default value for most properties, it will trigger too many data-driven properties. In this pr, it will only trigger data-driven properties that users set. So theoretically, this will improve the performance.

Can we measure init with previous approach and this one? To see some numbers.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/Circle.kt
```diff
@@ -66,8 +66,8 @@ class Circle(
      */
     get() {
       val value = jsonObject.get(CircleOptions.PROPERTY_CIRCLE_SORT_KEY)
-      if (!value.isJsonNull) {
-        return value.asString.toDouble()
+      value?.let {
```
kiryldz, line 69: I obviously do not see all the context here but let's imagine that `value = null` here.
Previously that check `!value.isJsonNull` was `false` and now we're also not entering `let {}` block here.
Where do we gain performance in that case?

Chaoba: In Option class, they have default values, thus `value` here will not be null and they will always pass the check.

tobrun:
> Can we measure init with previous approach and this one? To see some numbers.

The changes in this PR is not to improve performance but to not hit the hardware limitation of using many data driven properties. The amount of data driven properties you can apply on a single layer is hardware constrained due to the vertex attribute limit of the GPU.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/Circle.kt
kiryldz: I see now, thanks.

kiryldz:
> > Can we measure init with previous approach and this one? To see some numbers.
> 
> The changes in this PR is not to improve performance but to not hit the hardware limitation of using many data driven properties. The amount of data driven properties you can apply on a single layer is hardware constrained due to the vertex attribute limit of the GPU.

@tobrun thanks for clarifying!
However if user would still set all of those properties - same error will happen, do we have that documented somewhere? I guess it should be clearly stated in README.md for example.

tobrun (approved):

kiryldz (approved):

tobrun:
> However if user would still set all of those properties - same error will happen, do we have that documented somewhere? I guess it should be clearly stated in README.md for example.

Agreed, we should highlight this as a limitation of the annotation plugin as it misuses data driven styling to achieve constant values. Same applies to general style integration but this happens less frequently.

pengdev:
@Chaoba I'm wondering if this change will cause issues in this situation: when there are multiple symbols in the same symbol manager, and only one symbol is specified the data driven property, then gl-native will try to use `get` expression to get the property from the feature property, would it be null then for the symbols that hasn't specified the property?

Chaoba:
> @Chaoba I'm wondering if this change will cause issues in this situation: when there are multiple symbols in the same symbol manager, and only one symbol is specified the data driven property, then gl-native will try to use `get` expression to get the property from the feature property, would it be null then for the symbols that hasn't specified the property?

In SymbolActivity we have one symbol with `TextFont` and others not. Haven't seen any issue.

PR #178, opened 2021-03-16
[sdk] clear listeners for success/error style loading

tobrun:
This PR optimizes for the following use-case:

```
I load a style -> error occurs
I load a new style -> loaded successful
```

Currently the second action results in invoking 2 callbacks:
-  the style loaded callback of the first attempt 
-  the style loaded callback of the second attempt

This logic is encapsulated in the added test and has been optimized for. 
Additionally this PR also move awaitingStyleGetters to be made part of MapboxMap.
This an implementation detail of MapboxMap and unrelated to NativeObserver.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -51,6 +51,8 @@ class MapboxMap internal constructor(
   @VisibleForTesting(otherwise = PRIVATE)
   internal var gesturesPlugin: WeakReference<GesturesPlugin>? = null
 
+  internal val awaitingStyleGetters = mutableListOf<Style.OnStyleLoaded>()
+
   /**
    * Will load a new map style asynchronous from the specified URI.
    *
@@ -202,10 +204,10 @@ class MapboxMap internal constructor(
       onStyleLoaded?.onStyleLoaded(style)
 
       // notify style getters
-      for (styleGetter in nativeObserver.awaitingStyleGetters) {
+      for (styleGetter in awaitingStyleGetters) {
         styleGetter.onStyleLoaded(style)
       }
-      nativeObserver.awaitingStyleGetters.clear()
+      awaitingStyleGetters.clear()
     }
     onMapLoadErrorListener?.let {
       addOnMapLoadErrorListener(it)
@@ -224,11 +226,11 @@ class MapboxMap internal constructor(
         onStyleLoaded.onStyleLoaded(style)
       } else {
         // style load is occurring now, add callback
-        nativeObserver.awaitingStyleGetters.add(onStyleLoaded)
+        awaitingStyleGetters.add(onStyleLoaded)
       }
     } else {
       // no style has loaded yet, add callback
-      nativeObserver.awaitingStyleGetters.add(onStyleLoaded)
+      awaitingStyleGetters.add(onStyleLoaded)
     }
   }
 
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1184,4 +1186,8 @@ class MapboxMap internal constructor(
     }
     return null
   }
+
+  internal fun clearListeners() {
```
kiryldz, line 1190: Do we even need that function?
`awaitingStyleGetters` are also internal, we could clear them directly.

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
```diff
@@ -43,13 +41,19 @@ internal class NativeObserver(
       MapEvents.MAP_IDLE -> onMapIdleListeners.forEach { it.onMapIdle() }
       MapEvents.MAP_LOADING_ERROR -> if (onMapLoadErrorListeners.isNotEmpty()) {
         val loadingError = event.getMapLoadingErrorEventData()
+        onStyleLoadingFinishedListeners.clear()
```
kiryldz, line 44: That should work now because `onStyleLoadingFinishedListeners` and `onMapLoadErrorListeners` are managed internally by us. But if we expose it to user some day - it will not work correctly.
Can we just use single `onStyleLoadingFinishedListener` and `onMapLoadErrorListener`? In current conditions it seems array is useless.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
tobrun: This is to make sure we don't get into the following:
```
 - user startups a map and loads a style
 - exists activity before style has loaded
 - listener is leaked
```

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
tobrun: > But if we expose it to user some day

We do expose it to users and plugins already:
```
MapboxMap#addOnStyleLoadingFinishedListener
```

> it will not work correctly.

Is it because you expect a listener to get continous style load updates versus 1 shot invocation? This is indeed maybe not correct and needs to be altered to only allow Style.OnStyleLoaded to be invoked once but all `onStyleLoadingFinishedListeners` should be invoked always. Going to revisit this. 

> In current conditions it seems array is useless.

We can have multiple components listening to this (ranging from users to plugin to our internal style loaded callback system).

pengdev:
Looks good 👍 , we should also change the documentation to the add/remove listeners to these functions. Since we changed the logic to only emitting some events once.

This also makes the behaviour different from adding event listeners versus observing the event directly, we should probably document it clearly.

tobrun:
>  Since we changed the logic to only emitting some events once.

With @kiryldz remark above I realised we don't want to make this one shot event but only optimise for the callback injected with setStyle. Back to the drawing board ;)

pengdev:
Note that the style loading event has changed in the gl-native beta.17 release, we should rebase this PR after the integration is merged.

pengdev:
> >  Since we changed the logic to only emitting some events once.
> 
> With @kiryldz remark above I realised we don't want to make this one shot event but only optimise for the callback injected with setStyle. Back to the drawing board ;)

Yes, we shouldn't change the logic inside the Native observer, but optimize the setstyle callback logic in the MapboxMap class.

tobrun:
with change it requirements and stale code, following up on this in #300

PR #184, opened 2021-03-18
Patch scroll gesture with a pitched camera

tobrun:
This is not a true solution, just an intermediate patch to make the experience usable. Todo comment inline in the code indicates what true solution  would look like, calculating the y offset based on triangulation, but will require exposing some functionality in the C++ core to convert zoom level to a pixel value.


kiryldz (approved):
Nice!
The easier the better 😃

Issue #183, opened 2021-03-17
Location component: possible jitter while animating


kiryldz:
## Environment

- Android OS version: any
- Devices affected: any
- Maps SDK Version: android-v10.0.0-beta.15

## Observed behavior and steps to reproduce

If using location component any type of animation (let's say position) and interrupt it with animation of same type - puck will perform an immediate jump to end value of interrupted animation.

## Expected behavior

Last save animator value should be always in sync with ongoing animation (e.g. if animation is cancelled - start value for next animation should be the last interpolated one).

PR #185, opened 2021-03-18
[locationcomponent] Fix jitter animations if interrupting animations

kiryldz:
Fixes: #183 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[locationcomponent] Fix jitter animations if interrupting animations</changelog>`.

### Summary of changes

Fix jitter animations if interrupting animations.
Such jittering happened before if location (or bearing) updates were arriving not evenly and animations cancelled each other making new animation start not from actual interpolated value but from end value of interrupted animation.

### DIFF
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
@@ -14,6 +14,7 @@ internal abstract class PuckAnimator<T>(
   evaluator: TypeEvaluator<T>
 ) : ValueAnimator() {
 
+  private var puckUpdateListener: ((T) -> Unit)? = null
   @VisibleForTesting(otherwise = PRIVATE)
   internal var userConfiguredAnimator: ValueAnimator
   protected var locationRenderer: LocationLayerRenderer? = null
@@ -25,7 +26,9 @@ internal abstract class PuckAnimator<T>(
     setEvaluator(evaluator)
     addUpdateListener {
       @Suppress("UNCHECKED_CAST")
-      updateLayer(it.animatedFraction, it.animatedValue as T)
+      val updatedValue = it.animatedValue as T
+      updateLayer(it.animatedFraction, updatedValue)
+      puckUpdateListener?.invoke(updatedValue)
     }
     duration = LocationComponentConstants.DEFAULT_INTERVAL_MILLIS
     interpolator = DEFAULT_INTERPOLATOR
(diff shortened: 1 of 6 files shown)


plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
```diff
@@ -25,7 +26,9 @@ internal abstract class PuckAnimator<T>(
     setEvaluator(evaluator)
     addUpdateListener {
       @Suppress("UNCHECKED_CAST")
-      updateLayer(it.animatedFraction, it.animatedValue as T)
+      val updatedValue = it.animatedValue as T
+      updateLayer(it.animatedFraction, updatedValue)
+      puckUpdateListener?.invoke(updatedValue)
```
Chaoba, line 31: Should `puckUpdateListener` be reset when an animation is complete?

kiryldz: No, it actually should not matter.
All this API is internal and actually only thing we need is proxy updated value up to `LocationPuckManager` to always have latest value there.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
```diff
@@ -69,13 +72,13 @@ internal abstract class PuckAnimator<T>(
...
-    if (userConfiguredAnimator.isRunning) {
-      userConfiguredAnimator.cancel()
-    }
+  fun animate(
+    vararg targets: T,
+    onUpdate: ((T) -> Unit)? = null,
+    options: (ValueAnimator.() -> Unit)? = null
+  ) {
+    cancelRunning()
+    puckUpdateListener = onUpdate
```
pengdev, line 81: seems we only have one puckUpdateListener at a time, and it will be overwritten every time the new animate call is invoked. How about exposing the onPuckUpdatedListener directly from the PuckAnimator? so that the LocationPuckManager can just register `onPuckUpdatedListener` once and use the same listener for all the updates.

kiryldz: Sounds reasonable. 
My idea was here that we do not need it in each but it seems we do.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
kiryldz: @Chaoba reworked approach, PTAL

Chaoba (approved):

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
```diff
@@ -65,17 +69,21 @@ internal abstract class PuckAnimator<T>(
     return super.clone()
   }
 
+  fun setUpdateListener(updateListener: ((T) -> Unit)) {
+    if (this.updateListener != updateListener) {
```
pengdev, line 73: why not directly set the updateListener? Do we actually call the setUpdateListener multiple times?

pengdev (approved):
LGTM with one question.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimator.kt
kiryldz: Yes, via `updateSettings` we call it multiple times. Instead of checking there to do it only once I incapsulated it here.

Issue #182, opened 2021-03-17
Expose utility functions for CoordinateBounds


kiryldz:
## New Feature

Expose functions: `getCenter`, `contains`, `union`, `intersect`, `containsLatitude`, `containsLongitude`, `include`, `getLatitudeSpan`, `getLongitudeSpan` as extension functions for `CoordinateBounds`.

Full definition of `LatLngBounds` (ancestor for `CoordinateBounds`) could be found [here](https://github.com/mapbox/mapbox-gl-native-android/blob/main/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java).

## Why

Provide easier migration from previous maps version, improve parity.

pengdev:
seems duplicated with https://github.com/mapbox/mapbox-maps-android/issues/131

kiryldz:
That's available for testing in [v10.2.0-beta.1](https://github.com/mapbox/mapbox-maps-android/releases/tag/android-v10.2.0-beta.1).

PR #188, opened 2021-03-18
[sdk] Add some utility functions to CoordinateBounds

kiryldz:
Fixes: #182 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add some utility functions to CoordinateBounds.</changelog>`.

### Summary of changes

Add some utility functions to CoordinateBounds: `getCenter`, `union`, `contains`.

### DIFF
--- a/sdk-base/src/main/java/com/mapbox/maps/ExtensionUtils.kt
+++ b/sdk-base/src/main/java/com/mapbox/maps/ExtensionUtils.kt
@@ -4,6 +4,8 @@ import android.graphics.Bitmap
 import com.mapbox.bindgen.Value
 import com.mapbox.common.ValueConverter
 import com.mapbox.geojson.Point
+import com.mapbox.geojson.constants.GeoJsonConstants.MAX_LATITUDE
+import com.mapbox.geojson.constants.GeoJsonConstants.MIN_LATITUDE
 import java.nio.ByteBuffer
 import kotlin.math.abs
 import kotlin.math.max
@@ -49,9 +51,9 @@ fun MapSnapshotInterface.bitmap(): Bitmap {
  * @return [Point] center of this [CoordinateBounds]
  */
 fun CoordinateBounds.getCenter(): Point {
+  checkBounds()
   val latCenter = (northeast.latitude() + southwest.latitude()) / 2.0
   val lngCenter = (northeast.longitude() + southwest.longitude()) / 2.0
-
   return Point.fromLngLat(lngCenter, latCenter)
 }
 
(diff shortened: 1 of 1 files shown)


Chaoba (approved):

sdk-base/src/main/java/com/mapbox/maps/ExtensionUtils.kt
```diff
@@ -37,4 +40,59 @@ fun MapSnapshotInterface.bitmap(): Bitmap {
...
+    Point.fromLngLat(
+      min(this.southwest.longitude(), bounds.southwest.longitude()),
+      min(this.southwest.latitude(), bounds.southwest.latitude())
+    ),
+    // northeast
+    Point.fromLngLat(
+      max(this.northeast.longitude(), bounds.northeast.longitude()),
+      max(this.northeast.latitude(), bounds.northeast.latitude())
+    )
+  )
```
pengdev, line 97: can you also add tests?

kiryldz: I was hoping you will not notice it 😢 
Yep, will do

kiryldz: @pengdev - we don't even have `test` package for `sdk-base`.
Assuming current solution is a bit of workaround while we land better solution upstream - perhaps we could omit tests in that case?

pengdev: Since these functions involve some complex logic, I would suggest to add some unit tests to verify the corner cases. A few validation should be enough, especially around the border of the bounds and across the international date line. The support of wrapped and unwrapped coordinates should also be taken into consideration, and maybe state it clearly in the documentation.

kiryldz:
Will be fixed later upstream.

tobrun:
Capturing from upstream that I won't be feasible adding this functionality so we will continue with the approach here and add equivalent functionality to iOS

tobrun:
@kiryldz you mentioned only supporting a part of the LatLngBounds functionality, what is missing?
Can you create a ticket on iOS side and indicate what they need to add? (they should be able to also use extension functions)

cvb941:
What's the status on this?

tobrun:
> What's the status on this?

We are patching and exposing the APIs upstream. This has the benefit we only need to maintain one codebase versus Android/iOS specifically. It will be coming in one of the next releases.

Issue #156, opened 2021-03-05
[rendering] Expose function to schedule runnable on GL thread


kiryldz:
## New Feature

Expose two methods for render thread: `queueGlEvent` and `queueEvent` instead of one now.

## Why
It may be useful in some cases to split putting some `Runnable` in GL thread event queue and simply running it and putting that `Runnable` (most likely with GL commands) to a queue, executing them and performing buffer swap to see result on screen. Current implementation of `queueEvent` always performs drawing after executing `Runnable`.

kiryldz:
As discussed offline with @alexshalamov we could generally have 2 initial approaches:
 1. When we get some new `Runnable` that has come from `queueEvent` or core `scheduleTask`(and thus does not require actually posting a `doFrame` via Choreographer with drawing and swapping buffers) - we could add them to our own queue created in renderer and run all those runnable __after__ next VSYNC update (requested by `queueGlEvent` or `scheduleRepaint`).
 2. Simply post a `Runnable` on a render thread and execute is ASAP (without swapping buffers).

I am more to the first approach, @mapbox/maps-android WDYT? Although I'll check out both of them with some concrete numbers.

tobrun:
> I am more to the first approach

same

> I'll check out both of them with some concrete numbers.

that would be amazing

PR #192, opened 2021-03-19
[rendering] Differentiate render tasks and non-render tasks

kiryldz:
Fixes: #156 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[rendering] Differentiate render tasks and non-render tasks</changelog>`.

### Summary of changes

Renderer now could execute basically 2 types of events on render (GL) thread. One does require true drawing (swapping buffers) and another simply executes `Task` or `Runnable` on render thread without actually swapping buffers.

### User impact (optional)

`MapView` has updated API for `queueEvent`:
```
queueEvent(event: Runnable, needRender: Boolean)
```

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapControllable.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapControllable.kt
@@ -41,11 +41,12 @@ interface MapControllable {
   fun onSizeChanged(w: Int, h: Int)
 
   /**
-   * Called to queue an event on the render thread.
+   * Queue a runnable to be executed on the map renderer thread.
    *
-   * @param runnable the runnable to be queued
+   * @param event the runnable to queue
+   * @param needRender if we should force redraw after running event (e.g. execute some GL commands)
    */
-  fun queueEvent(runnable: Runnable)
+  fun queueEvent(event: Runnable, needRender: Boolean = true)
 
   /**
    * Called to capture a snapshot synchronously.
--- a/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
@@ -9,7 +9,8 @@ import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import com.mapbox.common.Logger
 import com.mapbox.maps.renderer.egl.EGLCore
-import java.util.LinkedList
+import java.util.concurrent.CopyOnWriteArrayList
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent.locks.ReentrantLock
 import javax.microedition.khronos.egl.EGL10
@@ -34,7 +35,8 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   private val createCondition = lock.newCondition()
 
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  internal val eventQueue = LinkedList<Runnable>()
+  internal val renderEventQueue = CopyOnWriteArrayList<Runnable>()
+  private val eventQueue = CopyOnWriteArrayList<Runnable>()
 
   private var surface: Surface? = null
   private var eglSurface: EGLSurface? = EGL10.EGL_NO_SURFACE
@@ -45,8 +47,8 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   internal val renderTimeNs = AtomicLong(0)
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   private var needRenderOnResume = false
-  private var previousVsyncFrameStartTimeNs = 0L
   private var expectedVsyncWakeTimeNs = 0L
+  private var awaitingNextVsync = AtomicBoolean(false)
   private var sizeChanged = false
   private var paused = false
   private var shouldExit = false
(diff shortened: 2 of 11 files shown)


sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -34,7 +35,8 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   private val createCondition = lock.newCondition()
 
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  internal val eventQueue = LinkedList<Runnable>()
+  internal val renderEventQueue = CopyOnWriteArrayList<Runnable>()
+  private val eventQueue = CopyOnWriteArrayList<Runnable>()
```
Chaoba, line 39: There are many adding operation for these two queues,  `CopyOnWriteArrayList` might get a lower performance.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -199,15 +201,18 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
-    eventQueue.poll()?.let {
-      it.run()
-      // assuming runnable will most likely contain GL commands or something else rendering related
-      // we post draw call not considering `requestRender` flag
+    renderEventQueue.apply {
+      if (isNotEmpty()) {
+        forEach {
+          it.run()
+        }
+        clear()
```
Chaoba, line 209: This will clear the events that are added during the iteration.

sdk/src/main/java/com/mapbox/maps/MapControllable.kt
```diff
@@ -41,11 +41,12 @@ interface MapControllable {
...
   /**
-   * Called to queue an event on the render thread.
+   * Queue a runnable to be executed on the map renderer thread.
    *
-   * @param runnable the runnable to be queued
+   * @param event the runnable to queue
+   * @param needRender if we should force redraw after running event (e.g. execute some GL commands)
    */
-  fun queueEvent(runnable: Runnable)
+  fun queueEvent(event: Runnable, needRender: Boolean = true)
```
tobrun, line 49: :+1: , like the signature and default value setup

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -199,15 +201,18 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       return
     }
     checkSurfaceSizeChanged()
-    eventQueue.poll()?.let {
-      it.run()
-      // assuming runnable will most likely contain GL commands or something else rendering related
-      // we post draw call not considering `requestRender` flag
+    renderEventQueue.apply {
+      if (isNotEmpty()) {
```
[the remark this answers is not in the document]
tobrun: With normal list that is correct. With `CopyOnWriteArrayList` it does create an iterator for an empty list and thus allocating memory when it shouldn't. :+1: on keeping this check with the CopyOnWrite impl.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -272,30 +281,50 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
-      previousVsyncFrameStartTimeNs = frameTimeNanos
-      draw()
+    try {
+      awaitingNextVsync.set(false)
+      // if frame started being rendered time is less or equal to previous one
+      // it means we did not have time to render everything on previous frame
+      // we do no drawing in order not to overload render thread
+      if (eglPrepared && !paused && !shouldExit) {
+        draw()
+      }
```
tobrun, line 291: The comment above the block seems outdated

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -272,30 +281,50 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+    } finally {
+      // regardless if we did drawing or not execute all non gl relative events
+      eventQueue.apply {
+        if (isNotEmpty()) {
+          forEach {
+            it.run()
+          }
+          clear()
+        }
+      }
```
tobrun, line 301: nit: duplicate code for iterating a runnable queue and running it, could be refactored into 1 function that takes either the `eventQueu` or `renderEventQueue`

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: Despite we need those queues to be thread-safe, I'll rework an approach a bit.

tobrun: What is the use-case for copy on write? are we mutating these objects from multiple threads? or is it to support a removing while in-use setup?

tobrun: What is the use-case for copy on write? are we mutating these objects from multiple threads? or is it to support a removing while in-use setup?

edit: you already replied in the meantime above :) can we avoid access from multiple threads? I'm conceptually ok with using thread safe components to not push the responsibility of synchronization on us.

kiryldz: Yes, I tried an approach with using mutex but it looks clumsy.

tobrun (approved):

PR #202, opened 2021-03-23
[animations] Fix interpolation for flyTo

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[animations] Fix interpolation for flyTo</changelog>`.

### Summary of changes

We were using regular pixels instead of dpi that actually made zoom and center interpolation not work as designed.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
@@ -318,9 +318,11 @@ class CameraAnimatorsFactory internal constructor(mapDelegateProvider: MapDelega
     // w₀: Initial visible span, measured in pixels at the initial scale.
     // Known henceforth as a <i>screenful</i>.
 
+    val size = mapTransformDelegate.getSize()
+    val pixelRatio = mapTransformDelegate.getMapOptions().pixelRatio
     val w0 = max(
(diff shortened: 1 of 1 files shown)


tobrun (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
```diff
@@ -318,9 +318,11 @@ class CameraAnimatorsFactory internal constructor(mapDelegateProvider: MapDelega
     // w₀: Initial visible span, measured in pixels at the initial scale.
     // Known henceforth as a <i>screenful</i>.
 
+    val size = mapTransformDelegate.getMapOptions().size!!
```
pengdev, line 321: we were using `mapTransformDelegate.getSize()` and it seems we don't have to use `!!`, is there any difference with these two methods?

kiryldz: We have because size is for some reason marked `Nullable`.
There is no difference I assume, the one we used before was the one I exposed myself (most likely I just did not notice that we already have size as part of `MapOptions` or it was not exposed at that point there).
But now it's more logical to use the one from `MapOptions` because I need pixel ratio as well.

pengdev (approved):

Issue #191, opened 2021-03-19
MapView#onStop blocks benchmark execution 


tobrun:
After integration the https://github.com/mapbox/mapbox-maps-android/commit/66e473f42142994813f0e39e16504e70775eb036 in a downstream benchmark project. I'm noticing that our tests don't finish. This is probably related to the scheduler rework we did upstream.

```
    fun finishTest() {
        mapView.onStop()
        resumeTestExecution()
    }
```

The `resumeTestExecution` is never called and seem that `mapView#onStop` results in a deadlock. 

cc @kiryldz

PR #204, opened 2021-03-23
[test] add integration testing of MapView through fragments

tobrun:
We recently had a regression with #191. This wasn't noticed directly in this repository but only in a downstream project we use for running benchmarks. To make sure we don't regress these use-cases, this PR adds a bunch of fragment integration tests. 

2 files with tests were added:
 - **FragmentTest**: contains traditional testing of fragments
 - **FragmentScenarioTest**: contains tests using the [fragment test library](https://developer.android.com/guide/fragments/test)

### DIFF
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/integration/fragment/FragmentTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/integration/fragment/FragmentTest.kt
@@ -1,0 +1,133 @@
+package com.mapbox.maps.testapp.integration.fragment
+
+import android.R
+import android.widget.FrameLayout
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentTransaction
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import com.mapbox.maps.testapp.EmptyActivity
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+/**
+ * Integration test for fragments
+ */
+class FragmentTest {
+
+  @get:Rule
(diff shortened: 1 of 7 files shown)


app/src/androidTest/java/com/mapbox/maps/testapp/integration/fragment/MapFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import com.mapbox.maps.MapView
+import com.mapbox.maps.Style
+
+class MapFragment : Fragment() {
```
kiryldz, line 11: Can we somehow use same `MapFragment` as in our examples? They look pretty alike and we could make it single source of truth.

app/src/androidTest/java/com/mapbox/maps/testapp/integration/fragment/FragmentTest.kt
```diff
@@ -0,0 +1,133 @@
...
+    val fragment = MapFragment()
+    val fragmentTransaction: FragmentTransaction =
+      activity.supportFragmentManager.beginTransaction()
+    fragmentTransaction.add(R.id.content, fragment)
+    fragmentTransaction.commit()
+    return fragment
+  }
+
+  companion object {
+    const val TAG = "foo"
```
kiryldz, line 131: Perhaps `FRAGMENT_TAG` to not mess up with logger tags 😄

kiryldz (approved):
Nice approach!
Generally LGTM as it is to :ship: , dropped couple of questions.

Chaoba (approved):

app/src/androidTest/java/com/mapbox/maps/testapp/integration/fragment/MapFragment.kt
tobrun: I actually like the idea of having 2 different implementations. The test app variant calls onDestroy in onDestroyView while this impl uses onDestroy instead.

PR #224, opened 2021-03-29
Track public API with metalava

tobrun:
This PR integrates [metalava tooling](https://android.googlesource.com/platform/tools/metalava/) into our project. We will be using this to prevent us from introducing semver breaking changes to our public API. This PR follows the setup conform to the [navigation SDK](https://github.com/mapbox/mapbox-navigation-android). This PR adds SuppressWarnings checks to places in code that metalava fails the build for. 

Metalava produces a textfile that list out the public API for a give module under `/api/current.txt` folder of each user facing module. 

```
// Signature format: 3.0
package com.mapbox.maps.module.telemetry {

  @com.mapbox.annotation.module.MapboxModule(type=com.mapbox.annotation.module.MapboxModuleType) public final class MapTelemetryImpl implements com.mapbox.maps.module.MapTelemetry {
    ctor public MapTelemetryImpl(android.content.Context appContext, String accessToken);
    ctor @VisibleForTesting public MapTelemetryImpl(com.mapbox.android.telemetry.MapboxTelemetry telemetry, android.content.Context appContext, String accessToken);
    method public void disableTelemetrySession();
    method public void onAppUserTurnstileEvent();
    method public void onPerformanceEvent(android.os.Bundle? data);
    method public void setDebugLoggingEnabled(boolean debugLoggingEnabled);
    method public boolean setSessionIdRotationInterval(int interval);
    method public void setUserTelemetryRequestState(boolean enabled);
  }

}
```

Tailwork: validate how we can incorporate into CI and how we can allow breaking changes from time-to-time when needed (this work is postponed until we have hit 10.0.0 final).


kiryldz (commented):
I think it's great initial start, amazing work @tobrun!

However what are next steps here? 
Do we need to check out all the APIs and fix it in this PR? 
Let's say we changed API during next release cycle - is there a way to see delta with previous one?

tobrun:
> However what are next steps here?
Do we need to check out all the APIs and fix it in this PR?
Let's say we changed API during next release cycle - is there a way to see delta with previous one?

We can use the following command to verify if the api files need updating:

```
make check-api
```

this will flag if we are breaking APIs (but also if we are adding APIs). The idea is that we want to be very intentional with this practice. 

The next steps are to see if we can integrate this as part of our CI, maybe as part of our `make check` command. 
We would be supporting a similar flow as with our license file generation / required updating.
I'm not sure how well this practice scales and it might become a burden pretty fast.

Long term, when we create a new release, we can run a script that makes a diff between two different commits. This allows us to quantify (breaking) changes between versions.

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: check
 check:
-	./gradlew ktlint lint;
+	./gradlew ktlint lint checkApi;
```
tobrun, line 3: remove checkApi

tobrun:
Capturing from discussion internally, we will now only commit and merge configuration. The generated interfaces files and CI integration will land once we hit a more stable API surface.

kiryldz (commented):
@tobrun can you rebase on latest main?

kiryldz (approved):

PR #230, opened 2021-04-01
Add CredentialsManager and MapInitOptions to handle access token and init process for MapView

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add CredentialsManager and MapInitOptions to handle access token and init process for MapView </changelog>`.

### Summary of changes
This pr add CredentialsManager to handle access tokens


kiryldz:
@Chaoba can you please reference this PR in an appropriate internal task?

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
+package com.mapbox.maps
+
+import android.content.Context
+
+/**
+ * Convenience class that holds MapboxMaps related token.
```
kiryldz, line 6: Can we use a reference to `MapboxMap` here instead of MapboxMaps?

kiryldz: And I guess as per our infrastructure it is more connected with `MapView` and not `MapboxMap`

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+ * It's possible to create `CredentialsManager` instances as you need them,
+ * however it's convenient to use the default `CredentialsManager.shared`.
```
kiryldz, line 8:
```suggestion
 * `CredentialsManager` could be created per `MapView` instance if given view should use some specific token.
 * Most common use-case when all `MapView`s should use same token could be handled by using `CredentialsManager.shared` static object. 
```

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @property accessToken the token will be applied.
+ */
+data class CredentialsManager(val accessToken: String? = null) {
+  /**
+   * Get the access token.
+   * If the supplied token is nil (which is the case for the `shared`),
```
kiryldz, line 15:
```suggestion
   * If the supplied token is null (which is the use-case for the `shared`),
```

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+   * then we will search for an access token in the application resource.
```
kiryldz, line 16:
```suggestion
   * then we will search for an access token in the application resources.
```
also we could add a link where token is located in resources, I think we have it on Mapbox site.

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+   */
+  fun getAccessToken(context: Context): String {
+    accessToken?.let {
+      // User have set the token, use this one.
```
kiryldz, line 20:
```suggestion
      // User has set the token, use it.
```

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    // Check in the resources
+    val tokenResId = context.resources.getIdentifier(
+      MAPBOX_ACCESS_TOKEN_RESOURCE_NAME,
+      "string",
+      context.packageName
+    )
+
+    return if (tokenResId != 0) context.getString(tokenResId) else ""
```
kiryldz, line 31: Why are we returning ""? I guess we need to throw runtime exception, we can't work without token.

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Static variables and methods.
+   */
+  companion object {
+    /**
+     * The default shared instance with empty assess token.
```
kiryldz, line 39: Although it's empty - we need to specify that token will be taken from resources.

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
Chaoba: We have the check in `ResourcesAttributeParser`

sdk/src/androidTest/java/com/mapbox/maps/MapboxMapIntegrationTest.kt
```diff
@@ -34,7 +34,11 @@ class MapboxMapIntegrationTest {
   @UiThreadTest
   @Test
   fun testGetResourceOptions() {
-    val defaultOptions = MapboxOptions.getDefaultResourceOptions(mapView.context)
+    val defaultOptions = MapboxMapOptions(
+      mapView.context,
+      mapView.context.resources.displayMetrics.density,
+      null
```
kiryldz, line 40: Perhaps I'm crazy but how can null be passed here when we have non-nullable parameter of `CredentialsManager` in `MapboxMapOptions` ctor?

sdk/src/test/java/com/mapbox/maps/CredentialsManagerTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  @Before
+  fun setUp() {
+    every { context.resources } returns resources
+    every { context.packageName } returns "com.mapbox.maps"
+    every { resources.getIdentifier(any(), any(), any()) } returns 0
+  }
+
+  @Test
+  fun default() {
```
kiryldz, line 27: I'd say it's not default but error behavior when we don't have token at all

sdk/src/androidTest/java/com/mapbox/maps/MapboxMapIntegrationTest.kt
Chaoba: It's for `val attrs: AttributeSet? = null`

kiryldz: it's null by default, can we remove this `null` then?

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
kiryldz: For better readability - can we move "" to constant like `ERROR_NO_TOKEN_FOUND` and use it everywhere instead of ""?

app/src/main/java/com/mapbox/maps/testapp/examples/LegacyOfflineActivity.kt
```diff
@@ -56,7 +56,7 @@ class LegacyOfflineActivity : AppCompatActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_offline)
-    offlineManager = OfflineRegionManager(MapboxOptions.getDefaultResourceOptions((this)))
+    offlineManager = OfflineRegionManager(MapboxMapOptions(this, resources.displayMetrics.density, null).resourceOptions)
```
pengdev, line 59: I would suggest we provide a defaultResourceOption getter to the credentials manager.

app/src/main/java/com/mapbox/maps/testapp/examples/snapshotter/DataDrivenMapSnapshotterActivity.kt
```diff
@@ -32,7 +32,7 @@ class DataDrivenMapSnapshotterActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     val snapshotMapOptions = MapSnapshotOptions.Builder()
       .size(Size(512.0f, 512.0f))
-      .resourceOptions(MapboxOptions.getDefaultResourceOptions(this))
+      .resourceOptions(MapboxMapOptions(this, resources.displayMetrics.density, null).resourceOptions)
```
pengdev, line 35: It shouldn't be necessary to create a `MapboxMapOptions` to get the `resourceOptions`.

app/src/main/java/com/mapbox/maps/testapp/examples/LegacyOfflineActivity.kt
```diff
@@ -56,7 +56,7 @@ class LegacyOfflineActivity : AppCompatActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_offline)
-    offlineManager = OfflineRegionManager(MapboxOptions.getDefaultResourceOptions((this)))
+    offlineManager = OfflineRegionManager(MapInitOptions.getDefaultResourceOptions(this))
```
pengdev, line 59: would it make sense to move the `getDefaultResourceOptions(context)` to CredentialsManager?

pengdev: cc @julianrex since this is still not aligned with iOS.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -0,0 +1,59 @@
...
+data class MapInitOptions constructor(
+  val context: Context,
+  var resourceOptions: ResourceOptions = getDefaultResourceOptions(context),
+  var mapOptions: MapOptions = getDefaultMapOptions(context)
+) {
+
+  /**
+   * Camera options when using a MapView.
+   */
+  var cameraOptions: CameraOptions? = null
```
pengdev, line 26: we could rename it to initialCameraOptions and move it to the constructor?

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Flag indicating to use a TextureView as render surface for the MapView. Default is false.
+   */
+  var textureView: Boolean = false
```
pengdev, line 31: can be moved to constructor(with default value)?

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -0,0 +1,59 @@
...
+ * using custom XML tags.
+ *
+ * @property context the context of the application.
+ * @property resourceOptions Resource options when using a MapView. Access token required when using a Mapbox service. Please see [https://www.mapbox.com/help/create-api-access-token/](https://www.mapbox.com/help/create-api-access-token/) to learn how to create one.More information in this guide [https://www.mapbox.com/help/first-steps-android-sdk/#access-tokens](https://www.mapbox.com/help/first-steps-android-sdk/#access-tokens).
+ * @property mapOptions Describes the map options value when using a MapView.
+ */
+data class MapInitOptions constructor(
+  val context: Context,
+  var resourceOptions: ResourceOptions = getDefaultResourceOptions(context),
+  var mapOptions: MapOptions = getDefaultMapOptions(context)
```
pengdev, line 20: I think we also need pixel ratio here, so that it can overwrite the one from context

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -63,45 +63,32 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
+   * Build a [MapView] with a [Context] object, a [MapInitOptions] object,
    * a [AttributeSet] object, an [Int] which represents a style attribute file,
    * and an [Int] which represents a style resource file.
    */
   @Suppress("UNCHECKED_CAST")
   @TargetApi(Build.VERSION_CODES.LOLLIPOP)
   constructor(
     context: Context,
-    options: MapboxMapOptions,
+    initOptions: MapInitOptions?,
```
pengdev, line 81:
```suggestion
    initOptions: MapInitOptions = MapInitOptions(context),
```

app/src/main/java/com/mapbox/maps/testapp/examples/MapViewCustomizationActivity.kt
```diff
@@ -25,28 +25,30 @@ class MapViewCustomizationActivity : AppCompatActivity() {
...
+      )
+      .build()
+
+    // set token and cache size for this particular map view
+    val resourceOptions = ResourceOptions.Builder()
+      .accessToken(getString(R.string.mapbox_access_token))
+      .cacheSize(75_000L)
+      .build()
+
+    val mapboxMapOptions = MapInitOptions(this, resourceOptions, mapOptions).apply {
```
pengdev, line 44: we can make cameraOptions and textureView as part of `MapInitOptions`'s constructor.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
Chaoba: Do you mean make pixelRatio as a param for MapInitOptions?

app/src/main/java/com/mapbox/maps/testapp/examples/MapViewCustomizationActivity.kt
Chaoba: Right

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
pengdev: yeah, something like that, we are also parsing the pixel ratio from the xml, right? It would not be obvious if the user can set it manually

app/src/main/java/com/mapbox/maps/testapp/examples/LegacyOfflineActivity.kt
pengdev: It sounds reasonable to me, but need to align with iOS on this.

Chaoba: Just checked the [codes](https://github.com/mapbox/mapbox-maps-ios/pull/254/files#diff-46cda2fdea49d4ea82517a0b932c27ede2548397ba7e735ecc6798c44095b130R27) for iOS, it seems there is no such API

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
Chaoba: That's right. But we must make a clear doc to tell users it's only the default value, will not take effect if they use a custom MapOptions instance here.

pengdev: thinking again, I think user could still specify the pixel ratio manually by building the MapOptions themselves.. @kiryldz do you have thoughts here?

app/src/main/java/com/mapbox/maps/testapp/examples/LegacyOfflineActivity.kt
pengdev: we had similar conversation [here](https://github.com/mapbox/mapbox-maps-ios/pull/254/files#r611443274), let's discuss and create follow-up PR for iOS cc @julianrex

app/src/main/java/com/mapbox/maps/testapp/examples/MapViewCustomizationActivity.kt
```diff
@@ -62,7 +63,7 @@ class MapViewCustomizationActivity : AppCompatActivity() {
 
   private fun configureMapViewFromXml() {
     // let's set `custom` token to MapView from code (however it will be same token from resources so that map will work)
-    MapboxOptions.setDefaultResourceOptions(this, getString(R.string.mapbox_access_token))
+    CredentialsManager.shared = CredentialsManager(getString(R.string.mapbox_access_token))
```
kiryldz, line 66: By the way it looks more like there was a bug: method is named `configureMapViewFromXml` and we're setting `CredentialsManager.shared` when actually map view from xml is already inflated with some token.
Can we rebuild that example showcasing following scenario:
 - map view from xml uses default token (it goes from resources)
 - map view created programatically in `configureMapViewFromCode` uses some custom token

kiryldz: Actually, we're already doing it via `.accessToken(getString(R.string.mapbox_access_token))` when creating `resourceOptions`. Then I guess we should either remove this line or use it __before__ `setContentView` and state clearly in comment that we're overriding default `CredentialsManager.shared` meaning all map views inflated from XML will use token coming not from resources file.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
kiryldz: @pengdev why would we actually need to override pixel ratio from context as part of default options? I know we have issues if accessing pixel ratio from static instances in case of multi-display but what's the use-case when pixle ratio taken from concrete context is not correct?

kiryldz: Another potential problem I see is if we override it here and pass to core but will rely from pixel ratio from context in some other methods.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Static methods
+   */
+  companion object {
+    /**
+     * Get a default [ResourceOptions] with token from default [CredentialsManager]
+     * @property context the context of the application.
+     */
+    fun getDefaultResourceOptions(context: Context): ResourceOptions {
```
kiryldz, line 35: nit: could be `fun getDefaultResourceOptions(context: Context) = ResourceOptions.Builder()...`

kiryldz (approved):
LGTM for me.
It all seems pretty logical in general for me and globally aligned with iOS as well (although still have some differences).
If we missed something - imho we could iterate.

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
```diff
@@ -63,16 +65,16 @@ MapPluginRegistry(private val mapDelegateProvider: MapDelegateProvider) {
       if (instance is ViewPlugin) {
         val pluginView = instance.bind(
           mapView!!,
-          mapboxMapOptions.attrs,
-          mapboxMapOptions.pixelRatio
+          attrs,
```
pengdev, line 68: one question: the attrs was a part of the MapboxMapOptions, would it make sense to move it to the MapInitOptions? I also do see some benefits to separate these.

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -69,39 +69,25 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
-    options: MapboxMapOptions,
     attrs: AttributeSet?,
     defStyleAttr: Int,
-    defStyleRes: Int
+    defStyleRes: Int,
+    initOptions: MapInitOptions = MapInitOptions(context),
   ) : super(context, attrs, defStyleAttr, defStyleRes) {
-    val view = if (options.textureView) {
+    val resolvedMapInitOptions = if (attrs != null) {
+      parseTypedArray(context, attrs)
```
pengdev, line 86: Does this mean that the options defined in xml attributes will overwrite the ones defined by user programmatically?

Chaoba: Good catch, need to rethink about it.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
pengdev: @kiryldz 
> @pengdev why would we actually need to override pixel ratio from context as part of default options? I know we have issues if accessing pixel ratio from static instances in case of multi-display but what's the use-case when pixle ratio taken from concrete context is not correct?

The pixel ratio is configurable, it doesn't have to be the context's default density. But on the other hand, I think it can be configured through xml and when users build the ResourceOptions themselves. I think we can leave the current API as it is for now.

> Another potential problem I see is if we override it here and pass to core but will rely from pixel ratio from context in some other methods.

The pixel ratio set in the MapInitOptions#MapOptions is saved in the MapboxMap object and Style object, and we make sure it is used across our APIs instead of using the pixel ratio from the context.

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.content.Context
+
+/**
+ * Convenience class that holds [MapboxMap] related token.
+ * `CredentialsManager` could be created per `MapView` instance if given view should use some specific token.
+ * Most common use-case when all `MapView`s should use same token could be handled by using `CredentialsManager.shared` static object.
+ *
+ * @property accessToken the token will be applied.
+ */
+data class CredentialsManager(val accessToken: String? = null) {
```
pengdev, line 12: can be private val

pengdev: let's follow up with iOS @Chaoba

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+   * Static variables and methods.
+   */
+  companion object {
+    /**
+     * The default shared instance with empty assess token, will search for an access token in the application resources.
+     */
+    var shared: CredentialsManager = CredentialsManager()
```
pengdev, line 41: make it val

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
pengdev: change to `var`

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -37,14 +37,14 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   private var mapController: MapController
 
   /**
-   * Build a [MapView] with [Context] and [MapboxMapOptions] objects.
+   * Build a [MapView] with [Context] and [MapInitOptions] objects.
    */
-  constructor(context: Context, mapboxMapOptions: MapboxMapOptions) : this(
+  constructor(context: Context, mapInitOptions: MapInitOptions) : this(
```
pengdev, line 42:
```suggestion
  constructor(context: Context, mapInitOptions: MapInitOptions = MapInitOptions(context)) : this(
```

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
pengdev: rename to `default`

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
Chaoba: Yeah, that makes sense.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -279,20 +277,20 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
-      val plugin = if (attrs != null) {
+      val plugin = if (mapInitOptions.attrs != null) {
         createPlugin(
           mapView,
           gesturePluginClass,
-          Pair(Context::class.java, context),
-          Pair(AttributeSet::class.java, attrs),
-          Pair(Float::class.java, pixelRatio)
+          Pair(Context::class.java, mapInitOptions.context),
+          Pair(AttributeSet::class.java, mapInitOptions.attrs!!),
```
pengdev, line 285: shall we try to avoid using !! and use smart cast etc?

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -0,0 +1,54 @@
...
+ * @property textureView Flag indicating to use a TextureView as render surface for the MapView. Default is false.
+ * @property attrs The [AttributeSet] object that init the MapView.
+ */
+data class MapInitOptions constructor(
+  val context: Context,
+  var resourceOptions: ResourceOptions = getDefaultResourceOptions(context),
+  var mapOptions: MapOptions = getDefaultMapOptions(context),
+  var initialCameraOptions: CameraOptions? = null,
+  var textureView: Boolean = false,
+  var attrs: AttributeSet? = null
```
pengdev, line 27: Trying to think this again, the attrs doesn't feel a part of MapInitOptions, it can be used for creating the MapInitOptions, but it's not the actual configuration. wdyt? @Chaoba

pengdev (approved):
LGTM % nits

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
Chaoba: Would keep it because it makes our codes clean

PR #248, opened 2021-04-12
Load streets by default if no style is provided

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Load streets by default if no style is provided</changelog>`.

### Summary of changes
This pr add check fin `onResume`, if there is no style provided, we will load MAPBOX_STREET by default.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -109,6 +109,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     renderer.onStart()
     pluginRegistry.onStart()
+    if (!mapboxMap.isStyleLoadInited) {
+      // Load the default style if no style has loaded yet.
+      mapboxMap.loadStyleUri(Style.MAPBOX_STREETS)
+    }
   }
 
   fun reduceMemoryUse() {
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -46,7 +46,7 @@ class MapboxMap internal constructor(
 
   private val nativeMapWeakRef = WeakReference(nativeMap)
   internal lateinit var style: Style
-
+  internal var isStyleLoadInited = false
   private val handlerMain = Handler(Looper.getMainLooper())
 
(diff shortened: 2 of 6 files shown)


sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -109,6 +109,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     renderer.onStart()
     pluginRegistry.onStart()
+    if (!mapboxMap.isStyleLoadInited) {
```
kiryldz, line 112: What will happen if user will load his style in `Activity.onStart()`? This will actually mean that we will already start loading `Style.MAPBOX_STREETS` from `mapView.onStart()`. Will everything work OK then? I guess desired behaviour is to result in user style loaded and not default one.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -176,6 +176,7 @@ class MapboxMap internal constructor(
     onStyleLoaded: Style.OnStyleLoaded? = null,
     onMapLoadErrorListener: OnMapLoadErrorListener? = null
   ) {
+    isStyleLoadInited = true
```
kiryldz, line 179: Do we need to set it to false somewhere? in some destroy method.

sdk/src/main/java/com/mapbox/maps/MapController.kt
Chaoba: User's style will be loaded instead of the default one.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
Chaoba: No, we will create new MapboxMap instance for new MapView

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: Without any glitches, correct?

kiryldz (approved):

sdk/src/main/java/com/mapbox/maps/MapController.kt
Chaoba: Yes, no glitches

Issue #114, opened 2021-02-18
java.lang.IndexOutOfBoundsException in test app.


pengdev:
## Environment

- Android OS version: Android 7.1 (API level 25)
- Devices affected: PH-1
- Maps SDK Version: v10.0.0-beta.13

## Observed behavior and steps to reproduce

Our CI caught the following crash:
```
java.lang.IndexOutOfBoundsException: Index: 6, Size: 2
     FATAL EXCEPTION: Thread-2
Process: com.mapbox.maps.testapp, PID: 2540
java.lang.IndexOutOfBoundsException: Index: 6, Size: 2
	at java.util.ArrayList.get(ArrayList.java:411)
	at com.mapbox.maps.testapp.ExampleOverviewActivity$onCreate$$inlined$apply$lambda$1.onItemClicked()
	at com.mapbox.maps.testapp.utils.ItemClickSupport$onClickListener$1.onClick()
	at android.view.View.performClick(View.java:5637)
	at android.view.View$PerformClick.run(View.java:22434)
	at android.os.Handler.handleCallback(Handler.java:751)
	at android.os.Handler.dispatchMessage(Handler.java:95)
	at androidx.test.espresso.base.Interrogator.a(Interrogator.java:11)
	at androidx.test.espresso.base.UiControllerImpl.n(UiControllerImpl.java:6)
	at androidx.test.espresso.base.UiControllerImpl.m(UiControllerImpl.java:1)
	at androidx.test.espresso.base.UiControllerImpl.a(UiControllerImpl.java:6)
	at androidx.test.espresso.action.MotionEvents.a(MotionEvents.java:15)
	at androidx.test.espresso.action.Tap.b(Tap.java)
	at androidx.test.espresso.action.Tap$1.sendTap(Tap.java:1)
	at androidx.test.espresso.action.GeneralClickAction.perform(GeneralClickAction.java:4)
	at androidx.test.espresso.ViewInteraction$SingleExecutionViewAction.perform(ViewInteraction.java:4)
	at androidx.test.espresso.ViewInteraction.a(ViewInteraction.java)
	at androidx.test.espresso.ViewInteraction$1.call(ViewInteraction.java)
	at java.util.concurrent.FutureTask.run(FutureTask.java:237)
	at android.os.Handler.handleCallback(Handler.java:751)
	at android.os.Handler.dispatchMessage(Handler.java:95)
	at android.os.Looper.loop(Looper.java:154)
[...]

PR #255, opened 2021-04-14
[example] Fix IndexOutOfBoundsException in test app

Chaoba:
Fixes: #114 

## Pull request checklist:

### Summary of changes
Root cause for the crash is that if we searched some text, the example list will show the searched result. Then if the search text is deleted, although it will show all examples, the `currentlyDisplayedExampleList` isn't updated and keeps the previous searched result. At this time if click an item with a bigger index the crash will occur.

This pr also fix the scenario that if we click the back arrow on the left top of an example back to the example list activity, the example list activity will be recreated and not be able to keep the previous state.


kiryldz (approved):
Great job, will make our CI more stable for robo tests!

PR #264, opened 2021-04-16
[docs] Fix directory name for 10.0.0-beta.17 docs

danswick:
This PR changes the 10.0.0-beta.17 docs directory name from `android-v10.0.0-beta.17` to `10.0.0-beta.17` to match the directory names for other versions. 

Since these Dokka PRs are so enormous, we will need a second PR to kick-start publisher, unfortunately.

cc @Chaoba


colleenmcginnis (approved):
🍩

Issue #277, opened 2021-04-21
Anchor point is not restored after gesture interaction


LukasPaczos:
## Environment
- Maps SDK Version: 10.0.0-beta.17

## Observed behavior and steps to reproduce

When the camera animations plugin is initialized and the anchor is `null`, playing around with gestures (zooming in/out, rotating) often leaves the anchor point set to a certain value, `null` is not restored correctly in all cases.

## Expected behavior

After gesture interaction is finished, the anchor point is restored back to `null`.

## Notes / preliminary analysis

This is very easily reproducible by logging out the `CameraAnimationsPlugin#anchor` value. You would quickly notice that when you stop executing a gesture, the anchor point would not be reset back to `null`.

This impacts all the other camera transitions that the app is trying to run afterward, which can result in very unexpected camera position changes.

For example:

https://user-images.githubusercontent.com/16925074/115553998-50e31280-a2ae-11eb-93e5-a3db66197c65.mp4

In the example above, if the anchor point was correctly reset to `null` by the gesture recognizer, the transition would correctly zoom in on top of the puck.

The same unexpected effect will happen for all user-driven transitions that modify the zoom level or bearing of the camera.

The workaround is to explicitly set the anchor to `null` before starting any transition.

cc @mapbox/maps-android

PR #279, opened 2021-04-22
[gestures] Properly reset anchor after some gestures

kiryldz:
Fixes: #277 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[gestures] Properly reset anchor after some gestures</changelog>`.

### DIFF
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
@@ -70,6 +70,7 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
   private var screenHeight: Double = 0.0
   private var startZoom: Double = 0.0
   private var flingInProcess: Boolean = false
+  private var scaleCachedAnchor: ScreenCoordinate? = null
 
   // Rotate
   private var minimumScaleSpanWhenRotating: Float = 0f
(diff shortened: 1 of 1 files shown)


tobrun (approved):
nit would have expected a test to be impacted by this change

pengdev (commented):
Agree with @tobrun that we need a test to cover it.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1051,10 +1062,13 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
 
       override fun onAnimationStart(animation: Animator) {}
 
-      override fun onAnimationCancel(animation: Animator) {}
+      override fun onAnimationCancel(animation: Animator) {
+        cameraAnimationsPlugin.anchor = scaleCachedAnchor
+        dispatchCameraIdle()
```
LukasPaczos, line 1067: Isn't `onAnimationEnd` always following `onAnimationCancel`? Do we need to duplicate the logic here?

kiryldz:
> Agree with @tobrun that we need a test to cover it.

Would also bring a valuable test here but sadly we have `// todo animator testing` in those tests, will create a follow-up task and do that later.

Chaoba (approved):

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: https://github.com/mapbox/mapbox-maps-android/pull/280

PR #290, opened 2021-04-26
[offline] Instrument tests for offline

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>[offline] Instrument tests for offline</changelog>`.

### Summary of changes

Add instrument tests for new offline. Those tests cover:

- Test that validates progress and completion blocks are called and result is as expected.
- Test that validates that in progress downloads can be canceled.
- Test that validates that resource loading comes from the tile store.
- Test that validates that an offline region and style pack could be correctly deleted.
- Test that validates that map can be rendered with no internet (waiting for IDLE event).


sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -0,0 +1,627 @@
...
+  fun loadStylePackNoCancel() {
+    val latch = CountDownLatch(1)
+    var downloadProgress: StylePackLoadProgress? = null
+    var downloadedStyle: StylePack? = null
+    var downloadedStyleError: StylePackError? = null
+    handler.post {
+      offlineManager.loadStylePack(
+        STYLE,
+        StylePackLoadOptions.Builder()
+          .glyphsRasterizationMode(GlyphsRasterizationMode.IDEOGRAPHS_RASTERIZED_LOCALLY)
```
pengdev, line 79: you will need to specify a fontFamily here.

kiryldz: can you provide more information what does it mean in this case? where should I specify `fontFamily` here?

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -0,0 +1,627 @@
...
+
+      Assert.assertNotEquals(
+        tileRegionProgress?.requiredResourceCount,
+        tileRegionProgress?.completedResourceCount
+      )
+      Assert.assertEquals(TileRegionErrorType.CANCELED, tileRegionError?.type)
+    }
+  }
+
+  // TODO revisit after https://github.com/mapbox/mapbox-sdk-common/issues/1311
```
pengdev, line 296: nit, remove links to internal issue.

kiryldz: @tobrun aren't we allowed to add such links in code?

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -0,0 +1,627 @@
...
+    val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+    device.waitForIdle()
+    // adopted for current CI device
+    device.findObject(UiSelector().textContains("Airplane")).click()
+    device.waitForIdle()
+    val latch = CountDownLatch(1)
+    // need to wait some time for pretty slow CI device
+    latch.await(5, TimeUnit.SECONDS)
+    device.pressBack()
+  }
```
pengdev, line 615: can we move this utility function to some base class so it can be reused? We could also use the function provided by common.

kiryldz: What is the function provided by common?

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
pengdev: oh, I am referring to something like https://github.com/mapbox/mapbox-maps-android/blob/5ad0667774d62e72b43cc469ca86b4f30521c7b8/sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt#L51-L54 , but it seems we don't have an option to specify it here. @pozdnyakov do you think we need an option here to specify the text fonts?

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -0,0 +1,627 @@
...
+              .networkRestriction(NetworkRestriction.NONE)
+              .criticalPriority(false)
+              .acceptExpired(true)
+              .build()
+          )
+          .build(),
+        {}
+      ) {
+        if (it.isValue) {
+          it.value?.let { latch.countDown() }
```
pengdev, line 568: This callback will only report the loading finished status for tile region, it will not reflect the status of the style pack, so  potentially the style pack can be still downloading at this point of time. And that's also the reason we were downloading the style pack and tile region separately in the OfflineActivity.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -0,0 +1,627 @@
...
+        MapEvents.MAP_LOADED,
+        MapEvents.STYLE_LOADED,
+        MapEvents.MAP_LOADING_ERROR
+      )
+    )
+    try {
+      // we do not throw exception here but verify request count
+      latch.await(10, TimeUnit.SECONDS)
+      // verify resource requests came from tile store
+      MatcherAssert.assertThat(resourceRequests, Matchers.greaterThan(0))
```
pengdev, line 335: shall we also assert that the MAP_LOADED and STYLE_LOADED event are actually emitted?

kiryldz: Full functionality of this test is broken as stated in a bunch of TODOs - so let's leave as is now to verify that tiles are loaded.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
kiryldz: @pozdnyakov can you please confirm that behaviour: `TileRegionCallback` could get invoked before style pack as part of descriptor gets downloaded. For me it seems more like a bug - I would expect `TileRegionCallback` be triggered after style pack being loaded as part of tile options descriptor.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
pengdev: Sorry I just noticed you have a separate test for it, `TileStoreMapboxSwitch`.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
pengdev: let's ticket out the tailwork in this repo to improve these test when up steam is fixed.

pengdev (approved):
LGTM with nits

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
pozdnyakov: Nope, font family is not an option for offline download but only for the map (for the local glyph rasterizer)

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
pozdnyakov: if the tileset descriptor is configured so that it creates a style pack, the style pack is created by the time the tileset descriptor gets resolved i.e. before the tile region loading starts

kiryldz: @pozdnyakov thanks for confirming!
@pengdev - it seems that we're OK here morevover it looks like a good practice to use such configuration to load tile + style pack as it requires less code and should work fine.
If loading style pack + tile region separately we need to preserve correct order - first load style pack and only then load tile region otherwise style will be taken from ambient cache and not style pack.

pengdev: Thanks for confirming, but I think we observed some potential different behaviour when building the offline example(cc @julianrex), probably we should verify the it again.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
tobrun: idk anymore :) we started off with never using any links to internal and now we are so :shrug:

kiryldz: looked through code, indeed we do not have such internal links anymore, will remove it and create a new task.

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
kiryldz: https://github.com/mapbox/mapbox-maps-android/issues/297

PR #292, opened 2021-04-27
Request layout when updating ornaments margins

tobrun:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Request layout when updating ornaments margins, making updates immidiate</changelog>`.

### Summary of changes
 - adds an example to reproduce the user reported bug
 - makes sure we call requestLayout after we have set margins of an ornament
 - correct calling set margins correctly with the LTRB order

### User impact (optional)

Developers now can see margins of the ornaments updating automatically when they set the values (workaround could be to call request layout yourself on the MapView). 

image

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -967,5 +967,17 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".ExampleOverviewActivity" />
         </activity>
+        <activity
+            android:name=".examples.OrnamentMarginActivity"
+            android:description="@string/description_ornaments_margin"
+            android:exported="true"
+            android:label="@string/activity_ornaments">
+            <meta-data
+                android:name="@string/category"
+                android:value="@string/category_config" />
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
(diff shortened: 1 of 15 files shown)


plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
```diff
@@ -72,6 +72,7 @@ class AttributionViewImpl @JvmOverloads constructor(
    */
   override fun setAttributionMargins(left: Int, top: Int, right: Int, bottom: Int) {
     (layoutParams as FrameLayout.LayoutParams).setMargins(left, top, right, bottom)
+    requestLayout()
```
kiryldz, line 75: Let's add test here and below

kiryldz (approved):

Chaoba:
Why there is no `requestLayout()` for scalebar plugin?

tobrun:
> Why there is no requestLayout() for scalebar plugin?

Because scalebar is drawing content on its own and isn't a standard view.

Chaoba (approved):

PR #300, opened 2021-04-29
[sdk] rework style loading to avoid invoking previously set listeners

tobrun:
Replaces https://github.com/mapbox/mapbox-maps-android/pull/178

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Style and map error loading listeners are only called for the style that was associated to the listeners when style loading started. This avoid calling the wrong listeners with multiple style loads. Renamed Style#isStyleLoadInited to Style#isStyleLoadInitiated.</changelog>`.

### Summary of changes

This PR reworks how we do style loading internally to expose these to developers. We make sure to clear any set listeners when a new style will be loaded. This avoids calling OnStyleLoaded on an old obsolete instance of a previous set style. 

### User impact (optional)

When a user load a style and it errors out and load a new style. Only the secondary interface provided to the second style loading will be invoked. The first one will be ignored.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -102,7 +102,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     renderer.onStart()
     pluginRegistry.onStart()
-    if (!mapboxMap.isStyleLoadInited) {
+    if (!mapboxMap.isStyleLoadInitiated) {
       // Load the style in mapInitOptions if no style has loaded yet.
       mapInitOptions.styleUri?.let {
         mapboxMap.loadStyleUri(it)
@@ -125,6 +125,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun onDestroy() {
+    mapboxMap.onDestroy()
     nativeObserver.clearListeners()
     renderer.onDestroy()
     pluginRegistry.cleanup()
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -15,7 +15,6 @@ import com.mapbox.maps.plugin.PLUGIN_GESTURE_CLASS_NAME
 import com.mapbox.maps.plugin.animation.CameraAnimationsPlugin
 import com.mapbox.maps.plugin.delegates.*
 import com.mapbox.maps.plugin.delegates.listeners.*
-import com.mapbox.maps.plugin.delegates.listeners.eventdata.StyleDataType
 import com.mapbox.maps.plugin.gestures.GesturesPlugin
 import java.lang.ref.WeakReference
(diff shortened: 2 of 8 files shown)


kiryldz (commented):
@tobrun I see several merge conflicts, can we resolve those prior to review?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1244,4 +1210,12 @@ class MapboxMap internal constructor(
     }
     return null
   }
+
+  internal fun onDestroy() {
+    styleObserver.onDestroy()
+  }
+
+  companion object {
+    private const val TAG = "MBX"
```
kiryldz, line 1219: tag not used as far as I see

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -0,0 +1,76 @@
...
+    loadedListener?.let {
+      awaitingStyleLoadListeners.add(loadedListener)
+    }
+    awaitingStyleErrorListener = onMapLoadErrorListener
+  }
+
+  /**
+   * Add a style listener invoked when a new style loaded events occurs.
+   */
+  fun addOnStyleLoadListener(loadedListener: Style.OnStyleLoaded) {
```
kiryldz, line 42: we have add, perhaps we also need a remove?

tobrun: what is the use-case? This is an internal API

kiryldz: I see `add` and get triggered to add `remove`. But agree, if internal API - then guess it's good as is.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -102,7 +102,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     renderer.onStart()
     pluginRegistry.onStart()
-    if (!mapboxMap.isStyleLoadInited) {
+    if (!mapboxMap.isStyleLoadInitiated) {
```
kiryldz, line 105: It's a bit strange we see it here, is this PR correctly rebased?

kiryldz (approved):
I like bunch of tests, I feel very safe :shipit:

sdk/src/main/java/com/mapbox/maps/MapController.kt
tobrun: We have an internal ticket tracking this rename #352, JFDI 
Updated changelog entry in description + added breaking change label

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
tobrun: :) 

fwiw, this API cleans up itself either as part of the callback invocation or as part of onDestroy.

PR #306, opened 2021-05-04
[geojson] Use worker thread for heavy operations

kiryldz:

## Pull request checklist:


extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
```diff
@@ -129,6 +133,17 @@ abstract class Source(
    */
   companion object {
     private const val TAG = "Mbgl-Source"
+
+    private val workerThread by lazy {
+      HandlerThread("STYLE_WORKER").apply {
```
LukasPaczos, line 138: Once this thread is started, it will remain running until the whole app process is killed. Just something to consider.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -471,7 +471,12 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    * @param value the feature
    */
   fun feature(value: Feature) = apply {
-    setProperty(PropertyValue("data", value.toValue()))
+    workerHandler.post {
```
LukasPaczos, line 474: By adding this async behavior, will this assertion pass?
```
val expected = ...
source.data(expected)
val resulting = source.data
assertEquals(expected, resulting)
```

I imagine it won't. Would it pass in the current main branch, is the round-tripping currently synchronous in this way? 

Should the getter have a callback instead of being a property, if the setter is asynchronous?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
tobrun: +1 on @LukasPaczos remark and additionally want to note that for each source that has lazy initialized it we will have a dedicated workerthread for it. I would prefer using a single - reusable that is managed as part of the MapView lifecycle.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
tobrun: To me above makes sense but in this case it's maybe not needed. gl-native doesn't allow to get back inline geojson, they don't keep a reference to it (see mapbox-gl-native-internal#522).

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,139 @@
...
+      },
+      result = {
+        style.addSource(it)
+        style.addLayer(lineLayer(LAYER, SOURCE) {
+          lineColor("blue")
+        })
+      }
+    )
+
+    // add some useless extra work to parse same geojson
```
tobrun, line 96: Test app acts as reference documentation for end-users. I would prefer making the geojson file itself bigger vs adding constructs like this. We can't use this code as documentation on mapbox.com. We could however move this to a test? or even the app-extension-style.

tobrun (commented):
Would love to see the UX for animating a GeoJsonSource geometry. We used to have an example with cars driving around a map https://github.com/mapbox/mapbox-gl-native-android/blob/main/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/annotation/AnimatedSymbolLayerActivity.java but feel free to create something more useful from scratch.

Are there any other metric data points that we capture to quantify the performance improvements? It's clear that unblocking the main thread, UX wise, is a lot better but quantifying it with numbers would be great for long term metric tracking.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: I like an idea to introduce a worker thread abstraction backing up given `MapView`. It may be started automatically in `onStart` or lazily and get killed in `onDestroy`. In future we could introduce thread pool implementing given abstraction.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: Of course this will be revisited, I was just adding it as a quick workaround to add much work on geojson parsing.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: > Would love to see the UX for animating a GeoJsonSource geometry. We used to have an example with cars driving around a map https://github.com/mapbox/mapbox-gl-native-android/blob/main/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/annotation/AnimatedSymbolLayerActivity.java but feel free to create something more useful from scratch.
> 
> Are there any other metric data points that we capture to quantify the performance improvements? It's clear that unblocking the main thread, UX wise, is a lot better but quantifying it with numbers would be great for long term metric tracking.

I think that basically we could simply add benchmark with sync / async loading geojson in our metrics repo and check out results. Or are you talking about something else?

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    mapView.getMapboxMap()
+      .apply {
+        setCamera(
+          CameraOptions.Builder()
+            .center(Point.fromLngLat(LONGITUDE, LATITUDE))
+            .zoom(START_ZOOM)
+            .build()
+        )
+        loadStyleUri(Style.MAPBOX_STREETS) { style ->
```
Chaoba, line 55: We have the default style now, can we change to `getStyle` here?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -480,7 +485,12 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    * @param value the feature collection
    */
   fun featureCollection(value: FeatureCollection) = apply {
-    setProperty(PropertyValue("data", value.toValue()))
+    workerHandler.post {
```
Chaoba, line 488: There are many similar code blocks, can we extract them as a method?

Chaoba (commented):
Instead of adding the large file `long_route.json` to the repo directly, can we load it from a http uri like we did in PointAnnotationClusterActivity?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
LukasPaczos: I was thinking about that as well, but tieing the lifecycle of that thread to a `MapView` might be harsh.

The `Source` object itself is in no way tied to the `MapView` lifecycle. It could be created before `MapView` is inflated and then added to it. If the thread is tied to the `MapView` for practical reasons, then the `Source` impl has to have some kind of a queue (coalesced) waiting for the thread to get started.

Also a note on the thread pool - it would have to be well synchronized within a source instance, otherwise, the updates might be competing and overwriting each other.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
pengdev: This code block is inside the onCreate, as far as I remember, the default style is loaded in the onStart? will the added RuntimeStyle be overwritten by the style?

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
pengdev: > We could however move this to a test? or even the app-extension-style.

This sounds like a good instrumentation test in `the app-extension-style`.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
Chaoba: No, we have check there, if one style is already set we will not set the default style.

Issue #308, opened 2021-05-05
Crash when setting camera in onStart


kiryldz:
## Environment

- Android OS version: any
- Devices affected: any
- Maps SDK Version: 10.0.0-beta.18

## Observed behavior and steps to reproduce

```
override fun onStart() {
    super.onStart()
    mapView.onStart()
    mapView.getMapboxMap().setCamera(
      CameraOptions.Builder()
        .center(Point.fromLngLat(17.045988781311138, 51.12341909978734))
        .bearing(27.254667247679752)
        .pitch(0.0)
        .padding(EdgeInsets(140.0, 140.0, 140.0, 140.0))
        .zoom(13.282170222962456)
        .build()
    )
  }
```

leads to a crash in compass plugin because of using wrong thread to update widget.

```
2021-05-05 15:31:11.744 19796-21231/com.mapbox.maps.testapp E/AndroidRuntime: FATAL EXCEPTION: MapboxRenderThread
    Process: com.mapbox.maps.testapp, PID: 19796
    android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
        at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:10689)
        at android.view.ViewRootImpl.focusableViewAvailable(ViewRootImpl.java:5556)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.ViewGroup.focusableViewAvailable(ViewGroup.java:980)
        at android.view.View.setFlags(View.java:17519)
        at android.view.View.setVisibility(View.java:12466)
        at android.widget.ImageView.setVisibility(ImageView.java:1702)
        at com.mapbox.maps.plugin.compass.CompassViewImpl.setCompassVisible(CompassViewImpl.kt:66)
        at com.mapbox.maps.plugin.compass.CompassViewPlugin.updateVisibility(CompassViewPlugin.kt:244)
        at com.mapbox.maps.plugin.compass.CompassViewPlugin.updateVisibility$default(CompassViewPlugin.kt:225)
[...]

PR #310, opened 2021-05-05
Fix crash if doing setCamera during map loading

kiryldz:
Fixes: #308 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix crash if doing setCamera during map loading</changelog>`.

### Summary of changes

Every consumer (in our case it's compass plugin) should make sure that `MapCameraPlugin.onCameraMove` is called from correct thread.

### DIFF
--- a/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
+++ b/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
@@ -183,7 +183,9 @@ open class CompassViewPlugin(
     bearing: Double,
     padding: Array<Double>
   ) {
+//    Handler(Looper.getMainLooper()).post {
     update(bearing)
+//    }
   }
 
   /**
--- a/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
+++ b/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
@@ -2,6 +2,8 @@ package com.mapbox.maps.plugin.compass
 
 import android.animation.ValueAnimator
 import android.content.Context
+import android.os.Handler
+import android.os.Looper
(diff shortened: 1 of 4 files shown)


plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
```diff
@@ -183,7 +185,9 @@ open class CompassViewPlugin(
     bearing: Double,
     padding: Array<Double>
   ) {
-    update(bearing)
+    Handler(Looper.getMainLooper()).post {
```
tobrun, line 188: it is expected that camera callbacks are happening on a non ui thread?

tobrun (approved):

plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
tobrun: maybe reuse Handler instance?

kiryldz: yes, main idea here is that they must return immediately from core and not get put to some scheduler. This call from render thread happens because `setSize` from renderer also calls `setCamera` under the hood. In all other cases I guess it will triggered from main thread.

tobrun: maybe setSize shouldn't happen on render thread?

kiryldz: I think it should as it is connected with textures and surfaces being ready. Even if we move it some other thread as said above camera changes are not scheduled on main thread under the hood meaning that if user will call `setCamera` from non-main thread - we will crash in compass.

PR #314, opened 2021-05-06
[sdk] rename ambient cache database to ambient_cache.db 

tobrun:
This is to align with iOS in https://github.com/mapbox/mapbox-maps-ios/pull/324.
No public API is impacted but I'm dubbing this as a breaking change as users will lose previous stored data from the ambient cache. 

## Pull request checklist:
 - [X] Briefly describe the changes in this PR.
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Rename ambient cache database to ambient_cache.db </changelog>`.

### Summary of changes

This PR changes the the ambient cache name of the database to ambient_cache.db to align with iOS.

### User impact (optional)

Previous stored ambient cache data will be lost when migrating to a newer version of the SDK.

cc @LukasPaczos

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapboxConstants.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxConstants.kt
@@ -7,7 +7,12 @@ import java.util.*
 /**
  * Name of the database file.
  */
-const val DATABASE_NAME = "mbx.db"
+const val DATABASE_NAME = "ambient_cache.db"
+
+/**
+ * Path of the database file.
+ */
+const val DATABASE_PATH = "mapbox/maps"
 
 /**
  * The default cache size, which is 50MB
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,6 +297,11 @@ jobs:
   run-unit-test:
     docker:
       - image: mbgl/android-ndk-r21e:latest
+    resource_class: xlarge
+    environment:
+      JVM_OPTS: -Xmx3200m
+      BUILDTYPE: Debug
+      GRADLE_OPTS: -Xmx4096m -Dorg.gradle.daemon=false -Dkotlin.compiler.execution.strategy=in-process
 
     steps:
       - checkout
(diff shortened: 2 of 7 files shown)


sdk/src/main/java/com/mapbox/maps/MapboxConstants.kt
```diff
@@ -7,7 +7,7 @@ import java.util.*
 /**
  * Name of the database file.
  */
-const val DATABASE_NAME = "mbx.db"
+const val DATABASE_NAME = "ambient_cache.db"
```
pengdev, line 10: I think ios is also using a mapbox directory to contain this db file.

julianrex: The folder originally was `<app id>/.mapbox/cache.db` - PR changes it to `com.mapbox.maps/ambient_cache.db` (since we're in the apps support directory)

julianrex (approved):
Not sure if you need a `com.mapbox.maps` directory with Android, but otherwise 👍🏻

kiryldz (approved):

sdk/src/main/java/com/mapbox/maps/MapboxConstants.kt
pengdev: Agree with @tobrun here, I'd suggest to use /mapbox/ambient_cache.db

kiryldz (approved):

tobrun:
Capturing some internal discussion:
 - moving forward now with: `mapbox/maps/ambient_cache.db`
 - long term, potentially migrating to `com/mapbox/maps/ambient_cache.db` pending on upstream tilestore changes

sdk/src/main/java/com/mapbox/maps/MapboxConstants.kt
```diff
@@ -7,7 +7,7 @@ import java.util.*
 /**
  * Name of the database file.
  */
-const val DATABASE_NAME = "mbx.db"
+const val DATABASE_NAME = "mapbox/maps/ambient_cache.db"
```
pengdev, line 10: Please verify the mapbox/maps folder is created, we need to create the path before passing to gl-native. The limitation is ticketed upstream.

pengdev:
Please rebase to latest main since the ResourceOptionsManager PR is merged.

PR #327, opened 2021-05-11
Parsing geojson on a worker thread

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Parsing geojson on a worker thread.</changelog>`.

### Summary of changes

All geojson properties to set data **feature()**, **featureCollection()**, **geometry()** will now use __worker thread__ to perform parsing given `GeoJson` before passing it as `String` via JNI to core on __main thread__.

Using large geojson parsing __without worker thread__ (you could see significant time to open an activity after click because main UI thread is busy with parsing):

https://user-images.githubusercontent.com/15800566/118973674-990f5680-b97a-11eb-8892-d9e6c2fc44ff.mp4

Using same geojson sources __with worker thread__ (there are still some minor lags happening when passing parsed large geojson as `String` via JNI but UI thread is much more free):

https://user-images.githubusercontent.com/15800566/118973787-ba704280-b97a-11eb-91e8-a743d9d69c33.mp4

### User impact 

If preparing geojson with DSL functions new logic is introduced.

```
/**
 * DSL function for [GeoJsonSource] performing parsing using background thread.
 * Immediately returns [GeoJsonSource] with no data set and starts preparing actual data
 * using a worker thread.
 *
 * If using runtime styling:
 *
 * loadStyle(style(Style.DARK) {
 *   +geoJsonSource(id) {
 *    featureCollection(collection)
 *   }
 *   ...
 * }
 *
 * compositing style will be performed correctly under the hood and
 * [Style.OnStyleLoaded] will be emitted in correct moment of time when all sources are parsed.
 *
 * If creating geojson sources for already loaded Style please consider using overloaded
 * geoJsonSource(String, GeoJsonSource.Builder.() -> Unit, onGeoJsonParsed: (GeoJsonSource) -> Unit) function
 * and use fully prepared [GeoJsonSource] in onGeoJsonParsed callback.
 */
fun geoJsonSource(
  id: String,
  block: GeoJsonSource.Builder.() -> Unit
): GeoJsonSource

/**
 * DSL function for [GeoJsonSource] performing parsing using a worker thread.
 * Immediately returns [GeoJsonSource] with no data set,
 * fully parsed [GeoJsonSource] is returned in [onGeoJsonParsed] callback.
 *
 * Using this method means that it is user's responsibility to proceed with adding this source,
 * layers or other style objects in [onGeoJsonParsed] callback.
 */
fun geoJsonSource(
  id: String,
  block: GeoJsonSource.Builder.() -> Unit,
  onGeoJsonParsed: (GeoJsonSource) -> Unit
): GeoJsonSource
```

If updating existing `GeoJsonSource` from some `Style` overloaded functions are introduced:

```
/**
   * Add a FeatureCollection to the GeojsonSource.
   * If [onDataParsed] is provided and not null - data will be loaded in async mode.
   * Otherwise method will be synchronous.
   *
   * @param onDataParsed optional callback notifying when data is parsed on a worker thread
   */
  fun featureCollection(
    value: FeatureCollection,
    onDataParsed: ((GeoJsonSource) -> Unit)? = null
  )
```

Similar functions are introduced for `feature` and `geometry`.


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -167,10 +168,34 @@ class MapboxMap internal constructor(
...
+      } else {
+        it.bindTo(style)
+      }
+    }
+    if (!hasGeojsonSources) {
+      loadAllExceptGeojsonSources(style, styleExtension, onStyleLoaded)
+    }
+  }
+
+  private fun loadAllExceptGeojsonSources(
```
pengdev, line 193: why don't we set the empty geojsonsource first and then update the geojsonSource data property? we don't need to complicate the process here..

kiryldz: We don't want complexity but we will have it in any case 😄 
It's all because that we will still need exact correct moment of time to emit `onStyleLoaded?.onStyleLoaded(style)` when **everything** is loaded. That brings a lot of complexity as we must emit it when all geojson sources are parsed, after that we add everything else in sync manner and could emit `onStyleLoaded?.onStyleLoaded(style)`.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -499,6 +539,8 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    */
   @SourceDsl
   class Builder(val sourceId: String) {
+
+    internal var rawData: Any? = null
```
pengdev: Can we save a `CompletableFuture<String>` instead of the rawData here, we can start processing the data already when user set it, and we can check if the future is done and set the geojson immediately or if it's not done, we could append async setting data property in the `completableFuture#thenAccept` callback.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,141 @@
...
+    mapView.getMapboxMap()
+      .apply {
+        setCamera(
+          CameraOptions.Builder()
+            .center(Point.fromLngLat(LONGITUDE, LATITUDE))
+            .zoom(START_ZOOM)
+            .build()
+        )
+        loadStyle(style(Style.DARK) {
+          +geoJsonSource("source") {
```
Chaoba, line 56: I would prefer using a for loop here to add sources and layers

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
```diff
@@ -130,6 +134,17 @@ abstract class Source(
    */
   companion object {
     private const val TAG = "Mbgl-Source"
+
+    private val workerThread by lazy {
```
Chaoba, line 138: Why create the workerThread here instead of GeojsonSouce?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -21,7 +21,32 @@ import com.mapbox.maps.extension.style.utils.toValue
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+class GeoJsonSource(
+  builder: Builder,
+  // must be instance of Geometry, Feature, FeatureCollection
+  private val rawData: Any? = null
```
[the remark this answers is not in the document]
kiryldz: oh, indeed, for some reason I though they all extend just `Object`.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -167,10 +169,43 @@ class MapboxMap internal constructor(
...
+            if (geojsonNotParsedCount == 0) {
+              loadAllExceptGeojsonSources(style, styleExtension, onStyleLoaded)
+            }
+          }
+        }
+      } else {
+        it.bindTo(style)
+      }
+    }
+    if (hasGeojsonSources) {
```
Chaoba, line 193: We can merge the else block here with `if(!hasGeojsonSources || geojsonNotParsedCount == 0)`

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -167,10 +169,43 @@ class MapboxMap internal constructor(
...
+        hasGeojsonSources = true
+        if (it.geoJsonParsed) {
+          it.bindTo(style)
+        } else {
+          geojsonNotParsedCount++
+          it.geoJsonParsedListener = {
+            geojsonNotParsedCount--
+            it.bindTo(style)
+            if (geojsonNotParsedCount == 0) {
+              loadAllExceptGeojsonSources(style, styleExtension, onStyleLoaded)
```
pengdev, line 185: if we set the empty geojsonSource first, we don't need to wait for all geojson sources to be loaded before adding layers etc.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: That will be revisited for sure 👍

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: Yep, moved to `GeojsonSouce`

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: Revisited overall logic here.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: Applied.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: Discussed offline, approach was changed.

extension-style/src/main/java/com/mapbox/maps/extension/style/StyleExtensionImpl.kt
```diff
@@ -11,7 +11,10 @@ import com.mapbox.maps.extension.style.terrain.generated.Terrain
 /**
  * The concrete implementation of style extension.
  */
-class StyleExtensionImpl private constructor(builder: Builder) : StyleContract.StyleExtension {
+class StyleExtensionImpl private constructor(
+  builder: Builder,
+  override var resourceCount: Int
```
pengdev, line 16: nit, can be val

pengdev: it could also just be a method to just count the sources/layers/light/images etc. once constructed.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -21,7 +25,38 @@ import com.mapbox.maps.extension.style.utils.toValue
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+class GeoJsonSource(
+  builder: Builder
+) : Source(builder.sourceId) {
+
+  var geoJsonParsed = false
+  var onGeoJsonParsedListener: ((GeoJsonSource) -> Unit)? = null
```
pengdev, line 33: can we abstract these two logic to some generic class?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -21,7 +25,38 @@ import com.mapbox.maps.extension.style.utils.toValue
...
+    rawGeoJson?.let {
+      onGeoJsonParsedListener = onGeoJsonParsed
+      workerHandler.post {
+        val property = PropertyValue(
+          "data",
+          when(it) {
+            is Geometry -> it.toValue()
+            is FeatureCollection -> it.toValue()
+            is Feature -> it.toValue()
+            else -> ""
```
pengdev, line 49: should we log some error here.

extension-style/src/main/java/com/mapbox/maps/extension/style/StyleExtensionImpl.kt
kiryldz: we decrease and increase it directly from `MapboxMap`, so no 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: that situation should never happen, will add runtime exception

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -498,7 +548,12 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    * @param sourceId the ID of the source
    */
   @SourceDsl
-  class Builder(val sourceId: String) {
+  class Builder(
+    val sourceId: String,
+    private val onGeoJsonParsed: (GeoJsonSource) -> Unit = {}
```
pengdev, line 553: should this be exposed to users?

pengdev: in case they want to get a callback, but by default we set it to {}

kiryldz: Would like to hear from you about it 😄 
Don't have strong opinion yet, I guess using this builder directly in general may not be very common as we have dsl function

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleContract.kt
```diff
@@ -10,6 +10,11 @@ interface StyleContract {
    * Define the interfaces for the Style Extension.
    */
   interface StyleExtension {
+
+    /**
+     * Total resource count including all sources, layers and images.
+     */
+    var resourceCount: Int
```
pengdev, line 17:
```suggestion
    val resourceCount: Int
      get() = sources.size + images.size + layers.size + if (light != null) 1 else 0 + if (terrain != null) 1 else 0
```

kiryldz: We are decreasing it in `MapboxMap` so will not work that way I guess.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -167,18 +168,38 @@ class MapboxMap internal constructor(
...
+          styleExtension.resourceCount--
+        } else {
+          it.onGeoJsonParsedListener = { geoJson ->
+            geoJson.bindTo(style)
+            styleExtension.resourceCount--
+            if (styleExtension.resourceCount == 0) {
+              onStyleLoaded?.onStyleLoaded(style)
+            }
+          }
+        }
```
pengdev, line 184: as commented before, I think we can always use this logic branch.

extension-style/src/main/java/com/mapbox/maps/extension/style/StyleExtensionImpl.kt
pengdev: why do we need it that way? I think the mapboxmap can use a var locally.

pengdev: consider that the style extension can be applied to again when the style is changed.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
pengdev: it can be used with DSL as well, just looks a bit weird..

extension-style/src/main/java/com/mapbox/maps/extension/style/StyleExtensionImpl.kt
kiryldz: current approach will handle it ok, but I like your more, will do it your way, I guess, thanks!

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Taken another look and now I don't understand what you're talking about 😄 
We basically need a setter for `onGeoJsonParsedListener` because in `MapboxMap.onFinishLoadingStyleExtension` we need to set new listener if necessary. 
At the same time I don't want bringing in generic class because all other sources are sync and it's pointless right now adding those abstractions.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: As discussed - not viable because `CompletableFuture` requires higher min API and using just `Future` is not sufficient.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: I think no. This function is responsible for decreasing the counter and actually does not have any information if geojson was already loaded until this point. 
We could add `onGeoJsonParsedListener` but it was already triggered beforehand.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Revisited.

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleContract.kt
kiryldz: Done.

extension-style/src/main/java/com/mapbox/maps/extension/style/StyleExtensionImpl.kt
kiryldz: Applied.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -467,38 +501,81 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+  ): GeoJsonSource = apply {
+    onDataParsed?.let { listener ->
+      workerHandler.post {
+        val property = data.toPropertyValue()
+        mainHandler.post {
+          setProperty(property)
+          listener.invoke(this)
+        }
+      }
+    } ?: setProperty(data.toPropertyValue())
```
pengdev, line 563: if no `onDataParsed` is passed, does it mean we use synchronized geojson parsing?

kiryldz: Yes, it is stated in docs above.
Please also keep in mind that synchronous mode is available here if updating `GeoJsonSource` (kinda supporting back compatibility) but setting via `GeoJsonSource.Builder` will be always async.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,167 @@
...
+            }
+            +symbolLayer("layer_marker", "source_marker") {
+              iconImage("icon")
+              iconAnchor(IconAnchor.BOTTOM)
+            }
+          }
+        ) {
+          loadAdditionalGeoJsonAfter(it)
+        }
+      }
```
tobrun, line 103: nit duplicate values, could be replaced with a constant

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
tobrun: :+1: eventhough majority of users would want async, I have heard about customers wanting to have the non-async method.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -1116,9 +1207,39 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+ * fully parsed [GeoJsonSource] is returned in [onGeoJsonParsed] callback.
+ *
+ * Using this method means that it is user's responsibility to proceed with adding layers or
+ * other style objects in [onGeoJsonParsed] callback.
  */
-fun geoJsonSource(id: String, block: GeoJsonSource.Builder.() -> Unit): GeoJsonSource =
-  GeoJsonSource.Builder(id).apply(block).build()
+fun geoJsonSource(
+  id: String,
+  block: GeoJsonSource.Builder.() -> Unit,
```
tobrun, line 1241: `block` naming is well chosen but I'm wondering if calling it async would be even more clear:
```
    geoJsonSource(
      id = "source5",
      async = {
        featureCollection(routePoints)
      },
      onGeoJsonParsed = {
        style.addSource(it)
        style.addLayer(
          lineLayer("layer5", "source5") {
            lineColor("green")
            lineOffset(20.0)
          }
        )
      }
    )
```

tobrun: actually the block itself isn't executed on asynchronously, only the result coming out of it is asynchronously parsed. So  `onAsyncParse` naming would make more sense?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -361,7 +361,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
         annotation
       }
 
-      geoJsonSource.featureCollection(FeatureCollection.fromFeatures(features))
+      geoJsonSource.featureCollection(FeatureCollection.fromFeatures(features), onDataParsed = {})
```
tobrun, line 364: can `onDataParsed` be made optional so we don't need to provide an empty implementation?

tobrun (commented):
Like the approach taken!

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -1116,9 +1207,39 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+  onGeoJsonParsed: (GeoJsonSource) -> Unit
```
pengdev, line 1242: shall we make the onGeoJsonParsed null by default?
```suggestion
  onGeoJsonParsed: ((GeoJsonSource) -> Unit)? = null
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Well, all the properties are applied synchronously except of setting `data`. Data could be represented by `feature`, `featureCollection` or `geometry` while all of them are actually representing `GeoJson`. Thus I think that `block` may be renamed to `builderBlock` and `onGeoJsonParsed` sounds OK. Using word `async` here is pretty pointless because docs are clearly stating that this function is async in any case.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: It is optional and `null` is used as default providing sync implementation.
Reason for using empty callback here is to force using async mode that actually uses one worker thread under the hood and message queue with `post`. 
That actually means that loading initial style + updating here is synchronised on worker thread.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: I was trying it, sadly breaks all our existing code if using it that way.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: Example will be revisited for sure, thanks for heads up!

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -21,7 +25,37 @@ import com.mapbox.maps.extension.style.utils.toValue
...
+          geoJsonParsed = true
+          onGeoJsonParsedListenerList.forEach {
+            it.invoke(this)
+          }
+        }
+      }
+    }
+  }
+
+  fun addOnGeoJsonParsedListener(listener: (GeoJsonSource) -> Unit) {
```
pengdev, line 56: shall we have a method to remove listener?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -21,7 +25,37 @@ import com.mapbox.maps.extension.style.utils.toValue
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+class GeoJsonSource(
+  builder: Builder
+) : Source(builder.sourceId) {
+
+  var geoJsonParsed = false
```
pengdev, line 32: could this be internal? users could always use addOnGeoJsonParsedListener to get the parsed geojson source, if the source is already parsed, we can emit the parsed geojsonsource immediately.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
pengdev: if we make the `geoJsonParsed` private like proposed in  onhttps://github.com/mapbox/mapbox-maps-android/pull/327/files#r634347933 , we could always use the `onGeoJsonParsedListener` and decrease the counter inside the `onGeoJsonParsedListener`.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: hmmm, perhaps I see your point, let me try it out.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: it is more of internal logic but I guess we could add remove method

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -1116,9 +1207,39 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+ *
+ * loadStyle(style(Style.DARK) {
+ *   +geoJsonSource(id) {
+ *    featureCollection(collection)
+ *   }
+ *   ...
+ * }
+ *
+ * compositing style will be performed correctly under the hood.
+ * [Style.OnStyleLoaded] will be emitted in correct moment of time when all sources are parsed.
```
pengdev, line 1224: these only applies to using the StyleExtension object to load the style. If uses just use this single method, we can not guarantee the [Style.OnStyleLoaded] will be emitted in correct moment of time.

We could explain the behaviour like: 
Immediately returns [GeoJsonSource] with no data set and starts preparing actual data using a worker thread, and call the `setData` after the parsing is finished.

And this sync processing also doesn't apply to geoJson urls, it only applies to the Feature, FeatureCollection and  Geometry types.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
pengdev: ok, now we have two different methods for geoJsonSource, so I guess we can leave out the default value, as it will have generate same Java signature like
```
fun geoJsonSource(  id: String,  block: GeoJsonSource.Builder.() ->; Unit): GeoJsonSource
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Updated docs to state it clearer.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Applied, thanks!

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: Added this function.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_simple_map)
+
+    routePoints = FeatureCollection.fromFeature(
+      Feature.fromGeometry(
+        LineString.fromPolyline(
+          DirectionsResponse.fromJson(
+            AnnotationUtils.loadStringFromAssets(
```
Chaoba, line 44: Instead of including this large file in the repo, you can use the same geojson URL with the one in `PointAnnotationClusterActivity` which contains 10k points.

Chaoba (approved):
Left one comments, others LGTM

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
tobrun: for testing these changes in this PR that setup wont work

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -827,21 +919,33 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+    private val workerThread by lazy {
+      HandlerThread("STYLE_WORKER").apply {
+        priority = Thread.MAX_PRIORITY
+        start()
+      }
+    }
+    private val workerHandler by lazy {
+      Handler(workerThread.looper)
+    }
+    private val mainHandler = Handler(Looper.getMainLooper())
```
tobrun, line 948: Love that we are sharing a HandlerThread across geojsonsource instances but it seems it will run forever once used? Should we introduce management of it and call https://developer.android.com/reference/android/os/HandlerThread#quit() ?

kiryldz: @tobrun I completely agree, but personally I do not have an idea how it should be managed in general. What should be a correct moment to call `quit` for a handler thread? It's not tied to `MapView` lifecycle as was mentioned [here](https://github.com/mapbox/mapbox-maps-android/pull/306#discussion_r627292315).
I think that leaving worker thread alive (which is basically doing no work) is better in that case then introducing management which might bring in some corner situations. However I'm happy to discuss, we might revisit it as tailwork.

app/src/main/java/com/mapbox/maps/testapp/examples/AnimatedMarkerActivity.kt
```diff
@@ -75,15 +75,11 @@ class AnimatedMarkerActivity : AppCompatActivity(), OnMapClickListener {
...
     }
-    var tick = 0
     animator = ValueAnimator().apply {
       setObjectValues(currentPoint, point)
       setEvaluator(pointEvaluator)
       addUpdateListener {
-        // TODO remove throttle when https://github.com/mapbox/mapbox-maps-android/issues/554 fixed
-        if (++tick % 2 == 0) {
-          geojsonSource.geometry(it.animatedValue as Point)
-        }
```
pengdev, line 86: by processing the GeojsonSouce async on platform, does it solve the issue that frequent GeoJsonSource update not showing up in the UI?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
tobrun: what about using coroutines and push the thread management issue to the language? :)

pengdev:
Can we also add a gif showcasing the newly added large geojson parsing activity?

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+          loadAdditionalGeoJsonAfter(it)
+        }
+      }
+  }
+
+  private fun loadAdditionalGeoJsonAfter(style: Style) {
+    // async method with adding layer in callback when geojson is completely parsed
+    geoJsonSource(
+      id = "${SOURCE}_$LARGE_SOURCE_COUNT",
+      block = {
```
pengdev, line 105: shall we rename the block as configs to be more specific?

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -0,0 +1,163 @@
...
+        style.addSource(it)
+        style.addLayer(
+          lineLayer("${LAYER}_$LARGE_SOURCE_COUNT", "${SOURCE}_$LARGE_SOURCE_COUNT") {
+            lineColor("green")
+            lineOffset(5.0 * LARGE_SOURCE_COUNT)
+          }
+        )
+      }
+    )
+    // async method with adding layer instantly without waiting for geojson to actually parse
```
pengdev, line 118: As far as I remember, without specifying the onDataParsed closure, the geojson parsing is processed in sync manner?

kiryldz: No, because we're using DSL function here. Sync mode will be if:
```
val source = style.getSourceAs(id)
source.feature(someFeature) // this will be sync
source.feature(someFeature, {}) // this will be async
```

app/src/main/java/com/mapbox/maps/testapp/examples/AnimatedMarkerActivity.kt
kiryldz: Basically it's already fixed in latest `main`. But this async mode should work even better I guess as we use single worker thread and because of that we preserve correct order of updates.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
pengdev: shall we rename async to for example `configs =` ?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -1116,9 +1220,43 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+ *
+ * Using this method means that it is user's responsibility to proceed with adding this source,
+ * layers or other style objects in [onGeoJsonParsed] callback.
  */
-fun geoJsonSource(id: String, block: GeoJsonSource.Builder.() -> Unit): GeoJsonSource =
-  GeoJsonSource.Builder(id).apply(block).build()
+fun geoJsonSource(
+  id: String,
+  block: GeoJsonSource.Builder.() -> Unit,
+  onGeoJsonParsed: (GeoJsonSource) -> Unit
```
pengdev, line 1259: shall we use a default value `{}` here? And document that user can optionally handle the callback?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -169,18 +170,40 @@ class MapboxMap internal constructor(
...
     this.style = style
-    styleExtension.images.forEach {
-      it.bindTo(style)
-    }
+    var resourceCount = styleExtension.resourceCount
     styleExtension.sources.forEach {
+      if (it is GeoJsonSource) {
+        it.bindTo(style)
+        it.addOnGeoJsonParsedListener {
+          resourceCount--
```
pengdev, line 178: shall we remove the `OnGeoJsonParsedListener` after the parsing is finished?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: we have another DSL overloaded function `fun geoJsonSource(id: String, block: GeoJsonSource.Builder.() -> Unit)`, why do we want default parameter here?

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
pengdev: I think below code(`style.addSource()`) doesn't use StyleExtension, so it won't be async?

app/src/main/java/com/mapbox/maps/testapp/examples/AnimatedMarkerActivity.kt
pengdev: 👍 this is a great fix

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
pengdev: configs/settings for example

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
pengdev: ok, right

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: Adding source of course is sync but parsing is still async. Whenever `GeoJsonSource.Builder` is used - we will always perform geojson parsing using worker thread.
The only way to parse featureCollection etc on main thread is to call method `featureCollection` directly on `GeoJsonSource` object.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
pengdev: in case the source will be modified later by users, and this block will still be called.

kiryldz: I don't think it's that important in this case tbh. We can't do it right now because of there's no direct reference to anonymous lambda.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: So in this case we immediately add source to style (with no data), connect it with layer and when data is parsed in async mode we update to source and data becomes visible. Main difference with not using `onGeoJsonParsed` here is that we do not know exact moment of time when actual data becomes visualised, it will simply become visualised ASAP.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: OK, I'll remove it, you're correct.

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
pengdev: ok, now I remember.. We probably need to provide some guide on the GeoJson parsing, since this is getting complex 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -1116,9 +1220,43 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+ * [Style.OnStyleLoaded] will be emitted in correct moment of time when all sources are parsed.
+ *
+ * If creating geojson sources for already loaded Style please consider using overloaded
+ * geoJsonSource(String, GeoJsonSource.Builder.() -> Unit, onGeoJsonParsed: (GeoJsonSource) -> Unit) function
+ * and use fully prepared [GeoJsonSource] in onGeoJsonParsed callback.
+ */
+fun geoJsonSource(
+  id: String,
+  block: GeoJsonSource.Builder.() -> Unit
+): GeoJsonSource = GeoJsonSource.Builder(id) {}.apply(block).build()
```
pengdev, line 1246: the behaviour change should be considered as a breaking change and requires user's adoption, worth mentioning separately in the change log.

pengdev: sorry, found it in the PR description, we could copy that description to the change log.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: yes, will do for overloaded function where both lambda blocks are used for geojson, thanks!

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
kiryldz: Thanks, but le'ts go with current solution. I do not want to depend heavily on network + have some viable routes.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: I guess it should be done a bit later, not in this PR, WDYT?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -169,18 +171,45 @@ class MapboxMap internal constructor(
...
+        callback = OnGeoJsonParsed { _ ->
+          resourceCount--
+          if (resourceCount == 0) {
+            onStyleLoaded?.onStyleLoaded(style)
+          }
+          callback?.let { callback ->
+            it.removeOnGeoJsonParsedListener(callback)
+          }
+        }
+        it.addOnGeoJsonParsedListener(callback)
```
pengdev, line 188:
```suggestion
        it.addOnGeoJsonParsedListener(object : OnGeoJsonParsed {
          override fun onGeoJsonParsed(source: GeoJsonSource) {
            resourceCount--
            if (resourceCount == 0) {
              onStyleLoaded?.onStyleLoaded(style)
            }
            it.removeOnGeoJsonParsedListener(this)
          }
        })
```

pengdev (approved):
LGTM % nit

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: idk how I missed it, thanks!

Issue #291, opened 2021-04-27
Crash when changing the style to one with no layers matching previous annotation manager belowLayerId


cramptonism:
## Environment

- Android OS version: API 30
- Devices affected: Emulator and Pixel 5, presumably more
- Maps SDK Version: 10.0.0-beta.17 

## Observed behavior and steps to reproduce

```
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err: java.lang.RuntimeException: Add layer failed: Cannot find layer with id: pitch-outline
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.extension.style.layers.Layer.bindTo(Layer.kt:114)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.extension.style.layers.LayerKt.addLayerBelow(Layer.kt:231)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.plugin.annotation.AnnotationManagerImpl.initLayerAndSource(AnnotationManagerImpl.kt:185)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.plugin.annotation.AnnotationManagerImpl.onStyleLoaded(AnnotationManagerImpl.kt:257)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.plugin.annotation.AnnotationPluginImpl.onStyleChanged(AnnotationPluginImpl.kt:58)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.plugin.MapPluginRegistry.onStyleChanged(MapPluginRegistry.kt:164)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.MapController$2$1.onStyleLoaded(MapController.kt:69)
2021-04-26 17:29:09.485 12218-12218/com.mapbox.maps.testapp W/System.err:     at com.mapbox.maps.MapboxMap.onFinishLoadingStyle$sdk_debug(MapboxMap.kt:212)
[...]

PR #330, opened 2021-05-12
Add removeAnnotationManager API;  Fix crash if the belowLayerId doesn't exist on the current style

Chaoba:
Fixes: #291 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add removeAnnotationManager API; Fix crash if the belowLayerId doesn't exist on the current style</changelog>`.

### Summary of changes

- Add removeAnnotationManager API to enable users remove the useless annotation manager and the inside layer and source

- If user set the belowLayerId in config and the current style doesn't have this layer, it will crash while adding the annotation layer. This pr adds the check for this below layer and will add the annotation layer directly if the below layer doesn't esist.

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PolylineAnnotationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PolylineAnnotationActivity.kt
@@ -8,6 +8,7 @@ import com.mapbox.geojson.FeatureCollection
 import com.mapbox.geojson.Point
 import com.mapbox.maps.extension.style.layers.getLayer
 import com.mapbox.maps.plugin.annotation.AnnotationConfig
+import com.mapbox.maps.plugin.annotation.AnnotationPlugin
 import com.mapbox.maps.plugin.annotation.annotations
 import com.mapbox.maps.plugin.annotation.generated.*
 import com.mapbox.maps.testapp.R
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PolylineAnnotationActivity.kt
```diff
@@ -101,7 +103,12 @@ class PolylineAnnotationActivity : AppCompatActivity() {
       }
     }
 
-    deleteAll.setOnClickListener { lineManager?.deleteAll() }
+    deleteAll.setOnClickListener {
+      lineManager?.let {
+        it.deleteAll()
```
tobrun, line 108: is there a reason why we need to delete all annotations?

plugin-annotation/src/test/java/com/mapbox/maps/plugin/annotation/AnnotationPluginImplTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import junit.framework.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(shadows = [ShadowValueConverter::class, ShadowLogger::class])
+class AnnotationPluginImplTest {
```
tobrun, line 31: for future reference, you can use `mockk(relaxed = true)` to fallback on default values and not needing to guard/implement every calls

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPlugin.kt
```diff
@@ -22,4 +22,9 @@ interface AnnotationPlugin : MapPlugin, MapSizePlugin, MapStyleObserverPlugin {
     type: AnnotationType,
     annotationConfig: AnnotationConfig?
   ): AnnotationManager<*, *, *, *, *, *, *>
+
+  /**
+   * Remove an annotation manger that created before, this will remove the inside layer and source too.
```
tobrun, line 27: suggest to rephrase as `Removes an annotation manager, this will remove the underlying layer and source from the style.`

tobrun (approved):

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PolylineAnnotationActivity.kt
Chaoba: No, we can delete this line.

PR #332, opened 2021-05-12
[style] validate fully loaded event of style

tobrun:
## Pull request checklist:

### Summary of changes
We currently maintain style loaded state in the SDK to workaround an upstream issue. 
This PR validates if we can remove that state checking from our end.


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -95,9 +95,7 @@ class MapboxMap internal constructor(
     onMapLoadErrorListener: OnMapLoadErrorListener? = null
   ) {
     initializeStyleLoad(onStyleLoaded, onMapLoadErrorListener)
-    handlerMain.post {
-      nativeMapWeakRef.call { (this as StyleManagerInterface).styleURI = styleUri }
-    }
+    nativeMapWeakRef.call { (this as StyleManagerInterface).styleURI = styleUri }
```
kiryldz, line 98: A bit concerned here.
We're removing handler logic here but leaving it in `sdk/src/main/java/com/mapbox/maps/NativeObserver.kt`. If I remember correctly some some metrics benchmarks were having a deadlock because of not posting here.

kiryldz: If tests pass - we can follow up later.

tobrun: agreed, let me dig up why we have that construct

kiryldz (approved):
LGTM but better check metrics before merging.

tobrun:
RC.3 tag landed, going to merge to be included in RC.4

Issue #323, opened 2021-05-11
Location not shown when accepting permissions


tobrun:
Issue: no location is visualized until I call disable/enable again 

https://user-images.githubusercontent.com/2151639/117802290-982d4500-b255-11eb-8f88-ce419788abe0.mp4

---------------

To be investigated if it's related to how we handle permissions in our test app, how test activity was setup or if it's an issue with the internal working of location component itself

Issue #324, opened 2021-05-11
Location shows null island 


tobrun:
null island:

> Null Island is the name used to refer to the point on the Earth's surface where the Prime Meridian and the Equator intersect, at zero degrees latitude and zero degrees longitude (0°N 0°E)

When running an activity with location component, without having a previous location fix. It shows the location at null island. This doesn't result in the best UX and is something we would want to optimize for.

https://user-images.githubusercontent.com/2151639/117802809-328d8880-b256-11eb-9e6f-272ae1ed38b0.mp4

Note that this video also shows the issue from #323

PR #333, opened 2021-05-13
Auto active LocationComponent while style loaded.

Chaoba:
Fixes: #323 #324 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Enable LocationComponent automatically when style loaded; fix null island location puck</changelog>`.

### Summary of changes

- The root cause for location not shown when accepting permissions is we need to get the style to active location component 
and the style has not been set yet while init location component.
This pr listen to the style load event and trigger `applySettings ` to reactive location component and apply a loop request to get the location update after location permission is granted.

- This pr also make the location puck invisible before getting the real location from the device to fix the null island issue

### DIFF
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
@@ -8,6 +8,7 @@ import com.mapbox.android.core.location.LocationEngineCallback
 import com.mapbox.android.core.location.LocationEngineProvider
 import com.mapbox.android.core.location.LocationEngineRequest
 import com.mapbox.android.core.location.LocationEngineResult
+import com.mapbox.android.core.permissions.PermissionsManager
 import com.mapbox.common.Logger
 import com.mapbox.geojson.Point
 import java.util.concurrent.CopyOnWriteArrayList
@@ -15,7 +16,7 @@ import java.util.concurrent.CopyOnWriteArrayList
 /**
  * Default Location Provider implementation, it can be overwritten by users.
  */
-internal class LocationProviderImpl(context: Context) : LocationProvider, LocationEngineCallback<LocationEngineResult> {
+internal class LocationProviderImpl(private val context: Context) : LocationProvider, LocationEngineCallback<LocationEngineResult> {
   private val locationEngine = LocationEngineProvider.getBestLocationEngine(context)
 
   private val locationEngineRequest =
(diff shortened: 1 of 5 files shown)


plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -28,9 +29,11 @@ internal class LocationProviderImpl(context: Context) : LocationProvider, Locati
 
   @SuppressLint("MissingPermission")
   private fun requestLocationUpdates() {
-    locationEngine.requestLocationUpdates(
-      locationEngineRequest, this, Looper.getMainLooper()
-    )
+    if (PermissionsManager.areLocationPermissionsGranted(context)) {
```
pengdev, line 32: shall we add some error log when the permission is not granted at this point of time?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -76,7 +79,9 @@ internal class LocationProviderImpl(context: Context) : LocationProvider, Locati
       requestLocationUpdates()
     }
     locationConsumers.add(locationConsumer)
-    locationEngine.getLastLocation(this)
+    if (PermissionsManager.areLocationPermissionsGranted(context)) {
```
pengdev, line 82: same as above

pengdev (approved):
Great catch, LGTM % comments

kiryldz (approved):

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
Chaoba: Yes, will do.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -26,11 +29,26 @@ internal class LocationProviderImpl(context: Context) : LocationProvider, Locati
...
+      )
+    } else {
+      if (handler == null) {
+        handler = Handler()
+        runnable = Runnable { requestLocationUpdates() }
+      }
+      handler?.postDelayed(runnable, UPDATE_DELAY)
+      Logger.w(
+        TAG,
+        "Missing location permission, location component will not take effect before location permission is granted."
```
tobrun, line 49: I'm ok with a loop that verifies if permissions have been expected but 2 remarks:
 - shouldn't we clean up the handler when the location/map is destroyed (we could end up with a loop that continuous forever)
 - can we apply an exponential backoff + decrease the initial UPDATE_DELAY so it eventually grows to the current value?

Chaoba: The runnable will be removed in `unRegisterLocationConsumer` which would be invoked while deactivateLocationComponent

tobrun (approved):

PR #352, opened 2021-05-20
Make sure to execute first camera update

LukasPaczos:
Fixes: #363 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Update map camera on first animator update.</changelog>`.

### Summary of changes

The camera plugin impl executes camera jumps on a tick of first animator. The animator is registered _also_ when it first ticks, but it was registered last. Because the logic to execute the updates checks for the registered animators, and the animators was registered last in that logical block, the first tick of the animation was effectively ignored.

This skipped update was causing a little bit of desynchronization when paired with other animations (like location puck animation)

### User impact (optional)

Example before (notice puck moving forward and back when animations starts):

https://user-images.githubusercontent.com/16925074/118990206-f7403780-b982-11eb-9b7d-ad035e5be161.mp4

Example after:

https://user-images.githubusercontent.com/16925074/118990221-fad3be80-b982-11eb-94ca-8c3139275e36.mp4

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -303,6 +303,11 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
   private fun registerInternalUpdateListener(animator: CameraAnimator<*>) {
     animator.addInternalUpdateListener {
+      // set current animator value
+      updateCameraValue(animator)
+      // add current animator to queue-set if was not present
+      runningAnimatorsQueue.add(it)
+
       // main idea here is not to update map on each option change
       // we perform jump based on update tick of first (oldest) animation
       val firstAnimator = if (runningAnimatorsQueue.iterator().hasNext()) {
@@ -333,10 +338,6 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
         // reset values
         cameraOptionsBuilder = CameraOptions.Builder()
       }
-      // set current animator value
-      updateCameraValue(animator)
-      // add current animator to queue-set if was not present
-      runningAnimatorsQueue.add(it)
     }
   }
 
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -207,6 +207,21 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     }
   }
 
+  private fun correctInitialCamera(
+    currentCameraBuilder: CameraOptions.Builder,
+    cameraAnimator: CameraAnimator<*>
+  ): CameraOptions {
+    return when (cameraAnimator) {
+      is CameraCenterAnimator -> currentCameraBuilder.center(cameraAnimator.animatedValue as? Point).build()
+      is CameraZoomAnimator -> currentCameraBuilder.zoom(cameraAnimator.animatedValue as? Double).build()
+      is CameraAnchorAnimator -> currentCameraBuilder.anchor(cameraAnimator.animatedValue as? ScreenCoordinate).build()
+      is CameraPaddingAnimator -> currentCameraBuilder.padding(cameraAnimator.animatedValue as? EdgeInsets).build()
+      is CameraBearingAnimator -> currentCameraBuilder.bearing(cameraAnimator.animatedValue as? Double).build()
+      is CameraPitchAnimator -> currentCameraBuilder.pitch(cameraAnimator.animatedValue as? Double).build()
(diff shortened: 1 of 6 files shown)


LukasPaczos:
This doesn't resolve all synchronization issues (you can even see a jump in the second video) but does improve the initial frame behavior.

LukasPaczos:
Your changes look good to me @kiryldz:
Before:

https://user-images.githubusercontent.com/16925074/119349824-ec491800-bc9e-11eb-8546-443e30ff6593.mp4

After (center of the puck is positioned correctly in the frame):

https://user-images.githubusercontent.com/16925074/119349839-efdc9f00-bc9e-11eb-82d8-231e90eb22dc.mp4

However, this only works because we specify the `center` animator as the first animator. Since in this specific use case only the lat-lng synchronization is important to us, we can get away with this. This is an issue that could be exposed in different ways in the future though if there are timing-critical transitions performed because we effectively ignore the rate of all other updates besides the first animator. The tradeoff of precision vs performance (battery consumption, etc) might be worth it in the majority of the use cases, but this is something that should be documented, and maybe even possible to be disabled.

Here's an example of this codebase where the center animator is the last animator in the list:

https://user-images.githubusercontent.com/16925074/119352641-38498c00-bca2-11eb-8cb2-93fc5ec946eb.mp4

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -21,8 +21,12 @@ import kotlin.properties.Delegates
  * It is responsible for:
  *   - Storing all the [ValueAnimator]'s that could be run. Only specific camera animators could be used in this plugin.
  *   - Controlling animation execution - only one [ValueAnimator] of certain type could be run at a time.
- *   If some another animation with same [CameraAnimator.type] is about to start previous one will be cancelled.
+ *     If some another animation with same [CameraAnimator.type] is about to start previous one will be cancelled.
  *   - Giving possibility to listen to [CameraOptions] values changes during animations via listeners.
+ *   - If several animations of different [CameraAnimator.type] are running simultaneously map camera
+ *     will be updated only on oldest animation update (the oldest is the one that was started first).
+ *     That actually means that animation start order matters. High-level animations [flyTo] and [easeTo]
+ *     will always trigger camera center animator first.
```
LukasPaczos, line 29: Would be great to also explain why (grouping updates to save some CPU time).

BTW, is this actually still necessary if the Choreographer is now coordinating when a frame is rendered? In the past, each change to GL-Native (like `setCamera`) would request a new frame, but that isn't the case anymore, is it?
Is this grouping of updates still necessary in this case to save a little bit of C++ code from running, if those updates do not trigger a frame re-render anyway? Did you have a chance to measure the performance difference versus calling `setCamera` on each animation tick?

kiryldz: @LukasPaczos we need to take into account that we reduce JNI calls in any case now + `setCamera` impl in C++ is also not that trivial so this optimisation __should__ be a bit valuable. 
I really liked your idea to add some kind of flag to turn this optimisation on / off. I'll create a follow-up task and measuring could organically happen there. WDYT?

kiryldz: https://github.com/mapbox/mapbox-maps-android/issues/366

LukasPaczos: Sound good :+1:

LukasPaczos (commented):
LGTM, I cannot :green_apple: my own PR though :joy:

sdk-base/src/main/java/com/mapbox/maps/ExtensionUtils.kt
```diff
@@ -41,14 +41,25 @@ fun MapSnapshotInterface.bitmap(): Bitmap {
 
 /**
  * Extension function to convert [CameraState] to [CameraOptions].
+ *
+ * If [anchor] is specified - [CameraOptions.center] will be explicitly set to NULL
+ * in order for the anchor to apply to the map camera.
```
pengdev, line 46: Is this also helpful to users that do not use animation plugin? If this change is only for the animation plugin, shall we make a separate  private/internal  extension function within the animations plugin?

pengdev (approved):
LGTM with one question.

sdk-base/src/main/java/com/mapbox/maps/ExtensionUtils.kt
kiryldz: There is no sense in setting both `anchor` and `center` at same time (in that case `anchor` is ignored). We did not face such issues before in our gestures because we were actually performing one more redundant initial camera update on `easeTo(duration = 0)`. So this `CameraState(anchor) <-> CameraOptions` should always look like the way it looks like now.

kiryldz (approved):

PR #354, opened 2021-05-20
Add 10.0.0-beta.20 API documentation.

MapboxCI:
cc: @mapbox/maps-android


danswick (approved):
This looks good to me locally! As usual we'll need to bump the publisher branch and then run tests for the docs site after this merges. cc @Guardiola31337

PR #365, opened 2021-05-24
Fix crash due to missing access token; delete CredentialsManager.kt

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix crash due to missing access token; delete CredentialsManager.kt</changelog>`.

### Summary of changes
While users only set token in attr of MapView, a `MapboxConfigurationException`  will be thrown as we can't find token from resource file and default value. This pr applies the token from attr to fix it.


sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +26,21 @@ class ResourceOptionsManager(
...
-    private lateinit var default: ResourceOptionsManager
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal var default: ResourceOptionsManager? = null
 
     /**
      * The default shared instance with default resource options.
      */
     @Synchronized
-    fun getDefault(context: Context): ResourceOptionsManager {
-      if (!this::default.isInitialized) {
```
pengdev, line 34: why not using lateinit var and use nullable variable instead?

sdk/src/main/java/com/mapbox/maps/ResourcesAttributeParser.kt
```diff
@@ -16,17 +16,11 @@ internal object ResourcesAttributeParser {
...
-    val builder = ResourceOptionsManager.getDefault(context).resourceOptions.toBuilder()
 
-    if (typedArray.hasValue(R.styleable.mapbox_MapView_mapbox_resourcesAccessToken)) {
-      builder.accessToken(
-        typedArray.getString(R.styleable.mapbox_MapView_mapbox_resourcesAccessToken)
-          ?: credentialsManager.getAccessToken(context)
-      )
-    }
+    val token = typedArray.getString(R.styleable.mapbox_MapView_mapbox_resourcesAccessToken)
+    val builder = ResourceOptionsManager.getDefault(context, token).resourceOptions.toBuilder()
```
pengdev, line 23: what if the user has already used the `ResourceOptionsManager.getDefault(context)` previously, the ResourceOptionsManager.default is set, and this method won't use the token from the configured xml attribute.

We should add some tests to validate the correct decision tree.

pengdev:
Also thinking the ResourceOptionsManager stores the static default ResourceOptions, and access token is part of the ResourceOptions, so feels like CredentialsManager is redundant, shall we discuss whether we should remove the CredentialsManager?

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: Lateinit will cache the default value and thus fail the test cases.

sdk/src/main/java/com/mapbox/maps/ResourcesAttributeParser.kt
Chaoba: Please refer to this [line](https://github.com/mapbox/mapbox-maps-android/pull/365/files#diff-fcbcc1ee0a08a0b88a933c358ca04796a5731db3cbb0696bc926635783e65f72R110) the current token has the highest priority

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
pengdev: We can overwrite the default ResourceOptionsManager#resourceOptions with default parameters after test.

sdk/src/main/java/com/mapbox/maps/ResourcesAttributeParser.kt
pengdev: I meant if the `if (default == null) {` is false, we will not get to the `applyDefaultParams`.

sdk/src/main/java/com/mapbox/maps/CredentialsManager.kt
```diff
@@ -1,51 +0,0 @@
...
-import android.content.Context
-
-/**
- * Convenience class that holds [MapboxMap] related token.
- * `CredentialsManager` could be created per `MapView` instance if given view should use some specific token.
- * Most common use-case when all `MapView`s should use same token could be handled by using `CredentialsManager.default` static object.
- *
- * @property accessToken the token will be applied.
- */
-data class CredentialsManager(private var accessToken: String? = null) {
```
kiryldz, line 12: Is this deletion aligned wit iOS? AFAIR we added it as part of alignment

Chaoba: Not yet, will align with iOS later

kiryldz: I guess it will be logical to first align change with iOS and then proceed with merging. We can discuss it today with voice.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -20,23 +20,41 @@ class ResourceOptionsManager(
     resourceOptions = resourceOptions.toBuilder().apply(block).build()
   }
 
+  /**
```
pengdev, line 23: ResourceOptionsManager should be a data class to be aligned with CredentialsManager.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -20,23 +20,41 @@ class ResourceOptionsManager(
...
+   * Reset all the properties of the inside resourceOptions to default
+   *
+   * @param context the application context
+   * @param defaultToken the token will applied as default during reset
+   */
+  fun reset(context: Context, defaultToken: String? = null) {
+    resourceOptions = ResourceOptions.Builder().applyDefaultParams(context, defaultToken).build()
```
pengdev, line 30: I'm thinking we should use previously set accesstoken as a fallback, instead of throwing runtimeException?

app/src/main/java/com/mapbox/maps/testapp/examples/MapViewCustomizationActivity.kt
```diff
@@ -21,11 +21,10 @@ class MapViewCustomizationActivity : AppCompatActivity() {
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    // Create  a custom CredentialsManager with a token and set it to CredentialsManager.shared, so that all MapViews created with default config will apply this token.
-    CredentialsManager.default.setAccessToken(getString(R.string.mapbox_access_token))
 
-    // Create a custom ResourceOptionsManager with customised tile store and tile size, so that all MapViews created with default config will apply these settings.
+    // Create a custom ResourceOptionsManager with customised token, tile store and tile size, so that all MapViews created with default config will apply these settings.
     ResourceOptionsManager.getDefault(this).update {
+      accessToken(getString(R.string.mapbox_access_token))
```
kiryldz, line 27: So do I get it correct that this token will be used by default for all other `MapView`s created afterwards?

Chaoba: That's right.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
pengdev: 👍 since the reset logic is pretty hidden(rarely used), just thinking it might be a surprise it will throw an exception here.

kiryldz (approved):
I'll drop 🚢  here as changes look ok to me and they fix a concrete bug. However perhaps we want to consider some alignment with iOS beforehand.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -91,10 +91,15 @@ fun MapOptions.Builder.applyDefaultParams(context: Context): MapOptions.Builder
 }
 
 /**
- * Get a default [ResourceOptions.Builder] with token from default [CredentialsManager]
- * @property context the context of the application.
+ * Get a default [ResourceOptions.Builder] with token from the token parameter or from the
+ * Android resources. If no token found, it will throw [MapboxConfigurationException].
+ *
+ * @param context the context of the application.
+ * @param token the default token for the builder, if not specified, the token will be fetched from Android resources. Will throw [MapboxConfigurationException] if there's no token found.
```
pengdev, line 98: the doc should be updated.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,41 @@ class ResourceOptionsManager(
...
+    internal var default: ResourceOptionsManager? = null
+    internal fun getTokenResId(context: Context): Int = context.resources.getIdentifier(
+      MAPBOX_ACCESS_TOKEN_RESOURCE_NAME,
+      "string",
+      context.packageName
+    )
+
+    fun reset() {
+      default = null
+    }
```
pengdev, line 36: On iOS, this is an internal function that's only used for testing purpose, shall we do the same?

cc @julianrex

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,41 @@ class ResourceOptionsManager(
...
+        } else {
+          val tokenResId = getTokenResId(context)
+          // Throw exception as no token could be found as default token.
+          if (tokenResId == 0) throw MapboxConfigurationException()
+        }
+        default = ResourceOptionsManager(builder.build())
       }
-      return default
+      return default!!
     }
```
pengdev, line 62: We could do something like 
```
    fun getDefault(context: Context, defaultToken: String? = null): ResourceOptionsManager {
      return (default ?: run {
        val builder = ResourceOptions.Builder().applyDefaultParams(context)

        if (defaultToken != null) {
          // Apply the user setting token as the default token
          builder.accessToken(defaultToken)
        } else {
          val tokenResId = getTokenResId(context)
          // Throw exception as no token could be found as default token.
          if (tokenResId == 0) throw MapboxConfigurationException()
        }
        ResourceOptionsManager(builder.build()).also {
          default = it
        }
      }).apply {
        // Always set the provided access token to default
        defaultToken?.let { token ->
          update { accessToken(token) }
        }
      }
    }
  }
```

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: Yes, there is no reason for a user to destroy it.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: Thanks for the suggestion, have changed to a more clear logic.

app/src/main/java/com/mapbox/maps/testapp/examples/MapViewCustomizationActivity.kt
pengdev: @Chaoba here will throw exceptions if the default accesstoken does not exist in Android resources, probably we should move the setting of the access token to the ResourceOptionsManager.getDefault(this, getString(R.string.mapbox_access_token))

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,51 @@ class ResourceOptionsManager(
    * Static variables and methods.
    */
   companion object {
-    private lateinit var default: ResourceOptionsManager
+    internal var default: ResourceOptionsManager? = null
```
pengdev, line 27: nit: can be private?

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,51 @@ class ResourceOptionsManager(
...
-        default = ResourceOptionsManager(
-          ResourceOptions.Builder().applyDefaultParams(context)
-            .build()
-        )
+    fun getDefault(context: Context, defaultToken: String? = null): ResourceOptionsManager {
+      // Apply the user setting token as the default token.
+      defaultToken?.let { token ->
+        val builder = ResourceOptions.Builder().applyDefaultParams(context).accessToken(token)
+        return ResourceOptionsManager(builder.build()).also { default = it }
       }
```
pengdev, line 55: There's a potential bug, if the default is not null, this will overwrite previously set default, right?

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,51 @@ class ResourceOptionsManager(
...
+
+      // No defaultToken provided, search in the resources to init default instance.
+      val tokenResId = getTokenResId(context).also {
+        // Throw exception as no token could be found as default token.
+        if (it == 0) throw MapboxConfigurationException()
+      }
+
+      // Build a new ResourceOptions with token from the resources.
+      val builder = ResourceOptions.Builder().applyDefaultParams(context)
+        .accessToken(context.getString(tokenResId))
```
pengdev, line 70: nit, the `applyDefaultParams` will already fill in the token from the Android resources.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
pengdev: we could add a test case for it.

Chaoba: Yes, this is desigend as we described in the doc `the default token will be set to the default [ResourceOptionsManager] instance`
We have a test for it `getDefaultWithUserTokenTest`

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: Yes, good catch.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
pengdev: If user has changed other settings, for example cacheSize, in side the ResourceOptionsManager.default previously, this will reset their previous custom setting to mapbox default.

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -91,10 +91,14 @@ fun MapOptions.Builder.applyDefaultParams(context: Context): MapOptions.Builder
 }
 
 /**
- * Get a default [ResourceOptions.Builder] with token from default [CredentialsManager]
- * @property context the context of the application.
+ * Get a default [ResourceOptions.Builder] with token from the token parameter or from the
```
pengdev, line 94:
```suggestion
 * Get a default [ResourceOptions.Builder] with token from the
```

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -91,10 +91,14 @@ fun MapOptions.Builder.applyDefaultParams(context: Context): MapOptions.Builder
...
+ * Android resources. If no token found, it will throw [MapboxConfigurationException].
```
pengdev, line 95:
```suggestion
 * Get a default [ResourceOptions.Builder] with Mapbox pre-defined options, and
 * with the access token taken from the Android resources.
```

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: Will reuse the previous config

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -91,10 +91,14 @@ fun MapOptions.Builder.applyDefaultParams(context: Context): MapOptions.Builder
 }
 
 /**
- * Get a default [ResourceOptions.Builder] with token from default [CredentialsManager]
- * @property context the context of the application.
+ * Get a default [ResourceOptions.Builder] with Mapbox pre-defined options, and
+ * with the access token taken from the Android resources.
```
pengdev, line 95:
```suggestion
 * with the access token taken from the Android resources(if available).
```

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,56 @@ class ResourceOptionsManager(
...
+      "string",
+      context.packageName
+    )
+
+    /**
+     * Convenience function to remove the default instance. Calling `getDefault` again will re-create the default instance.
+     */
+    fun destroyDefault() {
+      default = null
+    }
```
pengdev, line 39: @julianrex here we made this static `destroyDefault` function public to users, and removed the `reset` method on the `ResourceOptionsManager`, let's discuss and align it.

sdk/src/test/java/com/mapbox/maps/ResourceAttributeParserTest.kt
```diff
@@ -29,28 +30,34 @@ class ResourceAttributeParserTest {
...
     assertEquals("pk.foobar", resourceOptions.accessToken)
     assertEquals(null, resourceOptions.baseURL)
     assertEquals("/foobar/.mapbox/maps/ambient_cache.db", resourceOptions.cachePath)
     assertEquals(99L, resourceOptions.cacheSize)
   }
 
-  @Test(expected = MapboxConfigurationException::class)
-  fun noAccessTokenResource() {
+  @Test
+  fun noAccessTokenResourceHasTokenFromAttr() {
```
pengdev, line 49: shall we add a test that `noAccessTokenResourcesNoTokenFromAttr` and `hasAccessTokenResourcesHasTokenFromAttr` and `hasAccessTokenResourcesNoTokenFromAttr` ? seems a bit complicated, but I guess worth to add it since the logic(decision tree) here is complex.

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
```diff
@@ -60,5 +105,15 @@ class ResourceOptionsManagerTest {
...
+
+    ResourceOptionsManager.destroyDefault()
+    defaultResourceOptionsManager = ResourceOptionsManager.getDefault(context)
+    assertEquals("token", defaultResourceOptionsManager.resourceOptions.accessToken)
+    Assert.assertTrue(defaultResourceOptionsManager.resourceOptions.cachePath!!.endsWith("/.mapbox/maps/ambient_cache.db"))
+    assertEquals(DEFAULT_CACHE_SIZE, defaultResourceOptionsManager.resourceOptions.cacheSize)
+
+    ResourceOptionsManager.destroyDefault()
+    defaultResourceOptionsManager = ResourceOptionsManager.getDefault(context)
+    assertEquals("token", defaultResourceOptionsManager.resourceOptions.accessToken)
```
pengdev, line 117: Seems redundant?

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
```diff
@@ -21,28 +21,73 @@ class ResourceOptionsManagerTest {
...
+      cacheSize(1000L)
+    }
     assertEquals("token", defaultResourceOptionsManager.resourceOptions.accessToken)
     Assert.assertTrue(defaultResourceOptionsManager.resourceOptions.cachePath!!.endsWith("/.mapbox/maps/ambient_cache.db"))
-    assertEquals(DEFAULT_CACHE_SIZE, defaultResourceOptionsManager.resourceOptions.cacheSize)
+    assertEquals(1000L, ResourceOptionsManager.getDefault(context).resourceOptions.cacheSize)
+
+    defaultResourceOptionsManager = ResourceOptionsManager.getDefault(context, "newToken")
+    assertEquals("newToken", defaultResourceOptionsManager.resourceOptions.accessToken)
+    assertEquals(1000L, ResourceOptionsManager.getDefault(context).resourceOptions.cacheSize)
```
pengdev, line 77: could it be a separate test?

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
Chaoba: It's to verify the regenerated ResourceOptionsManager

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
Chaoba: Same as before, to verify getDefault with token will reuse the previous settings we can't separate it, must call `getDefault` twice.

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
```diff
@@ -60,5 +105,15 @@ class ResourceOptionsManagerTest {
       ResourceOptionsManager.getDefault(context).resourceOptions.tileStoreUsageMode
     )
     assertEquals("baseUrl", ResourceOptionsManager.getDefault(context).resourceOptions.baseURL)
+
+    ResourceOptionsManager.destroyDefault()
+    defaultResourceOptionsManager = ResourceOptionsManager.getDefault(context)
+    assertEquals("token", defaultResourceOptionsManager.resourceOptions.accessToken)
+    Assert.assertTrue(defaultResourceOptionsManager.resourceOptions.cachePath!!.endsWith("/.mapbox/maps/ambient_cache.db"))
+    assertEquals(DEFAULT_CACHE_SIZE, defaultResourceOptionsManager.resourceOptions.cacheSize)
```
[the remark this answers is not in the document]
Chaoba: Verify `getDefault` will generate the default settings after destroy.

sdk/src/test/java/com/mapbox/maps/ResourceOptionsManagerTest.kt
Chaoba: Yeah, this part could be deleted.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
```diff
@@ -24,20 +24,56 @@ class ResourceOptionsManager(
...
-          ResourceOptions.Builder().applyDefaultParams(context)
-            .build()
-        )
+    fun getDefault(context: Context, defaultToken: String? = null): ResourceOptionsManager {
+      // Apply the user setting token as the default token.
+      defaultToken?.let { token ->
+        default?.let {
+          // Reuse the config from previous default and only update the token
+          val builder = it.resourceOptions.toBuilder().accessToken(token)
+          return ResourceOptionsManager(builder.build()).also { default = it }
```
pengdev, line 56: Shall we make `update` return the updated ResourceOptionsManager instance and do the following?
```suggestion
          return it.update { accessToken(token) }
```

sdk/src/test/java/com/mapbox/maps/ResourceAttributeParserTest.kt
```diff
@@ -29,28 +30,57 @@ class ResourceAttributeParserTest {
...
+    every { resources.getIdentifier("mapbox_access_token", "string", "foobar") } returns 0
+    every { typedArray.getString(R.styleable.mapbox_MapView_mapbox_resourcesAccessToken) } returns "token"
+    val options = ResourcesAttributeParser.parseResourcesOptions(context, typedArray)
+    assertEquals("token", options.accessToken)
+  }
+
   @Test(expected = MapboxConfigurationException::class)
-  fun noAccessTokenResource() {
+  fun noAccessTokenResourceNosTokenFromAttr() {
```
pengdev, line 57:
```suggestion
  fun noAccessTokenResourceNoTokenFromAttr() {
```

pengdev (approved):
LGTM with nits

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
Chaoba: The return type of `update` method is Unit, we can't return it directly.

pengdev: we can make update return the updated `ResourceOptionsManager` instance, if it makes sense.

sdk/src/main/java/com/mapbox/maps/ResourceOptionsManager.kt
julianrex: Addressed in https://github.com/mapbox/mapbox-maps-ios/pull/406

PR #380, opened 2021-05-31
Geojson async data parsing: fixes and improvements

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Geojson async data parsing: fixes and improvements.</changelog>`.

### Summary of changes

This PR brings following changes to async geojson behavior:

- If __synchronised__ method is called to change `data` property (those are `GeoJsonSource.data(...)`, `GeoJsonSource.feature(value: Feature)`, `GeoJsonSource.featureCollection(value: FeatureCollection)`, `GeoJsonSource.geometry(value: Geometry)`) while worker thread is already busy parsing some previous data - handler thread queue should be cleared, current task should be canceled and new `data` should be applied immediately.
 - If __async__ method is called to change `data` property (those are  `GeoJsonSource.feature(value: Feature, { ... })`, `GeoJsonSource.featureCollection(value: FeatureCollection, { ... })`, `GeoJsonSource.geometry(value: Geometry, { ... })`) while worker thread is already busy parsing some previous data - handler thread queue should be cleared, current task should __not__ be canceled and new `data` parsing should be added to worker thread queue. This is done because mid updates in queue will make no sense as they will be overwritten by upcoming updates but current task being executed should be finished successfully in order to apply changes.
 - If using `data` function when preparing `GeoJsonSource` with `Builder` after calling function `GeoJsonSource.feature(value: Feature)`, `GeoJsonSource.featureCollection(value: FeatureCollection)` or `GeoJsonSource.geometry(value: Geometry)` in builder block - it will be applied correctly.

### DIFF
--- a/sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
@@ -4,8 +4,12 @@ import androidx.test.annotation.UiThreadTest
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
+import com.mapbox.geojson.Feature
+import com.mapbox.geojson.FeatureCollection
 import com.mapbox.geojson.Point
+import com.mapbox.maps.extension.style.layers.addLayer
 import com.mapbox.maps.extension.style.layers.generated.circleLayer
+import com.mapbox.maps.extension.style.sources.addSource
 import com.mapbox.maps.extension.style.sources.generated.geoJsonSource
 import com.mapbox.maps.extension.style.style
 import org.junit.After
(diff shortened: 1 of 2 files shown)


sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
```diff
@@ -88,6 +92,322 @@ class MapIntegrationTest {
...
+          val source = geoJsonSource("source") {
+            geometry(Point.fromLngLat(0.0, 0.0))
+          }
+          // sync method must clear async queue and not allow current running async task apply changes
+          source.geometry(Point.fromLngLat(0.1, 0.1))
+          // follow up with new async that must take effect
+          source.geometry(Point.fromLngLat(0.2, 0.2)) {}
+          style.addSource(source)
+          style.addLayer(layer)
+          mapView.postDelayed(
```
tobrun, line 333: this might result in flaky execution as it's not a guarantee that anything will be rendered, let's try it and if an issue revisit.

tobrun (approved):

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
kiryldz: It's not for rendering but for async parse to complete 😄 
I hope parsing one point should fit in 1 second otherwise we're in trouble!

Chaoba (approved):

pengdev:
Should the new geojson behavior be added to the documentation of corresponding functions? So that the user could expect the behaviour.

kiryldz:
> Should the new geojson behavior be added to the documentation of corresponding functions? So that the user could expect the behaviour.

I guess no as all those improvements seem to be logical. I can't say that it was expected for user that some background async parsing could silently overwrite data that was set synchronously.

pengdev (approved):
LGTM with one comment.

pengdev:
> > Should the new geojson behavior be added to the documentation of corresponding functions? So that the user could expect the behaviour.
> 
> I guess no as all those improvements seem to be logical. I can't say that it was expected for user that some background async parsing could silently overwrite data that was set synchronously.

I think these logic should appear in some document, either in the doc of the functions or in some guidelines.

kiryldz:
> > > Should the new geojson behavior be added to the documentation of corresponding functions? So that the user could expect the behaviour.
> > 
> > 
> > I guess no as all those improvements seem to be logical. I can't say that it was expected for user that some background async parsing could silently overwrite data that was set synchronously.
> 
> I think these logic should appear in some document, either in the doc of the functions or in some guidelines.

OK, I'll update docs for synchronized methods to make it clearer 👍

LukasPaczos:
This looks great :rocket: 

One small issue that I'm noticing is wrt the cancellation policy - let's imagine these 3 calls made one by one:
1. U1 async
2. U2 sync
3. U3 async

The cancellation policy is solely based on the `ignoreParsedGeoJson` flag, which means that `U2` will set it, but `U3` will immediately unset it. This will result in this order of updates being rendered:
1. U2
2. U1
3. U3

Since there was a sync update (`U2`) scheduled _after_ an async update (`U1`), that async update should in all cases be dismissed and never rendered, but in the current setup, I see this risk of it appearing anyway.

kiryldz:
@LukasPaczos thanks for catching, created follow-up ticket [here](https://github.com/mapbox/mapbox-maps-android/issues/381).

Issue #402, opened 2021-06-08
Proper handling of invalid RenderCacheOptions#size.


pengdev:
The `RenderCacheOptions#size` should be a unsigned int type, where the negative values are invalid.

This ticket tracks the work to properly handle the invalid RenderCacheOptions#size.

PR #425, opened 2021-06-14
Introduce size check for render cache

kiryldz:
Fixes: #402 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce size check for render cache.</changelog>`.

### Summary of changes

Introduce cache size check when setting render cache.

### DIFF
--- a/sdk/src/androidTest/java/com/mapbox/maps/RenderCacheTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/RenderCacheTest.kt
@@ -3,6 +3,7 @@ package com.mapbox.maps
 import androidx.test.annotation.UiThreadTest
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import org.junit.*
+import java.lang.RuntimeException
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
@@ -86,16 +87,12 @@ class RenderCacheTest {
     }
   }
 
-  @Test
+  @Test(expected = RuntimeException::class)
   fun setInvalidRenderCache() {
     rule.scenario.onActivity {
       it.runOnUiThread {
-        val customSize = -1
-        mapView.getMapboxMap().setRenderCacheOptions(RenderCacheOptions.Builder().size(customSize).build())
-        Assert.assertEquals(
-          RENDER_CACHE_DISABLED,
-          mapView.getMapboxMap().getRenderCacheOptions().size
-        )
+        val invalidSize = -1
+        mapView.getMapboxMap().setRenderCacheOptions(RenderCacheOptions.Builder().size(invalidSize).build())
       }
     }
   }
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1145,11 +1145,22 @@ class MapboxMap internal constructor(
...
    */
   @MapboxExperimental
   fun setRenderCacheOptions(options: RenderCacheOptions) {
-    nativeMapWeakRef.call { this.renderCacheOptions = options }
+    options.size?.let { size ->
+      nativeMapWeakRef.call {
+        this.renderCacheOptions = if (size >= 0)
+          options
+        else
+          RenderCacheOptions.Builder().setDisabled().build()
```
pengdev, line 1159: I think we should throw exceptions when the render cache is set to less than 0.

kiryldz: In that case should we also throw an exception when `size` is not specified explicitly when building an object? IMHO it does not matter much until we have it all documented. Having `size == null || size < 0` ~ `cache disabled` seems pretty consistent to me.

kiryldz: BTW I can still set cache to disabled if size < 0 but also log an error message @pengdev WDYT?

pengdev: IMO negative value is just wrong, user should never expect setting the size to a negative value should be permitted in any cases, I think we should throw illegal argument here.

kiryldz: You convinced me 👍

sdk/src/androidTest/java/com/mapbox/maps/RenderCacheTest.kt
```diff
@@ -87,14 +87,30 @@ class RenderCacheTest {
...
-  @Ignore("TODO Uncomment after https://github.com/mapbox/mapbox-maps-android/issues/402")
   fun setInvalidRenderCache() {
     rule.scenario.onActivity {
       it.runOnUiThread {
-        val customSize = -1
-        mapView.getMapboxMap().setRenderCacheOptions(RenderCacheOptions.Builder().size(customSize).build())
+        val invalidSize = -1
+        var caughtException = false
+        try {
+          mapView.getMapboxMap().setRenderCacheOptions(RenderCacheOptions.Builder().size(invalidSize).build())
```
pengdev, line 96: nit: we can use `@Test(expected = IllegalArgumentException.class)` to verify exceptions

pengdev (approved):
LGTM with nit

sdk/src/androidTest/java/com/mapbox/maps/RenderCacheTest.kt
kiryldz: Thanks, for some reason I thought it's not applicable for those tests 🤦

Issue #383, opened 2021-06-02
ConcurrentModificationException crash when properties changed shortly after annotation is created


e-hartig:
## Environment

- Android OS version: All
- Devices affected: All
- Maps SDK Version: 10.0.0-beta.12 - beta.20

## Observed behavior and steps to reproduce

If the properties of an annotation are changed quickly after an annotation is created, the applicaton crashes with a ConcurrentModificationException.

It is easy to reproduce this crash by adding this click listener to the CircleAnnotationActivity or elsewhere within the Mapbox Android sample app and then clicking the map a few times:
```
mapView?.getMapboxMap()?.addOnMapClickListener() { point ->
    val pointAnnotationPluginImpl = mapView.annotations.createPointAnnotationManager(mapView)
    // Create the annotation
    val annotation = pointAnnotationPluginImpl.create(PointAnnotationOptions().withDraggable(false)
        .withPoint(point)
        .withTextAnchor(TextAnchor.BOTTOM)
        .withTextColor(Color.BLUE)
        .withTextSize(12.0)
        .withTextField("Please don't crash!"))
    Thread.sleep(1)
    // Update its properties shortly after creation - often causes crash
    annotation.textField = "We're an enterprise customer!"
    annotation.textRotate = Math.random() * 180
    true
}
```

## Expected behavior

No crash.

## Notes / preliminary analysis

AnnotationManagerImpl.updateSource calls `geoJsonSource.featureCollection()` ([line 364](https://github.com/mapbox/mapbox-maps-android/blob/d092c5463172f6fe2033eb4c16925509d086f767/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt#L364)) which asynchronously iterates through the properties of the annotation feature.  If the set of properties on the annotation is updated by another thread during this async iteration, a ConcurrentModificationException is raised.

If `featureCollection` must be asynchronous, maybe it should act on an immutable copy of the feature's properties.

## Additional links and references

Stack trace:
```
2021-06-02 16:36:07.986 15470-15532/com.example.mapboxsample E/AndroidRuntime: FATAL EXCEPTION: STYLE_WORKER
    Process: com.example.mapboxsample, PID: 15470
    java.util.ConcurrentModificationException
        at com.google.gson.internal.LinkedTreeMap$LinkedTreeMapIterator.nextNode(LinkedTreeMap.java:544)
[...]

PR #427, opened 2021-06-16
Changes the visibility of  jsonObject in annotation to protected, fix ConcurrentModificationException

Chaoba:
Fixes: #383 
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Changes the visibility of  jsonObject in annotation to protected, fix ConcurrentModificationException</changelog>`.

### Summary of changes
In the past, users can add or remove properties of  the `jsonObject` of annotation directly, which makes it not safe and may cause potential bugs. This pr changes the visibility of  `jsonObject` in annotation to protected so that users can only add or remove properties by the public API we provide.
If users update the content of `jsonObject` while we are in the process of `updatingSource`, it may cause `ConcurrentModificationException`, to fix this issue, this pr add a method `getJsonObjectCopy` which will return a deep copy of `jsonObject`


app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationTest.kt
```diff
@@ -78,11 +78,15 @@ class UpdateAnnotationTest : BaseMapTest(), OnMapLoadedListener {
               .withTextTransform(TextTransform.UPPERCASE)
               .withTextSize(15.0)
               .withTextRadialOffset(1.0)
-              .withTextOpacity(0.8)
               .withTextOffset(listOf(1.0, 2.0))
               .withTextMaxWidth(10.0)
               .withPoint(Point.fromLngLat(0.0, 0.0))
           )
+          for (i in 0..100) {
```
kiryldz, line 85: Does it make sense to move it to separate test?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -356,7 +356,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
           }
         }
       val features = annotations.map {
-        val annotation = Feature.fromGeometry(it.geometry, it.jsonObject)
+        val annotation = Feature.fromGeometry(it.geometry, it.getJsonObjectCopy())
```
kiryldz, line 359: If we make `jsonObject` internal below - can we use the object directly here? or is it also a potential bug?

kiryldz (approved):
LGTM with couple questions

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: No, the purpose of adding a new method `getJsonObjectCopy` here is to avoid `ConcurrentModificationException`. While we updating the source, users can add or remove new properties.

pengdev (approved):
LGTM

app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationTest.kt
Chaoba: I would prefer to add it here as this case is used to verify annotation update and in this loop, we are updating this annotation.
Another reason is that with fewer test cases we could have a faster test run 😄

Issue #164, opened 2021-03-08
CameraAnimationsPluginImpl deadlock


LukasPaczos:
Maps SDK: `10.0.0-beta.13`.

I can't reproduce the issue but I run into a situation where an animator was added to `runningAnimatorsQueue` and was never cleared. This led to the camera's complete deadlock because the plugin animates on the first animator's ticks:
https://github.com/mapbox/mapbox-maps-android/blob/ab7290356e290dd3de717ded1cfba7873fc1879a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt#L317-L348

and since the first animator wasn't running (I was able to plug in a debugger and check, `firstAnimator.isRunning` and `firstAnimator.isStarted` were returning `false` as well), we had no camera updates being dispatched.

From the user's perspective, there was no other way to recover from this deadlock besides restarting the activity (recreating the map).

/cc @mapbox/maps-android

knov:
@kiryldz can you follow up with @LukasPaczos  to determine if this is still an issue and what the  priority is?

kiryldz:
Sadly could not reproduce. Was trying to simulate bunch of `flyTo` methods running one after another, constantly cancelling each other etc.
Would be glad if some determined scenario could be advised to reproduce this issue, looking at code it seems that everything should be fine.

LukasPaczos:
@kiryldz I'd recommend trying this out with gestures. For example, schedule transitions (post them with a handler) to trigger while you're executing a drag gesture. Looks like a mix of transitions coming from gestures and programmatically scheduled were contributing to the issue, I was able to occasionally reproduce like that. This might not be related in the end, but might be easier to repro.

kiryldz:
@LukasPaczos I was trying such code (changing numbers, with gesture / without) but could not still reproduce 😢 
```
for (i in 0..100) {
      mapView.postDelayed(
        {
          mapView.camera.flyTo(
            CameraOptions.Builder()
              .center(Point.fromLngLat(LONGITUDE + (i + 1) * 0.1, LATITUDE + (i + 1) * 0.1))
              .bearing(10.0 * i)
              .zoom(10.0 + (i / 2.0))
              .build()
          )
        },
        50L * i
      )
    }
```

LukasPaczos:
[...]

PR #439, opened 2021-06-18
Fix camera deadlock use-case

kiryldz:
Fixes: #164 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix camera deadlock use-case.</changelog>`.

### Summary of changes

When canceling animators manually inside `CameraAnimationsLifecycleListener.onAnimatorStarting` (typical Nav SDK use-case) before starting high-level animation (e.g. `easeTo`) - callbacks for high-level animation listener would be called incorrectly while camera itself would be completely stuck. That happens in concrete scenario when animator with delay was not started before canceling it.

This PR fixes it.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -231,19 +231,21 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
       override fun onAnimationStart(animation: Animator?) {
         (animation as? CameraAnimator<*>)?.apply {
-
-          if (runningAnimatorsQueue.isEmpty()) {
-            highLevelListener?.onAnimationStart(animation)
+          // check for a specific use-case when canceling an animation with start delay that
+          // has not yet started - in that case onAnimationStart logic must be skipped
+          if (canceled) {
+            return
           }
-
           lifecycleListeners.forEach {
             it.onAnimatorStarting(type, this, owner)
           }
+          if (runningAnimatorsQueue.isEmpty()) {
+            highLevelAnimatorSet?.started = true
+            highLevelListener?.onAnimationStart(animation)
+          }
           mapTransformDelegate.setUserAnimationInProgress(true)
-
           // check if such animation is not running already
           // if it is - then cancel it
-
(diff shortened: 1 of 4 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/HighLevelAnimatorSet.kt
```diff
@@ -7,6 +7,8 @@ internal data class HighLevelAnimatorSet(
   val animatorSet: AnimatorSet
 ) : Cancelable {
 
+  var started = false
+
   override fun cancel() {
     animatorSet.cancel()
```
Chaoba, line 13: Should we reset `started` to false here?

kiryldz: It's a higher level `cancel` function that may be called by user (or may be not) so we need to set this flag in the place where we set it now to make it all work properly.

Chaoba (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -231,19 +231,21 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
       override fun onAnimationStart(animation: Animator?) {
```
LukasPaczos, line 232: What if we checked if the `animation` that's being started here is part of the high-level animator instead of adding the `started` flag? Instead of mixing low-level and high-level animators together, we could have 2 logical branches for them.

Not a blocker, but might be something to consider for a refactor.

kiryldz: Nice suggestion, indeed that logic is becoming more and more complicated. It would be nice to differentiate those animators in some better way. I'll think about it and create some follow-up ticket shortly.

kiryldz: Taken another look - could not see an easy way right now to differentiate those.. Basically it does not matter much for camera animation engine if animation started is part of high-level animation or stand alone low-level animation. Guess current approach is fine for now, I will put this thing on my radar and perhaps try to prototype some easier way and then wrap it to a concrete ticket.

Issue #433, opened 2021-06-17
Annotation updates are being dropped


cramptonism:
## Environment

- Android OS version: API 25 through API 31 (note: I tried to test on API 23 and I got a SIGABRT with some permissions issue on the sample app)
- Devices affected: Tested Pixel 5 and Android Emulator (Pixel 4 API 25)
- Maps SDK Version: 33f4643e65db1a655b9908755b62b94e98e811e8 of sample app, beta21 on our app. This _does not_ repro on beta20.

## Observed behavior and steps to reproduce

"Linked" annotations updated during a very fast drag do not have all updates applied. I say linked because I have widgets in my app that are built from multiple annotations, and dragging one will affect the others.

## Expected behavior

All calls to `update` are respected and applied. Final state of the 3 annotations in my example should be the same point.

## Notes / preliminary analysis

I've verified all of the expected `.update(` calls are executed, but the annotations do not reflect the final state. In the above example the larger circle and the line are updated in `onAnnotationDrag` and end up in 3 different places. I will try to dig more into the SDK to verify the update calls make it to native but beyond that I can't do much more digging.

## Additional links and references

https://user-images.githubusercontent.com/13057523/122317999-42daf800-ced3-11eb-9bbc-36c6fbbc0e4a.mp4

Example CircleAnnotationActivity.kt code, applied to 33f4643e65db1a655b9908755b62b94e98e811e8
```
package com.mapbox.maps.testapp.examples.annotation

import android.graphics.Color
import android.os.Bundle
import androidx.appcompat.app.AppCompatActivity
import com.mapbox.geojson.Point
import com.mapbox.maps.plugin.annotation.Annotation
import com.mapbox.maps.plugin.annotation.AnnotationPlugin
import com.mapbox.maps.plugin.annotation.annotations
import com.mapbox.maps.plugin.annotation.generated.*
import com.mapbox.maps.testapp.R
import kotlinx.android.synthetic.main.activity_add_marker_symbol.mapView
import kotlinx.android.synthetic.main.activity_annotation.*
import java.util.*

/**
 * Example showing how to add Circle annotations
 */
class CircleAnnotationActivity : AppCompatActivity() {
  private val pairMap = mutableMapOf<CircleAnnotation, CircleAnnotation>()
  private val polylineMap = mutableMapOf<CircleAnnotation, PolylineAnnotation>()
  private val random = Random()
  private var circleManager: CircleAnnotationManager? = null
[...]

PR #441, opened 2021-06-21
Fix annotation update missing issue.

Chaoba:
Fixes: #433 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix annotation update missing issue</changelog>`.

### Summary of changes
Some annotation updates could be ignored as we will remove the pending GeojsonSource update in the queue.
This pr remove this logic to guarantee all annotation update could be executed.

### DIFF
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationWithMultiManagersTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationWithMultiManagersTest.kt
@@ -1,0 +1,155 @@
+package com.mapbox.maps.testapp.annotation
+
+import android.graphics.Color
+import android.os.Handler
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.LargeTest
+import com.google.gson.JsonPrimitive
+import com.mapbox.geojson.Point
+import com.mapbox.maps.Style
+import com.mapbox.maps.extension.style.utils.ColorUtils
+import com.mapbox.maps.plugin.annotation.annotations
+import com.mapbox.maps.plugin.annotation.generated.*
+import com.mapbox.maps.testapp.BaseMapTest
+import junit.framework.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+/**
+ * Instrumented test for verifying annotation updating while changing style
+ */
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class UpdateAnnotationWithMultiManagersTest : BaseMapTest() {
+  private val latch = CountDownLatch(4)
+  private lateinit var pointAnnotationManager: PointAnnotationManager
+  private lateinit var circleAnnotationManager: CircleAnnotationManager
+  private lateinit var polygonAnnotationManager: PolygonAnnotationManager
+  private lateinit var polylineAnnotationManager: PolylineAnnotationManager
+  private lateinit var pointAnnotation: PointAnnotation
+  private lateinit var circleAnnotation: CircleAnnotation
+  private lateinit var polylineAnnotation: PolylineAnnotation
+  private lateinit var polygonAnnotation: PolygonAnnotation
+  private lateinit var handler: Handler
+  private val updateTimes = 100
+  private val updateSteps = 0.1
+  private val delta = 0.0001
+  private val points = mutableListOf(
+    mutableListOf(
+      Point.fromLngLat(-3.363937, -10.733102),
+      Point.fromLngLat(1.754703, -19.716317),
+      Point.fromLngLat(-15.747196, -21.085074),
+      Point.fromLngLat(-3.363937, -10.733102)
+    )
+  )
+  private val polylinePoints = mutableListOf(
+    Point.fromLngLat(-4.375974, -2.178992),
+    Point.fromLngLat(-7.639772, -4.107888),
+    Point.fromLngLat(-11.439207, 2.798737),
+  )
+
+  @Test
+  fun testUpdateAnnotation() {
+    rule.scenario.onActivity {
+      handler = Handler(it.mainLooper)
+      it.runOnUiThread {
+        mapboxMap.loadStyleUri(Style.MAPBOX_STREETS) {
+          pointAnnotationManager = mapView.annotations.createPointAnnotationManager(mapView)
+          pointAnnotationManager.textFont = listOf("Open Sans Regular")
+          pointAnnotation = pointAnnotationManager.create(
+            PointAnnotationOptions()
+              .withIconColor(ColorUtils.colorToRgbaString(Color.RED))
+              .withIconImage("car-15")
+              .withPoint(Point.fromLngLat(0.0, 0.0))
+          )
+          circleAnnotationManager = mapView.annotations.createCircleAnnotationManager(mapView)
+          circleAnnotation = circleAnnotationManager.create(
+            CircleAnnotationOptions()
+              .withCircleColor(Color.RED)
+              .withCircleRadius(5.0)
+              .withPoint(Point.fromLngLat(0.0, 0.0))
+          )
+
+          polygonAnnotationManager = mapView.annotations.createPolygonAnnotationManager(mapView)
+          polygonAnnotation = polygonAnnotationManager.create(
+            PolygonAnnotationOptions()
+              .withPoints(points)
+              .withData(JsonPrimitive("Foobar"))
+              .withFillColor(Color.RED)
+          )
+
+          polylineAnnotationManager = mapView.annotations.createPolylineAnnotationManager(mapView)
+          polylineAnnotation = polylineAnnotationManager.create(
+            PolylineAnnotationOptions()
+              .withPoints(polylinePoints)
+              .withLineColor(Color.RED)
+              .withLineWidth(5.0)
+          )
+
+          Thread {
+            for (i in 1..updateTimes) {
+              pointAnnotation.point = Point.fromLngLat(
+                pointAnnotation.point.longitude() + updateSteps,
+                pointAnnotation.point.latitude() + updateSteps
+              )
+              pointAnnotationManager.update(pointAnnotation)
+              Thread.sleep(10)
+            }
+            latch.countDown()
+          }.start()
+
+          Thread {
+            for (i in 1..updateTimes) {
+              circleAnnotation.point = Point.fromLngLat(
+                circleAnnotation.point.longitude() + updateSteps,
+                circleAnnotation.point.latitude() + updateSteps
+              )
+              circleAnnotationManager.update(circleAnnotation)
+              Thread.sleep(10)
+            }
+            latch.countDown()
+          }.start()
+
+          Thread {
+            for (i in 1..updateTimes) {
+              points[0][0] = Point.fromLngLat(
+                points[0][0].longitude() + updateSteps,
+                points[0][0].latitude() + updateSteps
+              )
+              polygonAnnotation.points = points
+              polygonAnnotationManager.update(polygonAnnotation)
+              Thread.sleep(10)
+            }
+            latch.countDown()
+          }.start()
+          Thread {
+            for (i in 1..updateTimes) {
+              polylinePoints[0] = Point.fromLngLat(
+                polylinePoints[0].longitude() + updateSteps,
+                polylinePoints[0].latitude() + updateSteps
+              )
+              polylineAnnotation.points = polylinePoints
+              polylineAnnotationManager.update(polylineAnnotation)
+              Thread.sleep(10)
+            }
+            latch.countDown()
+          }.start()
+        }
+      }
+    }
+    if (!latch.await(3000, TimeUnit.MILLISECONDS)) {
+      throw TimeoutException()
+    }
+    assertEquals(pointAnnotation.point.latitude(), updateSteps * updateTimes, delta)
+    assertEquals(pointAnnotation.point.longitude(), updateSteps * updateTimes, delta)
+    assertEquals(circleAnnotation.point.longitude(), updateSteps * updateTimes, delta)
+    assertEquals(circleAnnotation.point.latitude(), updateSteps * updateTimes, delta)
+    assertEquals(polylineAnnotation.points[0].longitude(), polylinePoints[0].longitude(), delta)
+    assertEquals(polylineAnnotation.points[0].latitude(), polylinePoints[0].latitude(), delta)
+    assertEquals(polygonAnnotation.points[0][0].longitude(), points[0][0].longitude(), delta)
+    assertEquals(polygonAnnotation.points[0][0].latitude(), points[0][0].latitude(), delta)
+  }
+}
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
@@ -18,6 +18,7 @@ import com.mapbox.maps.extension.style.expressions.generated.Expression.Companio
 import com.mapbox.maps.extension.style.expressions.generated.Expression.Companion.toNumber
 import com.mapbox.maps.extension.style.layers.properties.generated.IconAnchor
 import com.mapbox.maps.extension.style.layers.properties.generated.TextAnchor
+import com.mapbox.maps.plugin.annotation.Annotation
 import com.mapbox.maps.plugin.annotation.AnnotationPlugin
 import com.mapbox.maps.plugin.annotation.annotations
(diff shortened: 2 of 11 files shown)


kiryldz (changes requested):
@Chaoba I don't think it's a good idea although I understand how this solves a problem.

Root cause of underlying issue is posting updates very frequently. It basically does not make any sense as we can't render let's say 100 times each 1 ms - we depend on vsync updates and render each 16 ms (on most screens).

If we remove the code here like you did - we will have issues in parsing large geojson files. E.g. user has set to parse enormous geojson_1 and while parsing is happening also queued 100 more. There is no sense to parse 99 of them because they will be basically overwritten by the most recent one and we do not overwrite - it will result in huge delay of showing this geojson data on map.

My proposal here for you case would be introducing handler inside annotation plugin and queue updates there.

Chaoba:
> @Chaoba I don't think it's a good idea although I understand how this solves a problem.
> 
> Root cause of underlying issue is posting updates very frequently. It basically does not make any sense as we can't render let's say 100 times each 1 ms - we depend on vsync updates and render each 16 ms (on most screens).
> 
> If we remove the code here like you did - we will have issues in parsing large geojson files. E.g. user has set to parse enormous geojson_1 and while parsing is happening also queued 100 more. There is no sense to parse 99 of them because they will be basically overwritten by the most recent one and if we do not overwrite - it will result in huge delay of showing this geojson data on map.
> 
> My proposal here for you case would be introducing handler inside annotation plugin and queue updates there.

The root cause for this issue is that there are too many updates and the most recent update cancels the former update. Simply introducing a handler in the annotation plugin will not resolve this issue, because all the updates in this handler will return immediately as the parse works are handled in the handler of GeojsonSource. 
For example, we have handler A in annotation plugin and handler B in GeojsonSource and we have 2 update jobs which will cost some time.
Handler A will process these 2 jobs immediately to push the update to Handler B and in B the first one will be canceled by the second one, the issue still exists.

kiryldz:
> > @Chaoba I don't think it's a good idea although I understand how this solves a problem.
> > Root cause of underlying issue is posting updates very frequently. It basically does not make any sense as we can't render let's say 100 times each 1 ms - we depend on vsync updates and render each 16 ms (on most screens).
> > If we remove the code here like you did - we will have issues in parsing large geojson files. E.g. user has set to parse enormous geojson_1 and while parsing is happening also queued 100 more. There is no sense to parse 99 of them because they will be basically overwritten by the most recent one and if we do not overwrite - it will result in huge delay of showing this geojson data on map.
> > My proposal here for you case would be introducing handler inside annotation plugin and queue updates there.
> 
> The root cause for this issue is that there are too many updates and the most recent update cancels the former update. Simply introducing a handler in the annotation plugin will not resolve this issue, because all the updates in this handler will return immediately as the parse works are handled in the handler of GeojsonSource.
> For example, we have handler A in annotation plugin and handler B in GeojsonSource and we have 2 update jobs which will cost some time.
> Handler A will process these 2 jobs immediately to push the update to Handler B and in B the first one will be canceled by the second one, the issue still exists.

@Chaoba again, I fully understand your fix. All what I'm saying is that changing that logic in geojson in not correct. Yes, it will fix annotations but this approach is not correct as we apply map updates asap and not queue them up.
I'm looking at code here:
https://github.com/mapbox/mapbox-maps-android/blob/d95bef4da2bb7c952a0df763b9e7b60cdabfb08b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt#L365
Why can't we add that logic here? Not updating next feature until previous one is parsed.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -658,5 +674,15 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     private const val POINT_COUNT = "point_count"
     private const val CLUSTER_TEXT_LAYER_ID = "mapbox-android-cluster-text-layer"
     private val DEFAULT_TEXT_FIELD = get("point_count")
+
+    private val workerThread by lazy {
+      HandlerThread("ANNOTATION_WORKER").apply {
+        priority = Thread.MAX_PRIORITY
+        start()
```
kiryldz, line 681: If we start a thread - we also need to stop it I guess.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -338,31 +343,42 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+            it.setUsedDataDrivenProperties()
+            annotation
           }
+
+          geoJsonSource.featureCollection(
+            FeatureCollection.fromFeatures(features),
+            onDataParsed = { lock.withLock { waitCondition.signal() } }
+          )
+          // wait until the former update is completed.
+          waitCondition.await()
```
kiryldz, line 379: Although it should work more or less OK now (apply all the updates) - I bet we can hit another problem: if interacting with map and annotations for pretty long time user might see increasing delay between actually sending annotation data to the map and seeing it on the map.
Could you try to simulate such scenario? IDK perhaps drag some annotation over the map for long period of time. If it's a problem perhaps we might want to add some API to `update` function whether to apply __all__ changes (but with possible delay) or apply latest changes only.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -338,31 +343,42 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
-          symbol.iconImage?.let { image ->
-            if (image.startsWith(PointAnnotation.ICON_DEFAULT_NAME_PREFIX)) {
-              // User set the bitmap icon, add the icon to style
-              symbol.iconImageBitmap?.let { bitmap ->
-                if (style.getStyleImage(image) == null) {
-                  val imagePlugin = image(image) {
-                    bitmap(bitmap)
+      workerHandler.post {
+        val lock = ReentrantLock()
+        val waitCondition = lock.newCondition()
```
kiryldz, line 348: I'd move it to `private val` in order not to create them each time.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: If there are too many annotations(>2000), there would be a delay, but I'm not seeing an increasing delay, this is due to the bad performance for annotation plugin for large-scale annotations.

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
```diff
@@ -118,7 +141,7 @@ class PointAnnotationActivity : AppCompatActivity() {
         }
         // random add symbols across the globe
         val pointAnnotationOptionsList: MutableList<PointAnnotationOptions> = ArrayList()
-        for (i in 0..20) {
+        for (i in 0..2000) {
```
kiryldz, line 144: is it intended to leave this large number here? or were you just testing? 😄 
perhaps better having an integration test appying 2000 updates and verifying that __all__ calls where executed (and not were skipped).

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: ok, thanks for confirming. 
if needed - we could extend this API later.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -658,5 +681,18 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+
+    private var workerThread: HandlerThread? = null
+    private var workerHandler: Handler? = null
+
+    /**
+     * Clean and stop the workerThread for updating source
+     */
+    fun cleanupWorkerThread() {
+      workerHandler?.removeCallbacksAndMessages(null)
+      workerThread?.quit()
```
kiryldz, line 693: Should be followed up with `workerThread.join` I believe like it's done here
https://github.com/mapbox/mapbox-maps-android/blob/1e3215006be349a19d1ea5c07fd728e55f71a103/sdk/src/main/java/com/mapbox/maps/renderer/WorkerHandlerThread.kt#L29-L37

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPlugin.kt
```diff
@@ -3,12 +3,11 @@ package com.mapbox.maps.plugin.annotation
 import android.view.View
 import com.mapbox.maps.plugin.MapPlugin
 import com.mapbox.maps.plugin.MapSizePlugin
-import com.mapbox.maps.plugin.MapStyleObserverPlugin
 
 /**
  * Plugin interface for the annotation.
  */
-interface AnnotationPlugin : MapPlugin, MapSizePlugin, MapStyleObserverPlugin {
+interface AnnotationPlugin : MapPlugin, MapSizePlugin {
```
kiryldz, line 10: If I remember correctly although we do not use `MapStyleObserverPlugin` - we still decided to leave it here but do not do any actions for now. Is this correct, @pengdev?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPluginImpl.kt
```diff
@@ -74,6 +76,9 @@ class AnnotationPluginImpl : AnnotationPlugin {
         }
       }
     }
+    if (managerList.isEmpty()) {
```
kiryldz, line 79: Perhaps cover with unit test? And also for `cleanup` below

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPlugin.kt
pengdev: yes, until we find a valid use case for it 😄

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -120,6 +127,13 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     gesturesPlugin.addOnMapClickListener(mapClickResolver)
     gesturesPlugin.addOnMapLongClickListener(mapLongClickResolver)
     gesturesPlugin.addOnMoveListener(mapMoveResolver)
+    if (workerThread == null) {
+      workerThread = HandlerThread("ANNOTATION_WORKER").apply {
+        priority = Thread.MAX_PRIORITY
+        start()
+        workerHandler = Handler(looper)
+      }
+    }
```
pengdev, line 136: nit: could workThread be lazily initialised?

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPlugin.kt
Chaoba: But AnnotationPlugin doesn't need MapStyleObserverPlugin anymore. We can add it back if needed in the future.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: No, we will stop the thread in cleanup and need to recreate it when opening a map view again.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPlugin.kt
kiryldz: It was discussed by me and @pengdev. If we remove it from location component plugin and annotations plugin - it means that `MapStyleObserverPlugin` is redundant and could be fully removed from codebase. However it's a breaking change then and potentially this observer interface may be useful in future. That's why it was decided to let location component and annotation plugins implement it but atm do not perform any particular actions on style change callback.

Chaoba: I see, will roll back this change.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -31,6 +32,11 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
   private val onGeoJsonParsedListenerList = mutableListOf<OnGeoJsonParsed>()
   private var ignoreParsedGeoJson = false
 
+  private val workerHandler by lazy {
+    Handler(workerThread.looper)
+  }
+  private val mainHandler = Handler(Looper.getMainLooper())
```
kiryldz, line 38: btw I guess it could be static, no need to create each time, wdyt?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -974,16 +980,11 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-    private val workerThread by lazy {
-      HandlerThread("STYLE_WORKER").apply {
-        priority = Thread.MAX_PRIORITY
-        start()
-      }
-    }
-    private val workerHandler by lazy {
-      Handler(workerThread.looper)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    var workerThread = HandlerThread("STYLE_WORKER").apply {
```
kiryldz, line 984: Could we revert lazy initialization? That thread in some situations may not even be started (if user does not use geojson source). I'm OK to leave it internal for tests.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: Yes, this one could be static.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: This is a compromise for tests. With lazy initialization, only the first case will succeed, in the following one, this thread will be in the inactive state.

app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationWithMultiManagersTest.kt
```diff
@@ -0,0 +1,154 @@
...
+            }
+            latch.countDown()
+          }.start()
+        }
+      }
+    }
+    if (!latch.await(3000, TimeUnit.MILLISECONDS)) {
+      throw TimeoutException()
+    }
+    assertEquals(pointAnnotation.point.latitude(), updateSteps * updateTimes, 0.1)
```
kiryldz, line 145: 0.1 should be moved to constants and also I guess 0.1 is too big. Let's use smth like 0.0001

app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationWithMultiManagersTest.kt
```diff
@@ -0,0 +1,154 @@
...
+          )
+
+          Thread {
+            for (i in 0..updateTimes) {
+              pointAnnotation.point = Point.fromLngLat(
+                pointAnnotation.point.longitude() + updateSteps,
+                pointAnnotation.point.latitude() + updateSteps
+              )
+              pointAnnotationManager.update(pointAnnotation)
+              Thread.sleep(10)
```
kiryldz, line 98: Why is sleep needed?

kiryldz (approved):
LGTM with couple small questions.

app/src/androidTest/java/com/mapbox/maps/testapp/annotation/UpdateAnnotationWithMultiManagersTest.kt
Chaoba: To simulate the annotation updating of the drag gesture.

PR #444, opened 2021-06-22
Coalesce to the default language if not supported

Chaoba:

## Pull request checklist:

### Summary of changes
We replaced all the `get` expression with selected language and for those with expression `coalesce` and only have a default language it will not show the default language.
This pr will make the replacement not applied on such expression like `get("name")` to show the default language. 
However, users might pass in a not supported language, then we will output a warning logcat.

### DIFF
--- a/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
+++ b/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
@@ -78,5 +78,5 @@ private const val LAYER_TYPE_SYMBOL = "symbol"
 private const val STREET_V7 = "mapbox.mapbox-streets-v7"
 private const val STREET_V8 = "mapbox.mapbox-streets-v8"
 private val SUPPORTED_SOURCES = listOf(STREET_V7, STREET_V8)
-private val EXPRESSION_REGEX = Regex("\\[\"get\",\\s*\"(name|name_.{2,7})\"\\]")
+private val EXPRESSION_REGEX = Regex("\\[\"get\",\\s*\"(name_.{2,7})\"\\]")
 private val EXPRESSION_ABBR_REGEX = Regex("\\[\"get\",\\s*\"abbr\"\\]")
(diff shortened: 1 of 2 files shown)


neelmistry94 (approved):

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -78,5 +78,5 @@ private const val LAYER_TYPE_SYMBOL = "symbol"
 private const val STREET_V7 = "mapbox.mapbox-streets-v7"
 private const val STREET_V8 = "mapbox.mapbox-streets-v8"
 private val SUPPORTED_SOURCES = listOf(STREET_V7, STREET_V8)
-private val EXPRESSION_REGEX = Regex("\\[\"get\",\\s*\"(name|name_.{2,7})\"\\]")
+private val EXPRESSION_REGEX = Regex("\\[\"get\",\\s*\"(name_.{2,7})\"\\]")
```
kiryldz, line 81: do we have a test for this case?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/SupportedLanguages.kt
```diff
@@ -136,5 +143,9 @@ internal fun getLanguageNameV8(locale: Locale): String {
     return if (locale == Locale.TAIWAN || locale.script == "Hant") NAME_ZH_HANT
     else NAME_ZH_HANS
   }
-  "name_${locale.language}".apply { return if (supportedV8.contains(this)) this else NAME }
+  return "name_${locale.language}".apply {
+    if (!supportedV8.contains(this)) {
+      Logger.w(TAG, "Language ${locale.displayLanguage} is not supported in the current style")
```
pengdev, line 148: would it be enough to log a warning message here?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
Chaoba: No, we don't have method to verify it in test case now

extension-localization/src/main/java/com/mapbox/maps/extension/localization/SupportedLanguages.kt
Chaoba: Any suggestions for this?

Chaoba: From my perspective, a warning message is enough because users can then switch to other languages.

kiryldz (approved):
LGTM as this is approved and aligned with iOS.
We can follow-up later imho with some questions here.

tobrun (approved):

PR #455, opened 2021-06-28
Perform annotation click synchronously

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Perform annotation click synchronously; change AnnotationManagerImpl#queryMapForFeatures function to synchronous</changelog>`.

### Summary of changes
By changing AnnotationManagerImpl#queryMapForFeatures function to synchronous, we are able to handle the return value of MapClick in AnnotationManager correctly

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
@@ -21,6 +21,7 @@ import com.mapbox.maps.extension.style.layers.properties.generated.TextAnchor
 import com.mapbox.maps.plugin.annotation.AnnotationPlugin
 import com.mapbox.maps.plugin.annotation.annotations
 import com.mapbox.maps.plugin.annotation.generated.*
+import com.mapbox.maps.plugin.gestures.addOnMapClickListener
 import com.mapbox.maps.testapp.R
 import com.mapbox.maps.testapp.utils.BitmapUtils.bitmapFromDrawableRes
 import kotlinx.android.synthetic.main.activity_add_marker_symbol.*
@@ -36,6 +37,7 @@ class PointAnnotationActivity : AppCompatActivity() {
   private var pointAnnotation: PointAnnotation? = null
   private val animators: MutableList<ValueAnimator> = mutableListOf()
   private var index: Int = 0
+  private var consumeClickEvent = false
   private val nextStyle: String
     get() {
       return AnnotationUtils.STYLES[index++ % AnnotationUtils.STYLES.size]
@@ -52,17 +54,33 @@ class PointAnnotationActivity : AppCompatActivity() {
 
         addClickListener(
           OnPointAnnotationClickListener {
-            Toast.makeText(this@PointAnnotationActivity, "Click: ${it.id}", Toast.LENGTH_SHORT)
+            Toast.makeText(this@PointAnnotationActivity, "Click1: ${it.id}", Toast.LENGTH_SHORT)
               .show()
-            false
+            consumeClickEvent
+          }
+        )
+
+        addClickListener(
+          OnPointAnnotationClickListener {
+            Toast.makeText(this@PointAnnotationActivity, "Click2: ${it.id}", Toast.LENGTH_SHORT)
+              .show()
+            consumeClickEvent
           }
         )
 
         addLongClickListener(
           OnPointAnnotationLongClickListener {
-            Toast.makeText(this@PointAnnotationActivity, "LongClick: ${it.id}", Toast.LENGTH_SHORT)
+            Toast.makeText(this@PointAnnotationActivity, "LongClick1: ${it.id}", Toast.LENGTH_SHORT)
               .show()
-            false
+            consumeClickEvent
+          }
+        )
+
+        addLongClickListener(
+          OnPointAnnotationLongClickListener {
+            Toast.makeText(this@PointAnnotationActivity, "LongClick2: ${it.id}", Toast.LENGTH_SHORT)
+              .show()
+            consumeClickEvent
           }
         )
 
@@ -135,6 +153,11 @@ class PointAnnotationActivity : AppCompatActivity() {
           create(FeatureCollection.fromJson(it))
         }
       }
+
+      mapView.getMapboxMap().addOnMapClickListener {
+        Toast.makeText(this@PointAnnotationActivity, "OnMapClick", Toast.LENGTH_SHORT).show()
+        true
+      }
     }
 
     deleteAll.setOnClickListener {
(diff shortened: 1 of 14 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -454,18 +455,24 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+      mapFeatureQueryDelegate.executeOnRenderThread {
+        queryMapForFeatures(point) {
+          selectAnnotation(it)
+          clickListeners.forEach { listener ->
+            listener.onAnnotationClick(it)
+          }
+          latch.countDown()
         }
       }
+      latch.await()
```
pengdev, line 475: do we actually want to set some timeout? Blocking main thread for a long time doesn't sound right

kiryldz: Sounds like a nice action, I was also thinking about it!
@Chaoba could you please add timeout here and throw an exception if timeout is hit.
IMHO 3 seconds is enough to say for sure that some issue with QRF has happened

kiryldz: Or 3 is too much, perhaps even 1 sec.. In general we query for 1 feature only, it should be almost immediate
cc @mapbox/maps-android for more thoughts here

Chaoba: I have already set it to 2 sec when polishing codes locally. In fact, the query will return all the features at this point, but we will only use the first one. Not very sure whether 1 sec is enough when there are too many features, so 2 should be enough.

kiryldz: It still does not sound very reliable.
I guess we should use let's say default of 3 sec and add an API to set this limit for the user as well.
I could imagine (in very-very theory) someone with 1 million annotations on the map and being OK to freeze UI for 10 seconds

pengdev: thinking it again, I think blocking the main thread for an unknown time is not really good, blocking the main thread for 5 seconds will result in an ANR. I think we should rethink the approach here.

kiryldz: Approach could not be changed sadly. Clicking on a map is UI Android operation, UI is single-threaded in Android and in this particular case click should be returned only after QRF is finished. So, sadly, there is no solution if we talk about threading. This approach works pretty fine in real conditions, it may result in significant lags if annotation count is enormous but that also could be optimized on another level.

We could re-think this approach with coroutines but assuming we do not use them at all this solution seems like a good start to me

pengdev: could we at least try to add some filter or limit the queried layer to the QRF so that it wouldn't return large amount of features?

Chaoba: We have the filter as a param while querying https://github.com/mapbox/mapbox-maps-android/pull/455/files/3f15b4e2c37c0c953af9193c60182baf0f66ab3b#diff-4439957e75cc10a9ecf0041db71889e34906978146ed8b09519c8e0cc7405a1fR641

Chaoba: Made a test and it would cost about 200ms for 2000 annotations in the same point and  800ms for 20000 annotations. So 1s should cover most of the cases and I set 2s to make it more robust.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -612,39 +620,44 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
    * Query the rendered annotation around the point
    *
    * @param point the point for querying
-   * @param callback query callback
+   * @return the queried annotation on this ppint
```
kiryldz, line 623:
```suggestion
   * @return the queried annotation at this point
```

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -612,39 +620,44 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+            if (queriedFeatureList.isNotEmpty()) {
+              val id = queriedFeatureList.first().feature.getProperty(getAnnotationIdKey()).asLong
+              annotation = annotationMap[id]
+            }
           }
+          latch.countDown()
         }
       }
+      // At most wait 2 seconds to prevent ANR
+      latch.await(2, TimeUnit.SECONDS)
```
kiryldz, line 658: Let's move to constants

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/PointAnnotationActivity.kt
```diff
@@ -52,17 +54,33 @@ class PointAnnotationActivity : AppCompatActivity() {
 
         addClickListener(
           OnPointAnnotationClickListener {
-            Toast.makeText(this@PointAnnotationActivity, "Click: ${it.id}", Toast.LENGTH_SHORT)
+            Toast.makeText(this@PointAnnotationActivity, "Click1: ${it.id}", Toast.LENGTH_SHORT)
```
kiryldz, line 57: Perhaps just `Click` and not `Click1`?

kiryldz: Oh, I see, there's a second one as well. If it's possible logically perhaps better use some names and 1 or 2. But I'm OK with 1 or 2 although sounds a bit verbose.

kiryldz (commented):
@Chaoba few questions:
1. Why is this a breaking change?
2. Tests are still failing, I did not look that deep but because of multithreading it seems that QRF should be covered with instrument test and not unit one.

Chaoba:
> @Chaoba few questions:
> 
> 1. Why is this a breaking change?

change AnnotationManagerImpl#queryMapForFeatures function to synchronous

> 2. Tests are still failing, I did not look that deep but because of multithreading it seems that QRF should be covered with instrument test and not unit one.

Guess something wrong with CI
`could not GET 'https://jcenter.bintray.com/org/jacoco/org.jacoco.agent/0.8.4/org.jacoco.agent-0.8.4.jar'.`

Chaoba (approved):

tobrun (approved):
:shipit:

PR #458, opened 2021-06-29
Introduce option to use continuous rendering for scale bar

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce option to use continuous rendering for scale bar. Continuous render mode will fix gfxinfo profiling.</changelog>`.

### Summary of changes

Scale bar custom view was previously rendering on demand. Drawing was happening when camera was moving. However when scalebar is disabled it actually means that no more draw calls would happen. That actually leads to incorrect data from `gfxinfo` in terms of `Total frames rendered` as number of such frames would we always 3 or 4 despite map itself is rendering OK. This is actually not visible if scalebar is visible and map is moving but results are totally incorrect if we do `mapView.scalebar.enabled = false`.

In order to fix that we introduce and option to perform drawing for scalebar continuously now. This option should not be used by users in regular cases but might be useful to get correct rendered frames count info during let's say benchmarking.

### User impact (optional)

Default rendering way for scale bar will remain the same (rendering on demand). To change rendering mode to continuous `mapView.scalebar.useContinuousRendering = true` should be called.

### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
+++ b/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
@@ -131,6 +131,19 @@ open class ScaleBarPluginImpl(
     set(value) {
       scaleBar.distancePerPixel = value
     }
+
+  /**
+   * If set to True scale bar will be triggering onDraw depending on [ScaleBarSettings.refreshInterval]
+   * even if actual data did not change. If set to False scale bar will redraw only on demand.
+   *
+   * Defaults to False and should not be changed explicitly in most cases.
+   * Could be set to True to produce correct GPU frame metrics when running gfxinfo command.
+   */
+  override var useContinuousRendering: Boolean
+    get() = scaleBar.useContinuousRendering
+    set(value) {
+      scaleBar.useContinuousRendering = value
+    }
 }
 
 /**
--- a/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
+++ b/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
@@ -114,7 +114,10 @@ open class ScaleBarPluginImpl(
     get() = internalSettings.enabled
     set(value) {
       if (value) {
+        mapListenerDelegate.addOnCameraChangeListener(cameraChangeListener)
         invalidateScaleBar()
+      } else {
+        mapListenerDelegate.removeOnCameraChangeListener(cameraChangeListener)
       }
       internalSettings.enabled = value
       scaleBar.enable = value
--- a/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
+++ b/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
@@ -114,10 +114,7 @@ open class ScaleBarPluginImpl(
     get() = internalSettings.enabled
     set(value) {
       if (value) {
-        mapListenerDelegate.addOnCameraChangeListener(cameraChangeListener)
         invalidateScaleBar()
-      } else {
-        mapListenerDelegate.removeOnCameraChangeListener(cameraChangeListener)
       }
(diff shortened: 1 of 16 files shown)


Chaoba:
Is there any reason why we need to change this part?

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
```diff
@@ -114,10 +114,7 @@ open class ScaleBarPluginImpl(
     get() = internalSettings.enabled
     set(value) {
       if (value) {
-        mapListenerDelegate.addOnCameraChangeListener(cameraChangeListener)
         invalidateScaleBar()
-      } else {
-        mapListenerDelegate.removeOnCameraChangeListener(cameraChangeListener)
```
pengdev, line 120: why would we remoce thewse add/remove listeners?

kiryldz: correct, we do not need it. it was working OK but that's an incorrect solution, I've updated the code

Chaoba:
The default `refreshInterval` is 15, which means we will draw the scalebar at fps more than 60 all the time,  it will have lower performance and higher battery drain. We should put an eye on matrics after this pr merged.

kiryldz:
> The default `refreshInterval` is 15, which means we will draw the scalebar at fps more than 60 all the time, it will have lower performance and higher battery drain. We should put an eye on matrics after this pr merged.

I'll run some more tests. Although for me it seems that this should be fine indeed it might bring some troubles especially for battery drain. Refresh interval is not a big deal, rendering still depends on vsync, even if we draw too often we will just spend some CPU cycles but rendering will be the same as before.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPluginImpl.kt
```diff
@@ -136,5 +136,5 @@ open class ScaleBarPluginImpl(
 /**
  * Extension val for MapView to get the ScaleBar plugin instance.
  */
-val MapPluginProviderDelegate.scalebar: ScaleBarPluginImpl
```
pengdev, line 139: nice catch

tobrun (approved):
:shipit:

Chaoba (approved):

sdk-base/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarPlugin.kt
```diff
@@ -12,4 +13,13 @@ interface ScaleBarPlugin : ViewPlugin, MapSizePlugin, ScaleBarSettingsInterface
...
   var distancePerPixel: Float
+
+  /**
+   * If set to True scale bar will be triggering onDraw depending on [ScaleBarSettings.refreshInterval]
+   * even if actual data did not change. If set to False scale bar will redraw only on demand.
+   *
+   * Defaults to False and should not be changed explicitly in most cases.
+   * Could be set to True to produce correct GPU frame metrics when running gfxinfo command.
+   */
+  var useContinuousRendering: Boolean
```
pengdev, line 24: Looks like these two options both can be generated through serialisation spec generator, wdyt?

kiryldz: I was thinking about it but found several points against:
1. This API is public but rather debug one and we do not want encourage user to use it that badly.
2. Putting it to style spec will also make iOS implement it (in theory) but this option seems to be very Android specific.

pengdev: > This API is public but rather debug one and we do not want encourage user to use it that badly.

Well splitting the public API surface also doesn't sounds right 😄 We intend to ask users to adjust the settings using scaleBar.updateSettings { }, but these APIs won't be there..  I prefer the generated approach since it will offer consistent API surface, and easier to manager our public API surfaces.

> Putting it to style spec will also make iOS implement it (in theory) but this option seems to be very Android specific.

We have option to mark it as an Android-only property in the spec 😏

pengdev (approved):

PR #461, opened 2021-06-30
[Localization] Add new param to allow users localize selected layers.

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add new param to allow users localize selected layers.</changelog>`.

### Summary of changes
This pr add a new param `layerIds` for API `localizeLabels`,  `layerIds` is a list, user can put all the layers they want to localize in this list, other layers not in this list will not be localized.
The default value of `layerIds` is null which means localize all the feasible layers.

https://user-images.githubusercontent.com/8577318/123926832-dece2e80-d9be-11eb-9817-5a92245803c1.mp4

### DIFF
--- a/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
+++ b/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
@@ -19,15 +19,21 @@ import java.util.*
  * `["format",["coalesce",["get","name_en"],["get","name"]],{}]` will be replaced by
  * `["format",["coalesce",["get","name_zh-Hans"],["get","name"]],{}]`
  */
-internal fun setMapLanguage(locale: Locale, style: StyleInterface) {
+internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: List<String>?) {
   style.styleSources
     .filter { sourceIsFromMapbox(style, it) }
     .forEach { source ->
       style.styleLayers
         .filter { it.type == LAYER_TYPE_SYMBOL }
+        .filter { layer ->
+          layerIds?.contains(layer.id) ?: true
+        }
         .forEach { layer ->
           val symbolLayer = style.getLayerAs<SymbolLayer>(layer.id)
           symbolLayer.textFieldAsExpression?.let { textFieldExpression ->
+            if (BuildConfig.DEBUG) {
+              Logger.i(TAG, "Localize layer id: ${symbolLayer.layerId}")
+            }
             val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
             else getLanguageNameV7(locale)
             convertExpression(language, symbolLayer, textFieldExpression)
@@ -42,7 +48,9 @@ private fun convertExpression(language: String, layer: SymbolLayer, textField: E
       EXPRESSION_REGEX,
       get(language).toJson()
     ).replace(EXPRESSION_ABBR_REGEX, get(language).toJson())
-
+    if (BuildConfig.DEBUG) {
+      Logger.i(TAG, "Localize layer with expression: $stringExpression")
+    }
     layer.textField(Expression.fromRaw(stringExpression))
   }
 }
--- a/extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
+++ b/extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
@@ -5,7 +5,10 @@ import java.util.*
 
 /**
  * Extension function to localize style labels
+ *
+ * @param locale the locale that applied for localization
+ * @param layerIds the id of layers that will localize on, default is null which means will localize all the feasible layers.
  */
-fun StyleInterface.localizeLabels(locale: Locale) {
-  setMapLanguage(locale, this)
+fun StyleInterface.localizeLabels(locale: Locale, layerIds: List<String>? = null) {
+  setMapLanguage(locale, this, layerIds)
 }
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/localization/LocalizationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/localization/LocalizationActivity.kt
@@ -35,6 +35,7 @@ class LocalizationActivity : AppCompatActivity() {
     }
   private lateinit var locale: Locale
   private lateinit var selectedLocale: Locale
+  private var layerIdList = mutableSetOf<String>()
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_map_localization)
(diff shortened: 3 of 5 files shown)


extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
```diff
@@ -5,7 +5,10 @@ import java.util.*
 
 /**
  * Extension function to localize style labels
+ *
+ * @param locale the locale that applied for localization
+ * @param layerIds the id of layers that will localize on, default is null which means will localize all the feasible layers.
  */
-fun StyleInterface.localizeLabels(locale: Locale) {
-  setMapLanguage(locale, this)
+fun StyleInterface.localizeLabels(locale: Locale, layerIds: List<String>? = null) {
```
kiryldz, line 12: Perhaps use `emptyList()` as default parameter and make `layerIds` not nullable then? imho it will work the same but code will be cleaner

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -19,15 +19,21 @@ import java.util.*
...
     .forEach { source ->
       style.styleLayers
         .filter { it.type == LAYER_TYPE_SYMBOL }
+        .filter { layer ->
+          layerIds?.contains(layer.id) ?: true
+        }
         .forEach { layer ->
           val symbolLayer = style.getLayerAs<SymbolLayer>(layer.id)
           symbolLayer.textFieldAsExpression?.let { textFieldExpression ->
+            if (BuildConfig.DEBUG) {
```
kiryldz, line 34: I don't remember using that anywhere else in project. Do we really need that logging? I'd remove it, we can always add it manually if we really need it.

Chaoba: Actually, I have added and removed the log output many times and these logs are really important for developing. Is there any particular reason we can't use this?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
Chaoba: As shown in this [line](https://github.com/mapbox/mapbox-maps-android/pull/461/files#diff-4d722ef66fccbb18db39c8a9f28f6d79472a40ca68110bf356983e5888eb649bR29) a nullable list can make the code more cleaner.

kiryldz: Wouldn't `layerIds.contains(layer.id)` work the same here? 🤔

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -19,15 +19,21 @@ import java.util.*
...
  * `["format",["coalesce",["get","name_zh-Hans"],["get","name"]],{}]`
  */
-internal fun setMapLanguage(locale: Locale, style: StyleInterface) {
+internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: List<String>?) {
   style.styleSources
     .filter { sourceIsFromMapbox(style, it) }
     .forEach { source ->
       style.styleLayers
         .filter { it.type == LAYER_TYPE_SYMBOL }
+        .filter { layer ->
```
kiryldz, line 28: Could concat with previous filter with `&&`

extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
kiryldz: I see now, I guess not.
My only concern - will the code work OK if user would pass `emptyList()` explicitly?

Chaoba: This list contains the layers they want to localize, an empty list means no layers will be localized and the filter will filter out all the layers.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
Chaoba: Yeah, I have thought in this way too, but with another filter, both the code and logic are cleaner.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
kiryldz: So having an empty list is equivalent to passing null. In that case I'd make this parameter not nullable if it acts the same as null and pass empty list as default.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
kiryldz: I don't have any particular reasons.
@mapbox/maps-android WDYT?

pengdev: If this message isn't printed frequently, I'd keep the log. And probably we need a centralised solution to filter out log messages for specific plugins.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/StyleInterfaceExtension.kt
pengdev: If I understand correctly, passing empty list will result in no localization at any layer at all, please correct me if I'm wrong @Chaoba

kiryldz: I see that iOS is also using null by default, ok let's leave it this way to be aligned. Although passing empty list basically meaning that function will not work at all seems pretty useless to me.

Chaoba: Passing null will localize all feasible layers, which is the default action, while an empty list will not localize any layer.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
Chaoba: There would be about a dozen of log output as we have several symbol layers in the style.

pengdev (approved):

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
tobrun: I'm fine with using BuildConfig.DEBUG, users won't ever hit this. The filter log feature is being developed upstream atm.

tobrun (approved):

Issue #274, opened 2021-04-20
Bearing animator unwraps the value unexpectedly


LukasPaczos:
## Environment

- Maps SDK Version: 10.0.0-beta.17

## Observed behavior and steps to reproduce

In the below repro example, I start with a certain camera position. Then I click the map to initiate a rotation to a very small negative number. This results in the camera wrapping the negative value to 360. Then I try to transition again from 360 to 360 expecting no changes in bearing, but actually, the camera spins around.

## Expected behavior

The bearing doesn't change at all visually.

## Notes / preliminary analysis

Video:

https://user-images.githubusercontent.com/16925074/115408413-1cf5e780-a1f1-11eb-8e3c-5449fb553004.mp4

Reproducible code:
```kotlin
    private fun test(mapboxMap: MapboxMap, cameraPlugin: CameraAnimationsPlugin) {
        val listener = object : Animator.AnimatorListener {
            override fun onAnimationRepeat(animation: Animator?) {
                Log.e("TEST_CAMERA", "onAnimationRepeat: " + animation)
            }

            override fun onAnimationEnd(animation: Animator?) {
                Log.e("TEST_CAMERA", "onAnimationEnd: " + animation)
                Log.e("TEST_CAMERA", "onAnimationEnd options: " + mapboxMap.getCameraOptions())
            }

            override fun onAnimationCancel(animation: Animator?) {
                Log.e("TEST_CAMERA", "onAnimationCancel: " + animation)
            }

            override fun onAnimationStart(animation: Animator?) {
                Log.e("TEST_CAMERA", "onAnimationCancel: " + animation)
            }
        }
        mapboxMap.setCamera(
            CameraOptions.Builder()
                .center(Point.fromLngLat(17.045988781311138, 51.12341909978734))
                .bearing(27.254667247679752)
                .pitch(0.0)
                .padding(EdgeInsets(140.0, 140.0, 140.0, 140.0))
                .zoom(13.282170222962456)
                .build()
        )
        mapboxMap.addOnMapClickListener(object : OnMapClickListener {
            var firstClick = true
            override fun onMapClick(point: Point): Boolean {
                Log.e("TEST_CAMERA", "current options: " + mapboxMap.getCameraOptions())
                if (firstClick) {
                    Log.e("TEST_CAMERA", "first click")
                    firstClick = false

                    val centerAnimator = cameraPlugin.createCenterAnimator(
                        CameraAnimatorOptions.cameraAnimatorOptions(
[...]

PR #473, opened 2021-07-02
Add shortest bearing path option for animators

kiryldz:
Fixes: #274 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add shortest bearing path option for animators.</changelog>`.

### Summary of changes

 - When creating low-level `CameraBearingAnimator` new flag `optimizePath` (defaulted to true) was introduced. If this flag is set to `true` - shortest bearing path will be used. If set to `false` - camera will rotate clock-wise if next target value is greater or equal to current target and backwards otherwise.
 - In order to get proper values from bearing listeners that could be added to `CameraAnimationPlugin` all map camera values are now taken from current `MapboxMap.cameraState`. This leads that those values for bearing will taken directly from core and wrapped to [0, 360).

### User impact (optional)

May be a breaking change for Java users as API was extended with parameter with default value `useShortestPath`.
```
  fun createBearingAnimator(
    options: CameraAnimatorOptions<Double>,
    useShortestPath: Boolean = true,
    block: (ValueAnimator.() -> Unit)? = null
  ): ValueAnimator
```
Kotlin users will not see it as a breaking change in code but all low-level animators already created before will be using shortest bearing path for animation.


sdk-base/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPlugin.kt
```diff
@@ -140,18 +140,22 @@ interface CameraAnimationsPlugin : MapPlugin {
...
    *
    * @param options animator options object to set targets and other non mandatory options
+   * @param optimizePath if set to True shortest bearing path will be applied while animating bearing values.
+   * If set to False clock-wise rotation will be used if next target is greater or equal than current one
+   * and counter clock-wise rotation will be used if next target less than current one.
    * @param block optional block to apply any [ValueAnimator] parameters
    */
   fun createBearingAnimator(
     options: CameraAnimatorOptions<Double>,
+    optimizePath: Boolean = true,
```
macdrevx, line 153: I wish Swift allowed us to set default values in interface (protocol) definitions.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -694,8 +701,9 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
   override fun createBearingAnimator(
     options: CameraAnimatorOptions<Double>,
+    optimizePath: Boolean,
```
tobrun, line 704: The optimizePath naming doens't indicate how it's optimizing. Would it make sense renaming this to `shortestPath` or name similar to this?

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPlugin.kt
kiryldz: @macdrevx you could always switch to [Kotlin](https://kotlinlang.org/lp/mobile/) for developing iOS as well 😉

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -176,21 +174,30 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
         "Animation ${cameraAnimator.type.name}(${cameraAnimator.hashCode()}): automatically setting start value $it."
       )
     }
-
-    val targets = cameraAnimator.targets
-    cameraAnimator.setObjectValues(
-      *Array(targets.size + 1) { index ->
+    val targets = if (cameraAnimator is CameraBearingAnimator && cameraAnimator.optimizePath) {
+      MathUtils.prepareOptimalBearingPath(
+        DoubleArray(cameraAnimator.targets.size + 1) { index ->
```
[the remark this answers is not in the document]
kiryldz: Sadly did not manage to get it to work that way 😢

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: Indeed, using `useShortestPath` flag now.

sdk-base/src/main/java/com/mapbox/maps/util/MathUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+    return optimized
+  }
+
+  /**
+   * Util for finding the shortest path from the current rotated degree to the new degree.
+   *
+   * @param targetHeading  the new position of the rotation
+   * @param previousHeading the current position of the rotation
```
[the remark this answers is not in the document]
kiryldz: Fixed.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimatorManager.kt
```diff
@@ -62,16 +62,7 @@ internal class PuckAnimatorManager(
...
-    targets.toTypedArray().apply {
-      for (i in 0 until size) {
-        optimized[i] = if (i == 0)
-          MathUtils.normalize(get(i))
-        else
-          MathUtils.shortestRotation(MathUtils.normalize(get(i)), optimized[i - 1])
-      }
-    }
-    bearingAnimator.animate(*optimized.toTypedArray(), options = options)
+    bearingAnimator.animate(*MathUtils.prepareOptimalBearingPath(targets).toTypedArray(), options = options)
```
LukasPaczos, line 65: I don't think we want to optimize the path for the puck. If a turn is tighter than 180 degrees, we still want to follow the true rotation, not use the shortest path.

LukasPaczos: Capturing a chat we had with @kiryldz - in the end, the baked-in shortest path is probably a safer assumption to have, it will work okay 99% of the time without developers having to think. We can always expose a `shortestPath` flag in the future if this becomes a request.

sdk-base/src/main/java/com/mapbox/maps/util/MathUtils.kt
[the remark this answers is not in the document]
Chaoba: It seems you fixed the wrong param. Should be `nextHeading` and `currentHeading`

Chaoba (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -180,21 +178,30 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
         }
       }
-    )
+    }
+    cameraAnimator.setObjectValues(*targets)
     return true
   }
 
-  internal fun notifyListeners(cameraOptions: CameraOptions) {
+  internal fun notifyListeners(cameraOptions: CameraState) {
```
pengdev, line 204: nit rename the variable to cameraState

kiryldz: nice catch, thanks!

pengdev:
> May be a breaking change for Java users as API was extended with parameter with default value useShortestPath.

We could maybe use [JvmOverloads](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/) annotation, so the Kotlin compiler can generate overloads for this function that substitute default parameter values.

Thus it wouldn't be a breaking change for both Java and Kotlin.

kiryldz:
> > May be a breaking change for Java users as API was extended with parameter with default value useShortestPath.
> 
> We could maybe use [JvmOverloads](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/) annotation, so the Kotlin compiler can generate overloads for this function that substitute default parameter values.
> 
> Thus it wouldn't be a breaking change for both Java and Kotlin.

Sadly is not supported for interfaces 😢

pengdev (approved):
> Sadly is not supported for interfaces 😢

😢

Issue #243, opened 2021-04-08
Allow to turn off CameraAnimationPlugin logs


yuryybk:
## New Feature
Allow to turn off CameraAnimationPlugin logs on the app side

## Why
CameraAnimationPlugin logs are too frequent and prevent analysis in some cases

tobrun:
Duplicate of internal ticket

kiryldz:
As additional request some frequent log messages will be downgraded to `verbose` instead of `info`

PR #474, opened 2021-07-02
Add API to control logs for animators

kiryldz:
Fixes: #243 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add API to control logging for animation plugin and disable debug logs by default.</changelog>`.

### Summary of changes

Add API to disable not critical logs for camera plugin. `debugMode` is set to false by default. Logs could be enabled with:

```
mapView.camera.debugMode = true
```

Additionally even if logs are enabled non-critical logs are downgraded to use `Logger.d` instead of `Logger.i`.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -57,6 +57,8 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
   private val lifecycleListeners = CopyOnWriteArraySet<CameraAnimationsLifecycleListener>()
 
+  override var debugMode: Boolean = true
+
   private var center by Delegates.observable<Point?>(null) { _, old, new ->
     new?.let {
       if (old != it) {
@@ -171,10 +173,12 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
       CameraAnimatorType.BEARING -> mapCameraManagerDelegate.cameraState.bearing
       CameraAnimatorType.PITCH -> mapCameraManagerDelegate.cameraState.pitch
     }.also {
-      Logger.i(
-        TAG,
-        "Animation ${cameraAnimator.type.name}(${cameraAnimator.hashCode()}): automatically setting start value $it."
-      )
+      if (debugMode) {
+        Logger.d(
+          TAG,
+          "Animation ${cameraAnimator.type.name}(${cameraAnimator.hashCode()}): automatically setting start value $it."
+        )
+      }
     }
 
     val targets = cameraAnimator.targets
@@ -262,7 +266,9 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
           if (isUpdated) {
             // finally register update listener in order to update map properly
             registerInternalUpdateListener(this)
-            Logger.i(TAG, "Animation ${type.name}(${hashCode()}) started.")
+            if (debugMode) {
+              Logger.d(TAG, "Animation ${type.name}(${hashCode()}) started.")
+            }
           }
         } ?: throw RuntimeException(
           "Could not start animation in CameraManager! " +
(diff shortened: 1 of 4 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -57,6 +57,8 @@ internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
   private val lifecycleListeners = CopyOnWriteArraySet<CameraAnimationsLifecycleListener>()
 
+  override var debugMode: Boolean = true
```
Chaoba, line 60: Can we make it false by default? Users who want to see the log can enable it intentionally, but I don't think it's a common requirement for most users.

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
```diff
@@ -962,6 +963,21 @@ class CameraAnimationsPluginImplTest {
     assertEquals(10.0, updateList[1].pitch!!, EPS)
   }
 
+  @Test
+  fun debugModeTest() {
+    // debug mode is enabled by default for backward compatibility
+    shadowOf(getMainLooper()).pause()
+    cameraAnimationsPluginImpl.easeTo(cameraOptions, mapAnimationOptions { duration(DURATION) })
+    shadowOf(getMainLooper()).idle()
+    assertNotEquals(0, ShadowLogger.logCount)
```
Chaoba, line 972: If I understand correctly when debug mode is enabled the logcount should not be 0.

tobrun:
Upstream we just merged the ability to filter out categories from common SDK. I rather see us moving to that solution instead. We will however need to add the log configuration by default to disable log messages for users and remove the filter if we want to enable it again.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
kiryldz: I've set it to true in order to not make it a breaking change in terms of behaviour as now those logs are always enabled.

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
kiryldz: Correct, that's why I'm using here `assertNotEquals` 😉

tobrun:
> Upstream we just merged the ability to filter out categories from common SDK. I rather see us moving to that solution instead. We will however need to add the log configuration by default to disable log messages for users and remove the filter if we want to enable it again.

After internal convo, le'ts move forward on exposing debugMode. This can use the internal API under the hood.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
tobrun: I also would prefer false as default. I don't consider logs to be part of breaking changes.

kiryldz: thanks for your thoughts here, folks, will disable by default 👍

Chaoba (approved):
LGTM

tobrun (approved):

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPlugin.kt
```diff
@@ -12,6 +12,12 @@ import com.mapbox.maps.plugin.MapPlugin
  */
 interface CameraAnimationsPlugin : MapPlugin {
 
+  /**
+   * If debug mode is enabled extra logs will be written about animation lifecycle and
+   * some other events that may be useful for debugging.
+   */
+  var debugMode: Boolean
```
pengdev, line 19: shall we generate this config option for all the plugins? cc @kiryldz @tobrun @Chaoba

kiryldz: No, I've taken a look at other plugins, there's no sense in adding it everywhere. We don't have that massive logs anywhere. That's why I closed internal tickets related to similar API for all the plugins and decided to implement it in camera only due to a lot of requests.

Issue #475, opened 2021-07-06
Click to the compass does nothing when map is used with NavigationCamera


zeac:
When camera is configured like this
```
fun getSinglePixelEdgeInsets(mapSize: Size): EdgeInsets {
    val point = ScreenCoordinate(
        mapSize.width.toDouble() / 2.0,
        mapSize.height.toDouble() / 2.0
    )
    return EdgeInsets(point.y, point.x, mapSize.height - point.y, mapSize.width - point.x)
}
 
CameraOptions.Builder()
    .padding(getSinglePixelEdgeInsets(mapboxMap.getSize()))
    .zoom(15.900000000000004)
    .bearing(266.929321)
    .build()
```

flyTo animation is not working:
```mapView.camera.flyTo(CameraOptions.Builder().bearing(0.0).build())```

It happens because CameraZoomAnimator's evaluator returns NaN.

Before https://github.com/mapbox/mapbox-maps-android/pull/352 click on compass did nothing.
After click on compass rotates camera to the north without animation.

Here is the broken test for this problem: https://github.com/mapbox/mapbox-maps-android/pull/476

Issue #177, opened 2021-03-16
CameraAnimatorsFactory#getFlyTo crashes when padding is a single pixel


LukasPaczos:
In the Navigation Camera use-case, we need to squish the padding into a single pixel to adjust the vanishing point of the camera. However, when this is done, the `CameraAnimatorsFactory#getFlyTo` crashes:
```
Process: com.mapbox.navigation.test_app, PID: 29587
java.lang.Error: latitude must not be NaN
    at com.mapbox.maps.Projection.unproject(Native Method)
    at com.mapbox.maps.MapboxMap.unproject(MapboxMap.kt:636)
    at com.mapbox.maps.plugin.animation.CameraAnimatorsFactory$getFlyTo$animators$2.evaluate(CameraAnimatorsFactory.kt:395)
    at com.mapbox.maps.plugin.animation.CameraAnimatorsFactory$getFlyTo$animators$2.evaluate(CameraAnimatorsFactory.kt:27)
    at android.animation.KeyframeSet.getValue(KeyframeSet.java:202)
    at android.animation.PropertyValuesHolder.calculateValue(PropertyValuesHolder.java:1017)
    at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1561)
    at android.animation.ValueAnimator.setCurrentFraction(ValueAnimator.java:692)
    at android.animation.ValueAnimator.setCurrentPlayTime(ValueAnimator.java:655)
    at android.animation.ValueAnimator.start(ValueAnimator.java:1087)
    at android.animation.ValueAnimator.start(ValueAnimator.java:1106)
    at com.mapbox.maps.plugin.animation.animator.CameraAnimator.start(CameraAnimator.kt:88)
    at android.animation.ValueAnimator.startWithoutPulsing(ValueAnimator.java:1099)
    at android.animation.AnimatorSet.handleAnimationEvents(AnimatorSet.java:1149)
    at android.animation.AnimatorSet.startAnimation(AnimatorSet.java:1234)
    at android.animation.AnimatorSet.start(AnimatorSet.java:729)
    at android.animation.AnimatorSet.start(AnimatorSet.java:684)
    at com.mapbox.maps.plugin.animation.CameraAnimationsPluginImpl.startAnimatorSet(CameraAnimationsPluginImpl.kt:338)
    at com.mapbox.maps.plugin.animation.CameraAnimationsPluginImpl.flyTo(CameraAnimationsPluginImpl.kt:699)
```

[...]

PR #478, opened 2021-07-06
Fix flyTo crash when using single-pixel paddings

kiryldz:
Fixes: #475 
Fixes: https://github.com/mapbox/mapbox-maps-android/issues/177

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix flyTo crash when using single-pixel paddings.</changelog>`.

### Summary of changes

Fix a crash (if changing center as well) or zoom reporting NaN (if changing zoom) when using `mapView.camera.flyTo` with 1-pixel padding or `mapView.width == padding.left + padding.right && mapView.height == padding.bottom + padding.top`

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
@@ -280,12 +280,13 @@ class CameraAnimatorsFactory internal constructor(mapDelegateProvider: MapDelega
 
     val size = mapTransformDelegate.getSize()
     val pixelRatio = mapTransformDelegate.getMapOptions().pixelRatio
-    val w0 = if (size.width.toDouble() != endPadding.left + endPadding.right
-      && size.height.toDouble() != endPadding.bottom + endPadding.top) {
-        max(
-          (size.width - endPadding.left - endPadding.right) / pixelRatio,
-          (size.height - endPadding.top - endPadding.bottom) / pixelRatio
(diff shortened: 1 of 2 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactory.kt
```diff
@@ -280,10 +280,19 @@ class CameraAnimatorsFactory internal constructor(mapDelegateProvider: MapDelega
 
     val size = mapTransformDelegate.getSize()
     val pixelRatio = mapTransformDelegate.getMapOptions().pixelRatio
-    val w0 = max(
-      (size.width - endPadding.left - endPadding.right) / pixelRatio,
-      (size.height - endPadding.top - endPadding.bottom) / pixelRatio
-    )
+    val w0 = if (size.width.toDouble() != endPadding.left + endPadding.right &&
+      size.height.toDouble() != endPadding.bottom + endPadding.top
```
Chaoba, line 284: Is it better to use`>` to replace `!=`?

kiryldz: I'd say no. We must exclude equality in order to avoid division by zero. However if paddings are set in a way that they in total are greater than size - it seems to be more of an issue in this logic but flyTo algo will work OK in that case.

Chaoba (approved):
LGTM

PR #485, opened 2021-07-09
Add lifecycle plugin

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add lifecycle plugin, users don't need to call onStart/onStop/onDestroy methods with this plugin</changelog>`.

### Summary of changes
This pr add lifecycle plugin, which will listen to `onStart/onStop/onDestroy/onLowMemory` events of the host Activity/Fragment and invoke the corresponding methods of MapView. So that users don't need to call these 4 methods anymore.


tobrun (commented):
Thank you for kicking of this discovery! I would love to see if we can make this truly automatic, this would however require to convert the extension to an actual plugin. Concretely what I mean is that developers don't need to call `attachLifeCycle(mapView,this)` but this is done by the plugin itself.

tobrun:
After talk with @Chaoba, the big issue about making this a plugin is that we don't know if the MapView is part of either a fragment or an activity to make sure we attach to the right component. For this reason, we can't fully automate the integration to be a true plugin but will have to rely on it being an extension instead.

tobrun:
After a search, I stumbled on https://developer.android.com/reference/androidx/lifecycle/ViewTreeLifecycleOwner which might allow us to not need the attachToLifecycle call and make it completely transparent and not require users to call anything.

app/src/main/java/com/mapbox/maps/testapp/examples/SimpleMapActivity.kt
```diff
@@ -23,26 +23,14 @@ class SimpleMapActivity : AppCompatActivity() {
       }
   }
 
-  override fun onStart() {
-    super.onStart()
-    mapView.onStart()
+  override fun onResume() {
+    super.onResume()
```
kiryldz, line 27: Why do we need it?

plugin-lifecycle/README.md
```diff
@@ -0,0 +1,41 @@
...
+                password = "INSERT_MAPBOX_ACCESS_TOKEN_HERE"
+            }
+        }
+    }
+}
+
+// In the app build.gradle file
+dependencies {
+  implementation 'com.mapbox.plugin:maps-lifecycle:10.0.0-rc.4'
+  // Make sure the version of appcompat is 1.3.0+
```
kiryldz, line 34: Did you think of putting `appcompat` dependency inside `lifecycle` plugin? Although I'm OK with current solution but just to make sure that idea was also given a shot.

kiryldz: well, I've taken a look at gradle file of `lifecycle` plugin and appcompat is actually there. Do we need to specify it in dependency then?

plugin-lifecycle/gradle.properties
```diff
@@ -0,0 +1,6 @@
+POM_GROUP_ID=com.mapbox.extension
```
kiryldz, line 1: it's now a plugin, isn't it?

sdk-base/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifeCyclePlugin.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+fun interface MapboxLifeCyclePlugin : MapPlugin {
+
+  /**
+   * Register a MapboxLifeCycleObserver to observe life cycle events from LifecycleOwner
+   *
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
+   */
+  fun registerLifeCycleObserver(mapView: FrameLayout, observer: MapboxLifeCycleObserver)
```
kiryldz, line 18: I'd go with `registerLifecycleObserver`, no need to emphasize word `cycle` imho

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
```diff
@@ -69,8 +69,7 @@ class StyleLoadTest {
       }
     }
     countDownLatch.await(10, TimeUnit.SECONDS)
-    // we should notify only once that style is loaded
-    Assert.assertEquals(1, styleLoadedCount)
+    Assert.assertEquals(2, styleLoadedCount)
```
kiryldz, line 72: why did it change?

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
```diff
@@ -41,6 +42,7 @@ MapPluginRegistry(private val mapDelegateProvider: MapDelegateProvider) {
   private val gesturePlugins = CopyOnWriteArrayList<GesturesPlugin>()
   private val styleObserverPlugins = CopyOnWriteArrayList<MapStyleObserverPlugin>()
   private val mapSizePlugins = CopyOnWriteArrayList<MapSizePlugin>()
+  private val mapboxLifecyclePlugins = CopyOnWriteArrayList<MapboxLifeCyclePlugin>()
```
kiryldz, line 45: imho it could the only such plugin, perhaps use just `var` and not a list?

plugin-lifecycle/README.md
```diff
@@ -0,0 +1,39 @@
+## Mapbox Maps Lifecycle Plugin for Android
+
+### Overview
+
+The Mapbox Maps Lifecycle Plugin for Android is an public library for handling Lifecycle of MapView. By default, Lifecycle Plugin will automatically listen to the lifecycle events(onStart/onStop/onDestroy) of the host Activity or Fragment, so that users don't need to invoke the corresponding lifecycle methods of MapView in these three lifecycle methods.
```
tobrun, line 5:
```suggestion
The Mapbox Maps Lifecycle Plugin for Android is an public library for handling Lifecycle events of the Mapbox SDK MapView. By default, the Lifecycle Plugin will automatically hook into the lifecycle events(onStart/onStop/onDestroy) of the host Activity or Fragment, so that developers doesn't need to manually invoke the corresponding lifecycle methods of MapView themselves.
```

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
tobrun: I would prefer changing the test setup itself to accommodate the changes with lifecycle plugin. The test was written to validate style load. Validating that logic kind falls apart if we are loading 2 styles.

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -100,6 +100,10 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
     mapController.initializePlugins(resolvedMapInitOptions, this)
   }
 
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    mapController.onAttachedToWindow(this)
+  }
```
tobrun, line 106:
```suggestion
  }

```

tobrun (commented):
While this is not a breaking change, it does change business logic as it potentially executes the lifecycle methods twice. I think we need to clearly note this in the changelog. 

----

What happens if a developer doesn't use AppCompatActivity? (eg. Activity)

---- 

Should we harden the execution of onStart/onStop to only execute in pairs once (and onDestroy once)? cc @kiryldz

Chaoba:
> What happens if a developer doesn't use AppCompatActivity? (eg. Activity)

We will not be able to hook the lifecycle observer and `onStart` method will not be invoked and the MapView is back.

app/src/main/java/com/mapbox/maps/testapp/examples/SimpleMapActivity.kt
```diff
@@ -23,26 +23,11 @@ class SimpleMapActivity : AppCompatActivity() {
...
-    super.onStart()
-    mapView.onStart()
-  }
-
-  override fun onStop() {
-    super.onStop()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
```
pengdev, line 26: can we handle the `onLowMemory` callback as well in the lifecycle plugin? we could maybe use `Context. registerComponentCallbacks` API to observe the onLowMemory event.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+ */
+class MapboxLifecyclePluginImpl : MapboxLifecyclePlugin {
+  /**
+   * Register a MapboxLifecycleObserver to observe life cycle events from LifecycleOwner
+   *
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
+   */
+  override fun registerLifecycleObserver(mapView: FrameLayout, observer: MapboxLifecycleObserver) {
+    ViewTreeLifecycleOwner.get(mapView)?.apply {
```
pengdev, line 22: What if the `ViewTreeLifecycleOwner.get(mapView)` returns null? shall we log some error messages or exceptions to notify the developer?

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -100,6 +100,11 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
     mapController.initializePlugins(resolvedMapInitOptions, this)
   }
 
+  override fun onAttachedToWindow() {
+    super.onAttachedToWindow()
+    mapController.onAttachedToWindow(this)
```
pengdev, line 105: do we have similar construct to the `MapSurface`?

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
Chaoba: The screen would be black. Yeah, we could log the error message.

sdk-base/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePlugin.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+fun interface MapboxLifecyclePlugin : MapPlugin {
+
+  /**
+   * Register a MapboxLifecycleObserver to observe life cycle events from LifecycleOwner
+   *
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
+   */
+  fun registerLifecycleObserver(mapView: FrameLayout, observer: MapboxLifecycleObserver)
```
pengdev, line 18: Shall we make it generic and use `LifecycleOwner` instead of `mapView: FrameLayout`? this way it is easier to integrate with Surface as well.

Chaoba: In order to get the LifecycleOwner, we must have the view firstly. However, if we want to support MapSurface, we will need to pass in a SurfaceView instance as a param for MapSurface. This will be a breaking change and example `MapWallpaper` will not works anymore since there is no view in it. I have upload a branch for MapSurface https://github.com/mapbox/mapbox-maps-android/compare/kl-lifecycle...kl-surface?expand=1
IMHO, we should not support MapSurface in this plugin.

tobrun: > In order to get the LifecycleOwner, we must have the view firstly. However, if we want to support MapSurface, we will need to pass in a SurfaceView instance as a param for MapSurface. This will be a breaking change and example MapWallpaper will not works anymore since there is no view in it. I have upload a branch for MapSurface https://github.com/mapbox/mapbox-maps-android/compare/kl-lifecycle...kl-mapsurface?expand=1
IMHO, we should not support MapSurface in this plugin.

:+1:, I do think instead of FrameLayout we can reference "View" instead.

plugin-lifecycle/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    targetSdkVersion(AndroidVersions.targetSdkVersion)
+    testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+  }
+}
+
+dependencies {
+  implementation(project(":sdk-base"))
+  implementation(Dependencies.mapboxBase)
+  implementation(Dependencies.kotlin)
+  implementation(Dependencies.androidxAppCompat)
```
pengdev, line 22: Is this needed?

sdk/build.gradle.kts
```diff
@@ -60,6 +60,7 @@ dependencies {
   api(project(":plugin-overlay"))
   api(project(":plugin-annotation"))
   api(project(":extension-localization"))
+  api(project(":plugin-lifecycle"))
```
pengdev, line 63: Do we actually want to bundle the lifecycle plugin with the sdk? since the lifecycle plugin depends on   `implementation(Dependencies.androidxAppCompat)` and `implementation(Dependencies.androidxLifecycle)`, and could introduce some binary size increase.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
+   */
+  override fun registerLifecycleObserver(mapView: FrameLayout, observer: MapboxLifecycleObserver) {
+    val lifecycleOwner = ViewTreeLifecycleOwner.get(mapView)
+    if (lifecycleOwner == null) {
+      Logger.w(
+        TAG,
+        "Can't get lifecycleOwner for mapview, please make sure the host Activity is AppCompatActivity"
```
pengdev, line 29: Could this be an error message? and we should provide what to do in this case, if the host Activity is not AppCompatActivity, the user should follow the instructions to manually overwrite their activity/fragment's lifecycle events.

sdk-base/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePlugin.kt
Chaoba: Yeah, we can use View here.

sdk/build.gradle.kts
tobrun: `androidxAppCompat` is the standard these days, 99% of developers will include them in their apps.
That we rely on lifecycle is also an expected outcome of this plugin.

plugin-lifecycle/build.gradle.kts
Chaoba: We can move it to app module.

sdk/build.gradle.kts
Chaoba: Yeah, androidxAppCompat can moved to app module, so only androidxLifecycle is included. Users could exclude this plugin if they are size sensitive.

plugin-lifecycle/README.md
```diff
@@ -0,0 +1,41 @@
+## Mapbox Maps Lifecycle Plugin for Android
+
+### Overview
+
+The Mapbox Maps Lifecycle Plugin for Android is an public library for handling Lifecycle events of the Mapbox SDK MapView. By default, the Lifecycle Plugin will automatically hook into the lifecycle events(onStart/onStop/onDestroy) of the host Activity or Fragment, so that developers doesn't need to manually invoke the corresponding lifecycle methods of MapView themselves.
```
kiryldz, line 5:
```suggestion
The Mapbox Maps Lifecycle Plugin for Android is an public library for handling Lifecycle events of the Mapbox SDK MapView. By default, the Lifecycle Plugin will automatically hook into the lifecycle events(onStart/onStop/onDestroy) of the host Activity or Fragment, so that developers don't need to invoke the corresponding lifecycle methods of MapView manually themselves.
```

plugin-lifecycle/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+A full overview of classes and interfaces can be found in our [API documentation](https://docs.mapbox.com/android/beta/maps/guides/).
```
kiryldz, line 7: Wrong link to guides (instead of API reference)

plugin-lifecycle/README.md
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// In the app build.gradle file
+dependencies {
+  implementation 'com.mapbox.plugin:maps-lifecycle:10.0.0-rc.4'
+  // Make sure the version of appcompat is 1.3.0+
+  implementation 'androidx.appcompat:appcompat:1.3.0'
+}
+```
+The Lifecycle Plugin will take effect automatically if added to dependency
```
kiryldz, line 38:
```suggestion
The Lifecycle Plugin will take effect automatically if added as dependency.
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -95,6 +96,11 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun onStart() {
+    if (lifecycleState != LIFECYCLE_STATE.STATE_PAUSED) {
```
kiryldz, line 99: Shouldn't we check for `STATE_STARTED` here?

kiryldz: imho check should be `if (lifecycleState == LIFECYCLE_STATE.STATE_STARTED)`

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -269,12 +285,23 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
   }
 
+  internal fun onAttachedToWindow(mapView: MapView) {
+    pluginRegistry.onAttachedToWindow(mapView)
+  }
+
+  private enum class LIFECYCLE_STATE {
+    STATE_PAUSED,
```
kiryldz, line 293: Could it be `STATE_STOPPED` as we call it in `onStop`?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -110,11 +116,12 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
 
-  fun reduceMemoryUse() {
-    mapboxMap.reduceMemoryUse()
-  }
-
   override fun onStop() {
+    if (lifecycleState != LIFECYCLE_STATE.STATE_STARTED) {
+      return
+    }
+    lifecycleState = LIFECYCLE_STATE.STATE_PAUSED
```
kiryldz, line 123:
```suggestion
    if (lifecycleState == LIFECYCLE_STATE.STATE_STOPPED) {
      return
    }
    lifecycleState = LIFECYCLE_STATE.STATE_STOPPED
```

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
```diff
@@ -41,6 +42,7 @@ MapPluginRegistry(private val mapDelegateProvider: MapDelegateProvider) {
   private val gesturePlugins = CopyOnWriteArrayList<GesturesPlugin>()
   private val styleObserverPlugins = CopyOnWriteArrayList<MapStyleObserverPlugin>()
   private val mapSizePlugins = CopyOnWriteArrayList<MapSizePlugin>()
+  private var mapboxLifecyclePlugins: MapboxLifecyclePlugin? = null
```
kiryldz, line 45:
```suggestion
  private var mapboxLifecyclePlugin: MapboxLifecyclePlugin? = null
```

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
```diff
@@ -35,4 +35,8 @@ const val PLUGIN_MAPOVERLAY_CLASS_NAME = "com.mapbox.maps.plugin.overlay.MapOver
 /**
  * Class name for annotation plugin.
  */
-const val PLUGIN_ANNOTATION_CLASS_NAME = "com.mapbox.maps.plugin.annotation.AnnotationPluginImpl"
\ No newline at end of file
+const val PLUGIN_ANNOTATION_CLASS_NAME = "com.mapbox.maps.plugin.annotation.AnnotationPluginImpl"
+/**
+ * Class name for life cycle plugin.
```
kiryldz, line 40:
```suggestion
 * Class name for lifecycle plugin.
```

sdk-base/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePlugin.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.mapbox.maps.MapboxLifecycleObserver
+import com.mapbox.maps.plugin.MapPlugin
+
+/**
+ * Defines interface of a MapViewPlugin.
+ */
+fun interface MapboxLifecyclePlugin : MapPlugin {
+
+  /**
+   * Register a MapboxLifecycleObserver to observe life cycle events from LifecycleOwner
```
kiryldz, line 13:
```suggestion
   * Register the MapboxLifecycleObserver to observe lifecycle events from LifecycleOwner
```

sdk-base/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePlugin.kt
```diff
@@ -0,0 +1,19 @@
...
+   *
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
```
kiryldz, line 16:
```suggestion
   * @param observer observer listening to lifecycle events
```

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -108,14 +108,15 @@ class MapControllerTest {
 
   @Test
   fun onStop() {
+    mapController.onStart()
```
kiryldz, line 111: why do we need `onStart()` here?

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
```diff
@@ -26,7 +26,7 @@ class StyleLoadTest {
   @UiThreadTest
   fun setUp() {
     rule.scenario.onActivity {
-      mapView = MapView(it)
+      mapView = MapView(it, MapInitOptions(it, plugins = listOf()))
```
kiryldz, line 29: Can you please clarify a bit why having lifecycle enabled actually will trigger `styleLoadedCount` twice?

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
Chaoba: Update the state to started stated

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
Chaoba: With lifecycle, onStart method will be triggered automatically and since there is no style set, the default style will be loaded automatically.

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
kiryldz: the test looks pretty strange then
can we introduce new enum value `LifecycleState.STATE_INIT` instead of setting `lifecycleState` to `LifecycleState.STATE_STOPPED`? why is it stopped by default?

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
kiryldz: Oh, I see, good point.
Perhaps we may want separate tests then to check it? And add a comment that if we have lifecycle plugin enabled - we will actually check that `styleLoadedCount` is triggered twice (and drop a comment why it happens like this as this is not obvious).

pengdev: Wondering how this would affect our Map initialisation process, would it cause surprise to the customers that some event might be triggered multiple times?

app/src/main/java/com/mapbox/maps/testapp/examples/SimpleMapActivity.kt
```diff
@@ -23,26 +23,11 @@ class SimpleMapActivity : AppCompatActivity() {
...
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 29: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/AnimatePointAnnotationActivity.kt
```diff
@@ -140,27 +140,11 @@ class AnimatePointAnnotationActivity : AppCompatActivity(), OnMapLoadedListener
...
-  override fun onStop() {
-    super.onStop()
-    cleanAnimation()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 146: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/CircleAnnotationActivity.kt
```diff
@@ -101,26 +101,11 @@ class CircleAnnotationActivity : AppCompatActivity() {
...
-
-  override fun onStop() {
-    super.onStop()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 107: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PointAnnotationActivity.kt
```diff
@@ -297,27 +297,11 @@ class PointAnnotationActivity : AppCompatActivity() {
...
-  override fun onStop() {
-    super.onStop()
-    animators.forEach { it.cancel() }
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 303: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PointAnnotationClusterActivity.kt
```diff
@@ -116,26 +116,11 @@ class PointAnnotationClusterActivity : AppCompatActivity(), CoroutineScope {
...
-
-  override fun onStop() {
-    super.onStop()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 122: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PolygoneAnnotationActivity.kt
```diff
@@ -106,23 +106,8 @@ class PolygoneAnnotationActivity : AppCompatActivity() {
...
-
-  override fun onStop() {
-    super.onStop()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 112: nit: can be removed since this is handled by the lifecycle plugin.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PolylineAnnotationActivity.kt
```diff
@@ -114,26 +114,11 @@ class PolylineAnnotationActivity : AppCompatActivity() {
...
-
-  override fun onStop() {
-    super.onStop()
-    mapView.onStop()
-  }
-
   override fun onLowMemory() {
     super.onLowMemory()
     mapView.onLowMemory()
   }
```
pengdev, line 120: nit: can be removed since this is handled by the lifecycle plugin.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -125,12 +132,21 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+
     mapboxMap.onDestroy()
     nativeObserver.clearListeners()
     renderer.onDestroy()
     pluginRegistry.cleanup()
   }
 
+  override fun onLowMemory() {
+    mapboxMap.reduceMemoryUse()
+  }
```
pengdev, line 148: Shall we add some state checks similar to onStart etc, to avoid calling mapboxMap.reduceMemoryUse() multiple times?

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
Chaoba: A normal lifecycle should be onStart->onStop. We should not let onStop make any changes to the sdk before onStart is invoked firstly.

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
Chaoba: No event is triggered multiple times, simply because we [load style](StyleLoadTest.kt) again, plus the one we load automatically in onResume, it turns out style is load 2 times.

_Perhaps we may want separate tests then to check it? And add a comment that if we have lifecycle plugin enabled - we will actually check that styleLoadedCount is triggered twice_

That's a good point to verify this plugin, will add new tests for it.

sdk/src/main/java/com/mapbox/maps/MapController.kt
Chaoba: No, this is not in the regular lifecycle and could be invoked multi times.

sdk/src/androidTest/java/com/mapbox/maps/LifecycleTest.kt
```diff
@@ -0,0 +1,53 @@
...
+  fun setUp() {
+    rule.scenario.onActivity {
+      mapView = MapView(it, MapInitOptions(it, plugins = listOf(PLUGIN_LIFE_CYCLE_CLASS_NAME)))
+      mapboxMap = mapView.getMapboxMap()
+      it.frameLayout.addView(mapView)
+    }
+  }
+
+  @Test
+  fun testOnStart() {
```
kiryldz, line 38: Ironically I don't see `onStart` being called in this test 😄 
Perhaps we do want to rename it and add another test with calling `onStart`, wdyt?

kiryldz: I actually meant another test calling `onStart` explicitly (here it's called automatically because of lifecycle plugin). to verify that we have same behaviour.

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
kiryldz: Yeah, I guess it sounds logical.

kiryldz (approved):
LGTM with last question tests related.
Amazing work @Chaoba! I believe we could proceed with merging and iterate later if needed.

app/build.gradle.kts
```diff
@@ -72,6 +72,7 @@ dependencies {
   implementation(Dependencies.squareRetrofit)
   implementation(Dependencies.androidxFragmentTest)
   implementation(Dependencies.squareRetrofitGsonConverter)
+  implementation(Dependencies.androidxAppCompat)
```
pengdev, line 75: nit: this dependency already exists above.

app/src/main/java/com/mapbox/maps/testapp/examples/annotation/AnimatePointAnnotationActivity.kt
pengdev: seems the onLowMemory still exists :D

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+   */
+  override fun registerLifecycleObserver(mapView: View, observer: MapboxLifecycleObserver) {
+    val lifecycleOwner = ViewTreeLifecycleOwner.get(mapView)
+    if (lifecycleOwner == null) {
+      Logger.e(
+        TAG,
+        """Can't get lifecycleOwner for mapview,
+          please make sure the host Activity is AppCompatActivity and the version of appcompat is 1.3.0+.
+          If the host Activity is not AppCompatActivity,
+          you need manually invoke the corresponding lifecycle methods in onStart/onStop/onDestroy methods of the host Activity"""
```
pengdev, line 32:
```suggestion
          you need manually invoke the corresponding lifecycle methods in onStart/onStop/onDestroy/onLowMemory methods of the host Activity"""
```

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+   * @param mapView the instance of mapView, will get the LifecycleOwner from mapview's parent
+   * @param observer the observer that listen to the life cycle events
+   */
+  override fun registerLifecycleObserver(mapView: View, observer: MapboxLifecycleObserver) {
+    val lifecycleOwner = ViewTreeLifecycleOwner.get(mapView)
+    if (lifecycleOwner == null) {
+      Logger.e(
+        TAG,
+        """Can't get lifecycleOwner for mapview,
+          please make sure the host Activity is AppCompatActivity and the version of appcompat is 1.3.0+.
```
pengdev, line 30: why appcompat 1.3.0?

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -0,0 +1,75 @@
...
+    private const val TAG = "MapboxLifecyclePlugin"
+  }
+}
+
+/**
+ * Extension function for MapView to get the lifecycle plugin instance.
+ *
+ * @return Lifecycle plugin instance
+ */
+fun MapPluginProviderDelegate.lifecycle(): MapboxLifecyclePlugin {
```
pengdev, line 73: we should use
```
val MapPluginProviderDelegate.lifecycle: MapboxLifecyclePlugin
  get() {
```
to align with other plugins.

sdk/src/androidTest/java/com/mapbox/maps/LifecycleTest.kt
Chaoba: With Lifecycle plugin, onStart is called automatically, so the default style could be load. This test case is used to verify this process.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
Chaoba: `ViewTreeLifecycleOwner.get(mapView)` will null and we can't hook into the lifecycle events.

pengdev: This seems to be important, we should state it clearly in the readme and maybe the separate lifecycle guide.

pengdev: Can we maybe add a lint error somewhere to validate it?

Chaoba: Yeah, already added in the readme. I'm currently working on adding the lint check for lifecycle events, however not sure whether could check library version.

pengdev: It would be good to have it in the migration guide, as well as maybe a separate guide?

PR #496, opened 2021-07-13
Add showLogo and showAttributes config for snapshotter

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add showLogo and showAttributes config for snapshotter. User can then hide logo and attributions in a snapshotter</changelog>`.

### Summary of changes
Created a new data class `SnapshotOptions` which wrap `MapSnapshotOptions` and `showLogo`/`showAttribution`. 
`Snapshotter` now take a `SnapshotOptions` as construct param.
image
image
image

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/snapshotter/MapSnapshotterActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/snapshotter/MapSnapshotterActivity.kt
@@ -26,7 +26,10 @@ class MapSnapshotterActivity : AppCompatActivity(), SnapshotStyleListener {
       .pixelRatio(1.0f)
       .build()
 
-    mapSnapshotter = Snapshotter(this, snapshotterOptions).apply {
+    mapSnapshotter = Snapshotter(
+      this, snapshotterOptions,
+      SnapshotOverlayOptions(showLogo = false, showAttributes = false)
+    ).apply {
       setStyleListener(this@MapSnapshotterActivity)
       setStyleUri(Style.MAPBOX_STREETS)
       setCamera(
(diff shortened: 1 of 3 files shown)


kiryldz (commented):
@Chaoba 
1. can we have a unit test verifying that when we set new flags to false - we do not call `drawAttribution` and `drawLogo`?
2. why is it a breaking change? having default values should not make it a breaking change I assume

kiryldz:
> @Chaoba
> 
> 1. can we have a unit test verifying that when we set new flags to false - we do not call `drawAttribution` and `drawLogo`?
> 2. why is it a breaking change? having default values should not make it a breaking change I assume

I see now that it's a breaking change because of metrics CI bot. Is there perhaps a way to not make it a breaking change somehow? Although I'm fine if it is

Chaoba:
> > @Chaoba
> > 
> > 1. can we have a unit test verifying that when we set new flags to false - we do not call `drawAttribution` and `drawLogo`?
> > 2. why is it a breaking change? having default values should not make it a breaking change I assume
> 
> I see now that it's a breaking change because of metrics CI bot. Is there perhaps a way to not make it a breaking change somehow? Although I'm fine if it is

Another solution is to let the SnapshottOptions only contain `drawAttribution` and `drawLogo` and have a default value as the last param of constructor method of `Snapshotter`. @tobrun What do you think?

Chaoba:
> @Chaoba
> 
> 1. can we have a unit test verifying that when we set new flags to false - we do not call `drawAttribution` and `drawLogo`?
> 2. why is it a breaking change? having default values should not make it a breaking change I assume

It's very different to test the snapshotter callback. Guess we can add an instrumentation test when we have a pixel compare test.

tobrun:
> Another solution is to let the SnapshottOptions only contain drawAttribution and drawLogo and have a default value as the last param of constructor method of Snapshotter. @tobrun What do you think?

Maybe that is better.. and have the default values being set to true

kiryldz (approved):
LGTM (the only concern left is not having a test for that but if I got correctly only pixel-match tests could help us here)

PR #516, opened 2021-07-23
Add lint check for lifecycle methods

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add lint check for lifecycle methods</changelog>`.

### Summary of changes
This pr add the lint module which includes a rule to notify users of no need to invoke these methods if the lifecycle plugin is applied.
image


mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  companion object Issues {
+    private const val ON_START = "onStart"
+    private const val ON_STOP = "onStop"
+    private const val ON_DESTROY = "onDestroy"
+    private const val ON_LOW_MEMORY = "onLowMemory"
+    private const val ISSUE_ID = "Lifecycle"
+    private const val BRIEF_DESCRIPTION = "You have redundant methods call"
```
kiryldz, line 29:
```suggestion
    private const val BRIEF_DESCRIPTION = "Redundant method call"
```

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,46 @@
...
+    private const val REPORT_MESSAGE =
+      "No need to invoke onStart/onStop/onDestroy/onLowMemory with Mapbox Lifecycle Plugin."
```
kiryldz, line 31:
```suggestion
      "No need to invoke onStart/onStop/onDestroy/onLowMemory explicitly when Mapbox Lifecycle Plugin is added as dependency."
```

kiryldz (approved):
@Chaoba absolutely great job!
LGTM with few proposals about texting.

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
pengdev: I think we could mention that the lifecycle plugin is a transitive dependency of Maps SDK by default, and we should mention the exception when appcompact is not used or the version of appcompact is below 13.x

pengdev: By the way can we also validate the appcompact version using lint?

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.jetbrains.uast.UCallExpression
+
+/**
+ * Lint detector for lifecycle events
+ */
+class LifecycleMethodDetector : Detector(), Detector.UastScanner {
+
+  override fun getApplicableMethodNames()= listOf(ON_START, ON_STOP, ON_DESTROY, ON_LOW_MEMORY)
+
+  override fun visitMethodCall(context: JavaContext, node: UCallExpression, method: PsiMethod) {
```
pengdev, line 14: nit: formatting, have we enabled ktlint for this module?

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    private val IMPLEMENTATION = Implementation(
+      LifecycleMethodDetector::class.java, Scope.JAVA_FILE_SCOPE
+    )
+
+    @JvmField
+    val ISSUE = Issue.create(
+      ISSUE_ID, BRIEF_DESCRIPTION, REPORT_MESSAGE,
+      Category.CORRECTNESS,
+      8,
```
ank27, line 41: nit: I think assigning a name instead of number`8` for Priority will be more readable.

ank27:
LGTM 👍 
Do we also want to give the developer a quick-fix shortcut for this? if they have used lifecycle methods in multiple files, it will be easy for them to just remove those with a single shortcut.

ank27 (approved):

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
Chaoba: Guess no

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
Chaoba: I think we can't validate appcompact version here.

Chaoba:
> Do we also want to give the developer a quick-fix shortcut for this? if they have used lifecycle methods in multiple files, it will be easy for them to just remove those with a single shortcut.

Yes, we can add the quick fix in the following iterate.

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,47 @@
...
+  companion object Issues {
+    private const val ON_START = "onStart"
+    private const val ON_STOP = "onStop"
+    private const val ON_DESTROY = "onDestroy"
+    private const val ON_LOW_MEMORY = "onLowMemory"
+    private const val ISSUE_ID = "Lifecycle"
+    private const val BRIEF_DESCRIPTION = "Redundant method call"
+    private const val PRIORITY = 8
+    private const val REPORT_MESSAGE =
+      "No need to invoke onStart/onStop/onDestroy/onLowMemory explicitly when Mapbox Lifecycle Plugin is added as dependency."
```
pengdev, line 32: we should also mention the limitation of appcompact 1.3.0 as well here. Since it would be required to overwrite manually if the user uses appcompact before 1.3.0

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
pengdev: For now we don't have to validate the appcompact version, but would be really good to add some instruction here as it would be required to overwrite manually if the user uses appcompact before 1.3.0

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
pengdev:
```suggestion
      "No need to invoke onStart/onStop/onDestroy/onLowMemory explicitly when Mapbox Lifecycle Plugin is added as dependency and appcompact 1.3.0+ is used."
```

pengdev (approved):

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
```diff
@@ -0,0 +1,47 @@
...
+  companion object Issues {
+    private const val ON_START = "onStart"
+    private const val ON_STOP = "onStop"
+    private const val ON_DESTROY = "onDestroy"
+    private const val ON_LOW_MEMORY = "onLowMemory"
+    private const val ISSUE_ID = "Lifecycle"
+    private const val BRIEF_DESCRIPTION = "Redundant method call"
+    private const val PRIORITY = 8
+    private const val REPORT_MESSAGE =
+      "No need to invoke onStart/onStop/onDestroy/onLowMemory explicitly when Mapbox Lifecycle Plugin is added as dependency and appcompact 1.3.0+ is used."
```
tobrun, line 32:
```suggestion
      "It's no longer required to manually invoke onStart/onStop/onDestroy/onLowMemory MapView lifecycle methods explicitly. The Mapbox Maps SDK for Android depends on the plugin-lifecycle that leverages androidx.lifecycle to make the MapView a lifecycle aware component. This behavior is automatically enforced when appcompact 1.3.0+ is used. The plugin-lifecycle can be removed from class path to opt-out of the associated behavior."
```

tobrun (approved):
Great work! Really cool use on how a plugin classpath dependency can cleanly be "documented" through lint warning. One suggestion on the error message language. Feel free to propose alterations to my suggestion.

mapbox-lint/src/main/java/com/mapbox/maps/lint/LifecycleMethodDetector.kt
pengdev: The suggestion looks great, will commit and merge this PR since @Chaoba is on holidays.

PR #519, opened 2021-07-26
Remove plugin reflection

kiryldz:
Fixes: #524 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Get rid of reflection for creating plugins. That will lead to decreasing MapView startup time if plugins are enabled.</changelog>`.

### Summary of changes

There is no need in using reflection for creating plugin instances.
This PR will replace reflection calls with simply creating plugin instances (if those plugins are available on classpath). 

### User impact (optional)

API breaking change (related to `MapView` and `MapSurface`):
```
/**
 * Definition of a map plugin provider delegate.
 */
interface MapPluginProviderDelegate {
  /**
   * Get the plugin instance.
   *
   * @param clazz the same class type that was used when instantiating the plugin
   * @return created plugin instance
   */
  fun <T> getPlugin(clazz: Class<T>): T?

  /**
   * Get the plugin instance from name.
   *
   * @param className
   * @return created plugin instance
   */
  fun <T> getPlugin(className: String): T?
}
```
is replaced with
```
/**
 * Definition of a map plugin provider delegate.
 */
interface MapPluginProviderDelegate {
  /**
   * Get the plugin instance.
   *
   * @param id plugin id
   * @return created plugin instance or null if no plugin is found for given id.
   */
  fun <T : MapPlugin> getPlugin(id: String): T?
}
```

Default Mapbox plugin ids could be found in `Plugin.kt`.


tobrun (approved):

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -226,61 +233,72 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+          else -> {
+            plugin.pluginInstance ?: throw RuntimeException("Custom non Mapbox plugins must have non-null pluginInstance parameter!")
           }
-        } else {
-          createPlugin(mapView, pluginClass)
         }
-        if (plugin is CameraAnimationsPlugin) {
-          mapboxMap.setCameraAnimationPlugin(plugin)
+        createPlugin(mapView, Plugin.Custom(plugin.pluginId, pluginObject))
+        if (pluginObject is CameraAnimationsPluginImpl) {
```
pengdev, line 292: should be pluginObject is CameraAnimationsPlugin?

kiryldz: No, we want `MapboxMap` extension functions for our plugins only.

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
```diff
@@ -1,42 +1,196 @@
...
+   */
+  val instance: MapPlugin?
+) {
+  /**
+   * Class for Camera plugin responsible for map camera.
+   *
+   * @param pluginId unique id.
+   * @param pluginInstance instance of [CameraAnimationsPlugin]
+   */
+  class Camera(pluginId: String, pluginInstance: CameraAnimationsPlugin? = null) :
```
pengdev, line 38: Do we need to distinguish the different plugin types? Seems the type is not used since there's a different ID for each plugin. The Plugin class itself has everything we need for the plugin, no?

pengdev: And if we want to distinguish between custom plugin and Mapbox plugin, maybe we can separate it that way.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -76,10 +76,8 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
 
   @Suppress("UNCHECKED_CAST")
   private var gesturesPlugin: GesturesPlugin = delegateProvider.mapPluginProviderDelegate.getPlugin(
-    Class.forName(
-      PLUGIN_GESTURE_CLASS_NAME
-    ) as Class<GesturesPlugin>
-  )
+    MAPBOX_GESTURES_PLUGIN_ID
+  ) as GesturesPlugin?
```
tobrun, line 80: Considering we are using actual typed plugins, would it makes sense making getPlugin generic? This was we don't need to cast in these situations

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -226,61 +243,72 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
-              Pair(Float::class.java, options.mapOptions.pixelRatio)
-            )
+        val pluginObject = when (plugin.id) {
+          MAPBOX_GESTURES_PLUGIN_ID -> {
+            val attrs = options.attrs
+            if (attrs != null) {
+              GesturesPluginImpl(options.context, attrs, options.mapOptions.pixelRatio)
+            } else {
+              GesturesPluginImpl(options.context, options.mapOptions.pixelRatio)
+            }
```
pengdev, line 268: Now thinking of this part of the initialisation of GesturesPlugin, why does it need to be a separate constructor for context, attrs and pixel ratio? As far as I see, the GesturesPlugin extends ContextBinder
```
fun interface ContextBinder {
  /**
   * Bind the ViewPlugin with current map context. This will create a View that
   * will be added to the MapView.
   *
   * @param context The hosting context
   * @param attrs parent attributes
   * @param pixelRatio the pixel ratio of the device
   * @return View that will be added to the MapView
   */
  fun bind(context: Context, attrs: AttributeSet?, pixelRatio: Float)
}
```
So the GesturesPlugin can be initialised properly when the plugin is created through the MapPluginRegistry.

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
kiryldz: It was my initial code but I've actually wanted to make API more typed if users would like to create implementations for our plugins. It also imho makes code clearer and gives us room to e.g. allow using one plugin of certain type (right now we are not limiting this as you asked 😉).

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: @pengdev if we think even deeper there is basically no sense in creating Mapbox plugins each time from scratch, we could create them just once and re-use in all upcoming `MapView` (we just need to make sure we `cleanup` properly in `MapView.onDestroy`.

My idea was to basically retain the existing logic and not overcomplicate this PR.
WDYT of leaving it as is here and creating a follow-up task?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: IMHO it's pretty the same in general but generic solution seems indeed a more clear one, I'll do that 👍

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
pengdev: Does it mean that we need to add a new plugin type to this class every time we come up with a new plugin?

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: > we could create them just once and re-use in all upcoming MapView (we just need to make sure we cleanup properly in MapView.onDestroy).

I don't think that's valid, different MapView have different states, we can not simply use the same instance for all the maps, let's say two MapViews in the activity or the multi display use cases.

> My idea was to basically retain the existing logic and not overcomplicate this PR.
> WDYT of leaving it as is here and creating a follow-up task?

Agree with this part, this can be tracked in a separate ticket and addressed as a follow-up.

kiryldz: Good catch about multiple map views - completely forgot about it!

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
kiryldz: Yes, is it a problem? 😄 
Add corresponding class + Mapbox id for our default impl - sounds pretty OK to me

pengdev: I don't see an issue here, it's just adding more wrapping around things, and seems not actually used in the code.

I'm fine with current setup, since eventually the functionality we expose to users should work as designed.

kiryldz: Gave it a final thought, will remove sealed class logic in favour of simplicity.

sdk-base/src/main/java/com/mapbox/maps/plugin/Plugin.kt
```diff
@@ -1,42 +1,62 @@
...
+data class Plugin(
+  /**
+   * Unique plugin id.
+   */
+  val id: String,
+  /**
+   * Plugin instance implementing [MapPlugin].
+   * Must be non-null for user-defined plugins or RuntimeException will occur when trying to initiate such plugin.
+   */
+  val instance: MapPlugin? = null
```
pengdev, line 18: Since the instance parameter is only applicable for custom plugins other than Mapbox. Does it make sense to distinguish Mapbox Plugin and Custom Plugin, like a sealed class with Custom and Mapbox?  
```
sealed class Plugin(
  /**
   * Unique plugin id.
   */
  val id: String,
  /**
   * Plugin instance implementing [MapPlugin].
   * Must be non-null for user-defined plugins or RuntimeException will occur when trying to initiate such plugin.
   */
  val instance: MapPlugin? = null) {
    class Mapbox(id: String) : Plugin(id, null)
    class Custom(id: String, instance: MapPlugin)
  }
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: https://github.com/mapbox/mapbox-maps-android/issues/522

sdk/src/main/java/com/mapbox/maps/MapInitOptions.kt
```diff
@@ -51,22 +51,25 @@ data class MapInitOptions constructor(
...
-        PLUGIN_GESTURE_CLASS_NAME,
-        PLUGIN_ATTRIBUTION_CLASS_NAME,
-        PLUGIN_LOCATION_COMPONENT_CLASS_NAME,
-        PLUGIN_SCALE_BAR_CLASS_NAME,
-        PLUGIN_MAPOVERLAY_CLASS_NAME,
-        PLUGIN_ANNOTATION_CLASS_NAME,
-        PLUGIN_LIFE_CYCLE_CLASS_NAME
-      )
-    }
+    val defaultPluginList = mutableListOf(
```
pengdev, line 61: nit: should this be defined int the Plugin.kt ?

Also thinking about multiple MapView use cases, if user passes the Custom plugin with a plugin instance, it seems doesn't work well, as the plugin depends on the state of the MapView.

kiryldz: Made it immutable.
Thus it does not matter much where it is located but let's leave it here.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -30,7 +30,7 @@ import kotlin.properties.Delegates
  * However, it doesn't have to be the UI thread.
  */
 
-internal class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
+class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
```
pengdev, line 33: em.. we make this class public, because we are referencing this class in the sdk module directly I guess?

kiryldz: Yes, exactly as __all__ other plugins 😄

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: thinking again, since we always interacting with interface rather than instances, I think it's a better practise to use interface rather than implementation.

> No, we want MapboxMap extension functions for our plugins only.

I'm not sure if we want to limit this to mapbox only, and remove the flexibility to implement the CameraAnimations separately from the one we developed.

wdyt ? @tobrun

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1247,7 +1245,7 @@ class MapboxMap internal constructor(
 
   internal fun setCameraAnimationPlugin(cameraAnimationsPlugin: CameraAnimationsPlugin?) {
     cameraAnimationsPlugin?.let {
-      if (it.javaClass.canonicalName == PLUGIN_CAMERA_ANIMATIONS_CLASS_NAME)
+      if (it is CameraAnimationsPluginImpl)
```
pengdev, line 1248: same as https://github.com/mapbox/mapbox-maps-android/pull/519/files#r677527136

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: We have only one reference to the plugin currently in `MapboxMap`. I know it sounds crazy but 2 animations plugins will not work then.
I believe we could easily add ability to have multiple references to gestures / camera plugins if there would be such need but as for now I'd stick to allowing use `MapboxMap` extenstion functions only for our plugins.

kiryldz: moreover it did work that way before the change (we were checking java class name to be exactly our impl).

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: Will resolve it here, let's not duplicate discussions 😉

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: @kiryldz I didn't mean two instances of the same type 😂 I meant if user want to switch Mapbox plugin with their own plugin, for example animations, and plugins that's dependent on the animation plugin.

pengdev: 👍 let's ticket it out for further discussion, this behaviour matches the previous behaviour.

pengdev (approved):
Great work 🚀

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: it does not work as expected now as well, we will handle this issue separately from this PR

Issue #378, opened 2021-05-28
AnnotationPlugin's layers/sources are not initialized when style is loaded


LukasPaczos:
## Environment

- Maps SDK Version: 10.0.0-beta.20

## Observed behavior and steps to reproduce

When `AnnotationManager` is created _before_ the style is loaded, the `AnnotationManagerImpl#onStyleLoaded` never gest called and the sources/layer are not created which means that all future updates are no-op.

Initializing the `AnnotationManager` _after_ the style is loaded works as expected.

Similar to https://github.com/mapbox/mapbox-maps-android/issues/321.

cc @mapbox/maps-android

PR #549, opened 2021-08-09
Add check for layer and source while creating annotations and init them if not initiated before

Chaoba:
Fixes: #378 
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add check for layer and source while creating annotations and init them if not initiated before. So that enable create the AnnotationManager before loading style.</changelog>`.

### Summary of changes
If create an AnnotationManager before loading a style, the layer and source will not be initiated because while loading a style, the previously added styleload listener will be removed. To fix this, this pr add a check in `updateSource` method and init layer and source if they are not initiated before

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -333,6 +333,9 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
       Logger.e(TAG, "Can't update source: style is not fully loaded.")
       return
     }
+    if (source == null || layer == null) {
+      initLayerAndSource(style)
+    }
     source?.let { geoJsonSource ->
       if (!style.styleSourceExists(geoJsonSource.sourceId)) {
         Logger.e(TAG, "Can't update source: source has not been added to style.")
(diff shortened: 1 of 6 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -333,6 +333,9 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
       Logger.e(TAG, "Can't update source: style is not fully loaded.")
       return
     }
+    if (source == null || layer == null) {
+      initLayerAndSource(style)
+    }
```
kiryldz, line 338: Guess we need to have a test for that now.

Chaoba: It's difficult to test it with a test case before we have pixel compare test, only can be done in the example.

kiryldz: Can't we have a unit test here somehow to verify that `initLayerAndSource` was called?

Chaoba: Rethink about it, we can verify it by verifying `style.addPersistentLayer`, will add a test case later.

kiryldz: I guess this PR is the best place to add the test (test that would have failed without changes that this PR is brining and which should start passing now).

plugin-annotation/src/test/java/com/mapbox/maps/plugin/annotation/generated/CircleAnnotationManagerTest.kt
```diff
@@ -168,6 +167,27 @@ class CircleAnnotationManagerTest {
     assertTrue(manager.longClickListeners.isEmpty())
   }
 
+  @Test
+  fun initializeBeforeStyleLoad() {
+    every { style.styleLayerExists("test_layer") } returns true
+    val captureCallback = slot<(StyleInterface) -> Unit>()
+    every { delegateProvider.getStyle(capture(captureCallback)) } answers {
```
kiryldz, line 174: can we use `just Runs` here?

kiryldz (approved):

PR #558, opened 2021-08-11
Introduce coroutine to GeoJsonSource.kt

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce coroutine to GeoJsonSource to remove static worker parsing thread that can't be precisely controlled.</changelog>`.

### Summary of changes
Use coroutine to parse GeoJson data instead of HandlerThread

### User impact (optional)
Following code could be removed from tests:
```
    GeoJsonSource.workerThread =
      HandlerThread("STYLE_WORKER").apply {
        priority = Thread.MAX_PRIORITY
        start()
      }
```

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/CircleAnnotationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/CircleAnnotationActivity.kt
@@ -69,7 +69,7 @@ class CircleAnnotationActivity : AppCompatActivity() {
 
         // random add circles across the globe
         val circleAnnotationOptionsList: MutableList<CircleAnnotationOptions> = ArrayList()
-        for (i in 0..2000) {
+        for (i in 0..20) {
           val color = Color.argb(255, random.nextInt(256), random.nextInt(256), random.nextInt(256))
           circleAnnotationOptionsList.add(
             CircleAnnotationOptions()
--- a/extension-style/build.gradle.kts
+++ b/extension-style/build.gradle.kts
@@ -23,6 +23,7 @@ dependencies {
   implementation(Dependencies.androidxAppCompat)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
+  implementation(Dependencies.coroutines)
   testImplementation(Dependencies.junit)
   testImplementation(Dependencies.mockk)
   testImplementation(Dependencies.androidxTestCore)
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -408,7 +408,6 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
 
     private var rawGeoJson: GeoJson? = null
     internal val properties = HashMap<String, PropertyValue<*>>()
-
     // Properties that only settable after the source is added to the style.
     internal val volatileProperties = HashMap<String, PropertyValue<*>>()
 
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -30,6 +30,9 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
   private val onGeoJsonParsedListenerList = mutableListOf<OnGeoJsonParsed>()
   private val scope = MainScope()
   private var pendingData: Pair<GeoJson, ((GeoJsonSource) -> Unit)>? = null
+  // Use a shared single thread to get a better performance for multi sources.
+  @ObsoleteCoroutinesApi
+  private val dispatcher = newSingleThreadContext("GeoJsonSourceParseThread")
 
   private constructor(
     builder: Builder,
@@ -62,7 +65,7 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
   }
 
   private suspend fun getPropertyValue(it: GeoJson) =
-    withContext(Dispatchers.Default) {
+    withContext(dispatcher) {
       it.toPropertyValue()
     }
 
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -30,9 +30,6 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
   private val onGeoJsonParsedListenerList = mutableListOf<OnGeoJsonParsed>()
   private val scope = MainScope()
   private var pendingData: Pair<GeoJson, ((GeoJsonSource) -> Unit)>? = null
-  // Use a shared single thread to get a better performance for multi sources.
-  @ObsoleteCoroutinesApi
-  private val dispatcher = newSingleThreadContext("GeoJsonSourceParseThread")
 
   private constructor(
     builder: Builder,
@@ -656,6 +653,10 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    * Static variables and methods.
    */
   companion object {
+    // Use a shared single thread to get a better performance for multi sources.
+    @ObsoleteCoroutinesApi
+    private val dispatcher = newSingleThreadContext("GeoJsonSourceParseThread")
+
     /** Registry to control if we need to ignore parsed result for given sourceId. */
     private val ignoreParsedGeoJsonRegistry = hashMapOf<String, Boolean>()
 
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -2,10 +2,6 @@
 
 package com.mapbox.maps.extension.style.sources.generated
 
-import android.os.Handler
-import android.os.HandlerThread
-import android.os.Looper
-import androidx.annotation.VisibleForTesting
 import com.mapbox.bindgen.Value
 import com.mapbox.geojson.Feature
 import com.mapbox.geojson.FeatureCollection
@@ -20,6 +16,7 @@ import com.mapbox.maps.extension.style.types.SourceDsl
 import com.mapbox.maps.extension.style.utils.TypeUtils
 import com.mapbox.maps.extension.style.utils.silentUnwrap
 import com.mapbox.maps.extension.style.utils.toValue
+import kotlinx.coroutines.*
 
 /**
  * A GeoJSON data source.
@@ -28,11 +25,12 @@ import com.mapbox.maps.extension.style.utils.toValue
  *
  */
 class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+  private var parsingJob: Job? = null
   private var geoJsonParsed = false
   private val onGeoJsonParsedListenerList = mutableListOf<OnGeoJsonParsed>()
-  private val workerHandler by lazy {
-    Handler(workerThread.looper)
-  }
+  private val scope = MainScope()
+  private var waitingData: Pair<GeoJson, ((GeoJsonSource) -> Unit)>? = null
+
   private constructor(
     builder: Builder,
     rawGeoJson: GeoJson?,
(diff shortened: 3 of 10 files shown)


extension-style/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ dependencies {
   implementation(Dependencies.androidxAppCompat)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
+  implementation(Dependencies.coroutines)
```
kiryldz, line 26: Just curious - how does this affect binary size?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -594,27 +597,47 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-      // remove any events from queue before posting this task
-      workerHandler.removeCallbacksAndMessages(null)
-      workerHandler.post {
-        val property = data.toPropertyValue()
-        mainHandler.post {
-          // we set parsed data when sync setter was not called during background work
-          if (ignoreParsedGeoJsonRegistry[sourceId] == false) {
-            setProperty(property, throwRuntimeException = false)
-            listener.invoke(this)
+      // No parseJob is working now, parse the data directly.
```
kiryldz, line 600: In other places `workerHandler.removeCallbacksAndMessages(null)` was replace by `waitingData = null` but in this case some another logic appears.

Chaoba: HandlerThread has a queue and could pop the events automatically, we don't have the queue and need to handle the logic mannually.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -41,22 +39,27 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
           }
         }
       }
-    } ?: run { geoJsonParsed = true }
+    } ?: run {
+      geoJsonParsed = true
+    }
   }
 
+  private suspend fun getPropertyValue(it: GeoJson) =
```
kiryldz, line 58: can we replace it with extension function for `GeoJson` and have another function name similar to `parseData`.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: my high level q is more about that this manual logic is used only here while in code above is technically almost the same and we do not perform similar logic there.
but I did not give it a deep look - I'll check out your branch later to validate it myself but please focus on retaining logic 1 to 1 as it was before, we need to be careful not to bring in unexpected bugs 😉

extension-style/build.gradle.kts
Chaoba: The release aar size for extension-style on master is 612kb, on this branch is 613 kb.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: ` GeoJson.toPropertyValue()` is an extension function, this function is just a wrap the function call int default dispatcher. IMO, we don't need another extension function have the same functionality with ` GeoJson.toPropertyValue()`

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: `code above` here means the part in the constructor?  One instance can't be constructed twice, so the logic for that part can be simpler than this part as this function can be invoked multiple times

extension-style/build.gradle.kts
pengdev: I guess the aar doesn't contain the binary size of coroutine dependency 🙂

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -28,11 +25,12 @@ import com.mapbox.maps.extension.style.utils.toValue
...
  */
 class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+  private var parsingJob: Job? = null
   private var geoJsonParsed = false
   private val onGeoJsonParsedListenerList = mutableListOf<OnGeoJsonParsed>()
-  private val workerHandler by lazy {
-    Handler(workerThread.looper)
-  }
+  private val scope = MainScope()
+  private var waitingData: Pair<GeoJson, ((GeoJsonSource) -> Unit)>? = null
```
pengdev, line 32: nit: rename to `pendingData`?

pengdev:
The only concern for me is the binary size increase by introducing the `coroutine` dependency, otherwise, as long as the tests pass and examples work as expected. It's a 🟢 for me 🙂

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
```diff
@@ -33,6 +34,24 @@ class MapIntegrationTest {
...
+    return Point.fromLngLat(
+      random.nextDouble() * -360.0 + 180.0,
+      random.nextDouble() * -180.0 + 90.0
+    )
+  }
+
+  private fun getFeatureCollection(): FeatureCollection {
+    val pointList = mutableListOf<Feature>()
+    for (i in 0..pointCount) {
+      val point = createRandomPoint()
```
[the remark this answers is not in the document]
Chaoba: That's a good point. But in this case, we only want to verify loading large data, the value in these points actually makes no difference.

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
```diff
@@ -480,6 +490,105 @@ class MapIntegrationTest {
...
+          .build()
+      )
+      mapboxMap.loadStyleUri(Style.MAPBOX_STREETS) { style ->
+        val source = geoJsonSource("source") {
+          data("")
+        }
+        // load a large feature collection
+        source.featureCollection(getFeatureCollection())
+        // change geometry immediately
+        source.geometry(Point.fromLngLat(0.0, 0.0)) {}
```
kiryldz, line 561: Comment seems to be confusing as we're applying not immediately here but in async manner after large `featureCollection` applied in a sync manner.

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
```diff
@@ -480,6 +490,105 @@ class MapIntegrationTest {
...
+
+        style.addSource(source)
+        style.addLayer(
+          circleLayer("layer", "source") {
+            circleColor("red")
+            circleRadius(10.0)
+          }
+        )
+        val min = ScreenCoordinate(mapView.width / 2.0 - 10, mapView.height / 2.0 - 10)
+        val max = ScreenCoordinate(mapView.width / 2.0 + 10, mapView.height / 2.0 + 10)
```
kiryldz, line 571: Why do we need a screen box here?
If I understand test correctly we should have exactly one point (0.0, 0.0) as a result

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -41,22 +39,33 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+          onGeoJsonParsedListenerList.forEach {
+            it.onGeoJsonParsed(this@GeoJsonSource)
           }
         }
+        pendingData?.let {
+          // have data waiting to parse, continue parse this data
+          parseData(it.first, it.second)
+          // clear waitingData, so that if new data coming could wait.
+          pendingData = null
+        }
```
kiryldz, line 57: This code is duplicated below (except notifying listeners). However it seems very important to me. Can we introduce new private function and re-use it?

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
Chaoba: This case failed on firebase in the past, but should be ok now, will change it back.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/CircleAnnotationActivity.kt
```diff
@@ -69,7 +69,7 @@ class CircleAnnotationActivity : AppCompatActivity() {
 
         // random add circles across the globe
         val circleAnnotationOptionsList: MutableList<CircleAnnotationOptions> = ArrayList()
-        for (i in 0..2000) {
+        for (i in 0..20) {
```
kiryldz, line 72: curious why did we decrease number here?

Chaoba: Other examples are all 20, 2000 is introduced to test drag performance, now we don't need it anymore.

kiryldz (approved):
LGTM

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -344,27 +357,42 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-            setProperty(property, throwRuntimeException = false)
-            listener.invoke(this)
+      // No parseJob is working now, parse the data directly.
+      if (parsingJob == null) {
+        parseData(data, listener)
+      } else {
+        parsingJob?.let {
+          if (it.isActive) {
+            // still have parse job working, save the current data to waitingData.
+            pendingData = Pair(data, listener)
```
alexander-kulikovskii, line 367: we save only one data for future parsing? 
maybe we should save it in list (queue)

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -41,22 +39,37 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-    } ?: run { geoJsonParsed = true }
+    } ?: run {
+      geoJsonParsed = true
+    }
+  }
+
+  private fun parsePendingData() {
+    pendingData?.let {
+      // have data waiting to parse, continue parse this data
+      parseData(it.first, it.second)
```
alexander-kulikovskii, line 62:
```suggestion
      parseData(data = it.first, listener = it.second)
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -41,22 +39,37 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-              it.onGeoJsonParsed(this)
-            }
+      parsingJob = scope.launch {
+        val property = getPropertyValue(it)
+        geoJsonParsed = true
+        // we set parsed data when sync setter was not called during background work
+        if (ignoreParsedGeoJsonRegistry[sourceId] == false) {
+          setProperty(property, throwRuntimeException = false)
+          onGeoJsonParsedListenerList.forEach {
+            it.onGeoJsonParsed(this@GeoJsonSource)
```
alexander-kulikovskii, line 49: should we handle exceptions here? coroutine error handling works in a different way

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -41,22 +39,37 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
-        val property = it.toPropertyValue()
-        mainHandler.post {
-          geoJsonParsed = true
-          // we set parsed data when sync setter was not called during background work
-          if (ignoreParsedGeoJsonRegistry[sourceId] == false) {
-            setProperty(property, throwRuntimeException = false)
-            onGeoJsonParsedListenerList.forEach {
-              it.onGeoJsonParsed(this)
-            }
+      parsingJob = scope.launch {
```
alexander-kulikovskii, line 42: should we somehow stop (cancel) this job?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: No need to parse the pending data if have new pending data coming, we only need to parse the last one.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: There is no lifecycle for GeoJsonSouce could enable us to cancel the job at the right time.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: We don't throw an exception but will show the error log in logcat.

kiryldz (changes requested):
@Chaoba after some offline discussion with @alexander-kulikovskii some idea came to our mind. Could you please check with a profiler if we spawn an actual `GeoJsonSourceParseThread` that lives all the time (as it uses static dispatcher).

If it does then technically this PR does not differ from original one that spawns `STYLE_WORKER` handler thread but could bring in some coroutine corner cases.

Chaoba:
After offline discursion, I will close this pr for now since we don't get too many pros by introducing coroutine and many have potential bugs.

PR #569, opened 2021-08-16
Fix snapshotter leak

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix snapshotter leak. Now calling MapSnapshotter#destroy truly frees memory.</changelog>`.

### Summary of changes

Fix snapshotter leak. Now calling `MapSnapshotter#destroy` truly frees memory.


sdk/src/main/java/com/mapbox/maps/Snapshotter.kt
```diff
@@ -25,29 +23,24 @@ import kotlin.math.min
...
+  private val snapshotOverlayOptions: SnapshotOverlayOptions = SnapshotOverlayOptions()
+) {
 
-  private val context: WeakReference<Context>
-  private val coreSnapshotter: MapSnapshotterInterface
-  private val pixelRatio: Float
-  private val mapSnapshotOptions: MapSnapshotOptions
-  private val snapshotOverlayOptions: SnapshotOverlayOptions
+  private val context: WeakReference<Context> = WeakReference(context)
+  private var coreSnapshotter: MapSnapshotterInterface?
```
pengdev, line 33: can we make it a val?
```suggestion
  private val coreSnapshotter: MapSnapshotterInterface = MapSnapshotter(options)
```

kiryldz: how can we then null it in `destroy`? 😄 
and actually make this fix work

pengdev: just realised that we are setting it to null during cancel. 

btw I noticed there's a `finalize` method on MapSnapshotter class, is it something we can utilise to clean it up?

kiryldz: checked it - same leak 😆 😆 😆

pengdev: Got it, LGTM. 
One suggestion, instead of using `!!`, shall we use ?.let and throw some meaningful exception, like IllegalStateException ("snapshotter has been destroyed") etc?

pengdev (approved):
Great work! LGTM with a nit 🚀

sdk/src/main/java/com/mapbox/maps/Snapshotter.kt
pengdev: And please add tests for it as well🙂

kiryldz:
> Great work! LGTM with a nit 🚀

I think current solution is almost the same. If you've seen I did add docs to each method
```
* @throws [NullPointerException] if [destroy] was called beforehand.
```
In any case it all seems as a temp solution, I'll be creating another issue blocked upstream and adding reference in code.

PR #581, opened 2021-08-19
Revert AGP bump

tobrun:
Proposal to revert Android Gradle Plugin upgrade for now. This breaks building from source as upstream projects are still using older syntax that is not compatible with the 4.2.2 version. To re-revert this commit, we should start bumping the android tools version from top down (common -> gl-native -> mapbox-maps-android). 

@mapbox/maps-android thoughts?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -98,30 +98,6 @@ License: [The Apache Software License, Version 2.0](http://www.apache.org/licens
 
 ===========================================================================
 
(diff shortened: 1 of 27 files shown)


pengdev:
@tobrun I'm fine with reverting it and do it from top down 👍 And the version bump is not urgent compared to blocking building from source. We can easily apply the commit again when the upstream project updates to 4.2.2

kiryldz (approved):

PR #582, opened 2021-08-20
Introduce drag layer/source for annotation plugin

Chaoba:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce drag layer/source for annotation plugin to improve drag performance.</changelog>`.

### Summary of changes
In the past we will update all annotations while dragging, so the more annotations, the worse performance will be.
In this pr, a drag layer/source is introduced and will only refresh the drag annotation on this layer while dragging. This can improve drag performance dramatically.

The following is a compare test with 2000+ CircleAnnotations  

before:

https://user-images.githubusercontent.com/8577318/130190840-9b07876a-f1c0-48ae-80e5-583f8ad2e9f1.mp4

after

https://user-images.githubusercontent.com/8577318/130190854-a89298ce-b63d-4b77-a6d6-29d754885c5e.mp4

### User impact (optional)
Since we need to add the drag annotation back the original source, this will trigger a refresh for all annotations, so the drag annotation will blinks at the end of drag.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -2,6 +2,7 @@ package com.mapbox.maps.plugin.annotation
 
 import android.graphics.PointF
 import android.view.View
+import androidx.annotation.VisibleForTesting
 import com.mapbox.android.gestures.MoveGestureDetector
 import com.mapbox.common.Logger
 import com.mapbox.geojson.Feature
@@ -67,12 +68,14 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private val mapClickResolver = MapClick()
   private val mapLongClickResolver = MapLongClick()
   private val mapMoveResolver = MapMove()
-  private var draggedAnnotation: T? = null
+  private var draggingAnnotation: T? = null
   private val annotationMap = ConcurrentHashMap<Long, T>()
   protected var touchAreaShiftX: Int = mapView.scrollX
   protected var touchAreaShiftY: Int = mapView.scrollY
   protected abstract val layerId: String
   protected abstract val sourceId: String
+  protected abstract val dragLayerId: String
+  protected abstract val dragSourceId: String
 
   @Suppress("UNCHECKED_CAST")
   private var gesturesPlugin: GesturesPlugin = delegateProvider.mapPluginProviderDelegate.getPlugin(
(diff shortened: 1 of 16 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -85,9 +88,15 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   /** The layer created by this manger. Annotations will be added to this layer.*/
   internal var layer: L? = null
 
-  /** The source created by this manger. Feature data will bed added to this source.*/
+  /** The source created by this manger. Feature data will be added to this source.*/
```
kiryldz, line 91:
```suggestion
  /** The source created by this manager. Feature data will be added to this source.*/
```
here and in other places

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -277,6 +316,19 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     }
   }
 
+  /**
+   * Add an annotation to MapView.
+   *
+   * @param annotation the annotation added to MapView.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  fun addAnnotation(annotation: T) {
```
kiryldz, line 325: should be `internal` then?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -572,17 +655,28 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
       }
       return false
     }
 
     private fun stopDragging() {
-      draggedAnnotation?.let { annotation ->
+      draggingAnnotation?.let { annotation ->
         dragListeners.forEach { it.onAnnotationDragFinished(annotation) }
+        draggingAnnotation = null
+        addAnnotation(annotation)
```
kiryldz, line 674: choose better naming smth like `restoreAnnotationOnDragEnd`

kiryldz (approved):
LGTM and great work!

Dropped some cosmetic comments but I believe we can merge and iterate.

Also it seems that on drag end when adding annotation again some flickering could happen - do you have ideas if that could be fixed? Then follow-up task may useful here.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: We need this function in test, so have to keep it as public.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: This is corresponding to the upper `delete` function. The actual behavior is to add a deleted annotation, might not be constrained to drag. So IMO `addAnnotation` is a better name.

Chaoba:
> Also it seems that on drag end when adding annotation again some flickering could happen - do you have ideas if that could be fixed? Then follow-up task may useful here.

@kiryldz Yeah, that's what I have described in PR description. This is a trade-off for performance as we need to update all annotations in the end and it costs time. One solution is to keep the drag annotation in the draglayer, but it would make the logic more complex and introduce potential bugs, so I would rather keep it.

PR #584, opened 2021-08-23
fix uppercase bug

ank27:

## Pull request checklist:

### Summary of changes
Fix uppercase bug.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
@@ -14,6 +14,7 @@ import com.mapbox.maps.plugin.annotation.Annotation
 import com.mapbox.maps.plugin.annotation.AnnotationManager
 import com.mapbox.maps.plugin.annotation.AnnotationType
 import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
+import java.util.Locale
 
 /**
  * Class for PointAnnotation
@@ -84,7 +85,7 @@ class PointAnnotation(
     get() {
       val value = jsonObject.get(PointAnnotationOptions.PROPERTY_ICON_ANCHOR)
       value?.let {
-        return IconAnchor.valueOf(it.asString.uppercase())
+        return IconAnchor.valueOf(it.asString.toUpperCase(Locale.US))
       }
       return null
     }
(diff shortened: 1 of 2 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
```diff
@@ -84,7 +84,7 @@ class PointAnnotation(
     get() {
       val value = jsonObject.get(PointAnnotationOptions.PROPERTY_ICON_ANCHOR)
       value?.let {
-        return IconAnchor.valueOf(it.asString.uppercase())
+        return IconAnchor.valueOf(it.asString.toUpperCase(Locale.US))
```
kiryldz, line 87: Why `US`? 😄 
Guess `Locale.getDefault()` should fit better

kiryldz (commented):
As this bug was introduced only recently label should be `skip changelog`.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
tobrun: actually specific US locale might be better. Have seen issues in the past with not using a specific locale. 
Will try finding that issue in old the repo.

ank27:
changed the label and fix extra line after import.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
kiryldz: tbh in this current scenario almost sure it does not matter, but if default locale may cause issues, I'm totally fine with `US`.

kiryldz (approved):
Don't forget to update internal repo as well

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
tobrun: https://developer.android.com/reference/java/util/Locale.html 
> Note that there are many convenience methods that automatically use the default locale, but using them may lead to subtle bugs.

pengdev (approved):

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
ank27: I saw we have used `Locale.US` in other generated files as well, so used the same for now. but yeah we can change it with default locale everywhere in one go.

kiryldz: No need, let's stick to `US` as @tobrun mentioned above that using default may lead to bugs.

Issue #596, opened 2021-08-30
Camera twitches if moved during ongoing animation


zeac:
## Environment

- Android OS version: Android 6
- Devices affected: Emulator
- Maps SDK Version: android-v10.0.0-rc.7

## Observed behavior and steps to reproduce

Maps SDK allows to start an animation that is not canceled if user interacts with the map.
It works well on Android 7+, but there is an issue on Android 6 captured on the attached video.

https://user-images.githubusercontent.com/227665/131336601-c6ff48a6-33bb-4594-92e8-f11f94d3e24d.mp4

## Notes / preliminary analysis

The root of problem in the ValueAnimator callbacks order again.

PR #597, opened 2021-08-30
Fix camera animators ordering on Android 6 and lower

zeac:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix issue with camera animators ordering on Android 6 and lower by revisiting overall approach of applying accumulated camera changes.</changelog>`.

Fixes: #596

### Summary of changes

The root of the problem is in the order of callbacks from ValueAnimators, that is different on Android 6. This PR changes core logic of applying animators making it easier to support and delegating updating accumulated `CameraOptions` to `Handler` instead of relying on update of "oldest" animator (which as said before is not working as expected on Android 6 and below). This should not affect camera animation in any way, the only possible drawback is applying camera changes on each animator update if user has added `ValueAnimator.AnimatorUpdateListener` instead of accumulating which should be a pretty rare use-case.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -1,8 +1,11 @@
 package com.mapbox.maps.plugin.animation
 
 import android.animation.Animator
+import android.animation.AnimatorListenerAdapter
 import android.animation.AnimatorSet
 import android.animation.ValueAnimator
+import android.os.Handler
+import android.os.Looper
 import androidx.annotation.VisibleForTesting
 import androidx.annotation.VisibleForTesting.PRIVATE
 import com.mapbox.common.Logger
@@ -42,8 +45,6 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
    * Consists of set owner + animator set itself.
    */
   private var highLevelAnimatorSet: HighLevelAnimatorSet? = null
-  @VisibleForTesting(otherwise = PRIVATE)
-  internal var highLevelListener: Animator.AnimatorListener? = null
 
   private val centerListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Point>>()
   private val zoomListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Double>>()
@@ -54,6 +55,16 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
   private val pitchListeners = CopyOnWriteArraySet<CameraAnimatorChangeListener<Double>>()
 
   private val lifecycleListeners = CopyOnWriteArraySet<CameraAnimationsLifecycleListener>()
+
+  private val handler = Handler(Looper.getMainLooper())
+  private var commitScheduled = false
+  private val commitChangesRunnable = Runnable {
+    performMapJump(cameraOptionsBuilder.anchor(anchor).build())
+
+    // reset values
+    cameraOptionsBuilder = CameraOptions.Builder()
+    commitScheduled = false
+  }
 
   /**
    * If debug mode is enabled extra logs will be written about animation lifecycle and
@@ -163,6 +174,7 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     paddingListeners.clear()
     lifecycleListeners.clear()
     animators.clear()
+    handler.removeCallbacks(commitChangesRunnable)
   }
 
   private fun performMapJump(cameraOptions: CameraOptions) {
@@ -238,21 +250,6 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     }
   }
 
-  private fun correctInitialCamera(
-    currentCameraBuilder: CameraOptions.Builder,
-    cameraAnimator: CameraAnimator<*>
-  ): CameraOptions {
-    return when (cameraAnimator) {
-      is CameraCenterAnimator -> currentCameraBuilder.center(cameraAnimator.animatedValue as? Point).build()
-      is CameraZoomAnimator -> currentCameraBuilder.zoom(cameraAnimator.animatedValue as? Double).build()
-      is CameraAnchorAnimator -> currentCameraBuilder.anchor(cameraAnimator.animatedValue as? ScreenCoordinate).build()
-      is CameraPaddingAnimator -> currentCameraBuilder.padding(cameraAnimator.animatedValue as? EdgeInsets).build()
-      is CameraBearingAnimator -> currentCameraBuilder.bearing(cameraAnimator.animatedValue as? Double).build()
-      is CameraPitchAnimator -> currentCameraBuilder.pitch(cameraAnimator.animatedValue as? Double).build()
-      else -> throw RuntimeException("Unsupported animator type!")
-    }
-  }
-
   private fun registerInternalListener(animator: CameraAnimator<*>) {
     animator.addInternalListener(object : Animator.AnimatorListener {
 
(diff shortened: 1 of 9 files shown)


plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
```diff
@@ -172,24 +172,27 @@ class CameraAnimationsListenersTest {
...
     cameraAnimationsPluginImpl.registerAnimators(bearingAnimator)
 
     Shadows.shadowOf(Looper.getMainLooper()).pause()
     bearingAnimator.start()
     Shadows.shadowOf(Looper.getMainLooper()).idle()
 
     Assert.assertEquals(5, valuesList.size)
-    Assert.assertArrayEquals(intArrayOf(0, 1, 2), valuesList.slice(0..2).toIntArray())
+    Assert.assertArrayEquals(
+      arrayOf("listener1", "listener2", "camera"),
```
zeac, line 193: I hope it is fine to change the order.

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
```diff
@@ -962,13 +965,9 @@ class CameraAnimationsPluginImplTest {
...
-    // first bearing tick as first animation -
-    // bearing updated to start value and pitch is not updated
-    assertEquals(20.0, updateList[0].bearing!!, EPS)
-    assertEquals(0.0, updateList[0].pitch!!, EPS)
-    // second bearing tick as first animation -
-    // pitch is updated to start value
-    assertEquals(10.0, updateList[1].pitch!!, EPS)
+    // Both, tick and bearing are updated to the start values at first tick.
+    assertEquals(20.0, updateList.first().bearing!!, EPS)
+    assertEquals(10.0, updateList.first().pitch!!, EPS)
```
zeac, line 970: The resulting behavior seems like more correct that the previous one.

zeac: Originally, the test was added here - https://github.com/mapbox/mapbox-maps-android/pull/352.

The problem i am trying to fix is similar, this solution covers them both.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -311,6 +304,7 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
 
       private fun finishAnimation(animation: Animator?, finishStatus: AnimationFinishStatus) {
         (animation as? CameraAnimator<*>)?.apply {
+          // add current animator to queue-set if was not present
```
kiryldz, line 307: It seems we're removing animation here and not adding, comment is a bit confusing then

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -342,17 +328,9 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
-          // so we check on `started` flag to validate that high-level animation truly started
-          if (runningAnimatorsQueue.isEmpty() && highLevelAnimatorSet?.started == true) {
-            when (finishStatus) {
-              AnimationFinishStatus.CANCELED -> {
-                highLevelListener?.onAnimationCancel(animation)
-                highLevelListener?.onAnimationEnd(animation)
-              }
-              AnimationFinishStatus.ENDED -> highLevelListener?.onAnimationEnd(animation)
-            }
+          if (runningAnimatorsQueue.isEmpty()) {
```
kiryldz, line 331: Could be collapsed with simialar `if` above?
Which is doing `mapTransformDelegate.setUserAnimationInProgress(false)`

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -342,17 +328,9 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+            if (commitScheduled) handler.removeCallbacks(commitChangesRunnable)
```
kiryldz, line 332: nit: let's use `{}` even for single line and separate with new lines

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -364,43 +342,28 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
-          cameraOptionsBuilder.anchor(anchor).build()
-        }
-        // do not perform jump if update is triggered for not first (oldest) animator
-        else -> {
-          null
+
+      if (animator.haveUserListeners) {
+        // If the animator have third-party listeners camera changes must be applied immediately
+        // to be seen from the listeners.
+        if (commitScheduled) handler.removeCallbacks(commitChangesRunnable)
```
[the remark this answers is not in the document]
kiryldz: Generally I'm seeing 
```
if (commitScheduled) handler.removeCallbacks(commitChangesRunnable)
commitChangesRunnable.run()
```

for 3 times, let's move it to private function

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -857,16 +820,17 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
-          null
-        } else {
-          it
-        }
+        addListener(object : AnimatorListenerAdapter() {
+          override fun onAnimationEnd(animation: Animator?) {
+            if (commitScheduled) handler.removeCallbacks(commitChangesRunnable)
+            commitChangesRunnable.run()
+
+            if (highLevelAnimatorSet?.animatorSet === animation) {
```
kiryldz, line 828: Can't get why we need it, can you explain a bit?

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
kiryldz: As discussed in person it's not OK and it's fixed now.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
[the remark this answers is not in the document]
kiryldz: @zeac so do I get it correctly that accumulating camera optimization is not working if we do have `userUpdateListeners`? Could there be a better way?

thanks to @LukasPaczos for stating it out.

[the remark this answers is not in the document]
zeac: Accumulated camera changes is flushed after value updates from Animators with third-party listeners only, so this optimization still works for NavigationCamera or GesturePlugin animations.

The only other way i see is to call all `userUpdateListeners` manually after camera is updated. I'm not sure that it is a better way.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
zeac: It seems possible that the `onAnimationEnd` call for the canceled high level animation will occur after the new high level animation is started. To make sure `highLevelAnimatorSet` is valid it should be cleared only when the exact same animation is complete.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
zeac: It will change the order for some callbacks, i'm not sure about it.

kiryldz (approved):
@zeac thanks for your fantastic work in preparing this contribution!
LGTM

PR #604, opened 2021-09-01
Bump gl-native to 10.0.0-rc.8, common to 18.0.0

Chaoba:

## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Bump gl-native to 10.0.0-rc.8, common to 18.0.0</changelog>`.

## Breaking changes ⚠️
* In offline mode (set by either `mapbox::common::OfflineSwitch` API or on platform side), the error notifications are send if the required resources are not present locally. The volatile tiles are not considered to be required in offline. 
* Adapt setBounds to gl-js behavior: constraining of coordinates and zoom level is now stricter to prevent out of bounds map area to be visible in the viewport 
* Add HTTP interceptor API - for anyone who is using HttpServiceInterface; there is a new method called setInterceptor that should be overridden
* Improve collision detection by using runtime calculated sizes for collision boxes. Previously collision boxes' sizes are constant, they are calculated during symbol layout time by using constant zoom level 
* Improve collision detection by using runtime calculated pixelated sizes for collision circles. Previously collision circles' sizes are constant, they are calculated during symbol layout time by using constant zoom level  
* Implement 'promoteId' feature for geojson and vector sources. The feature allows to promote feature's property to a feature id, so that promoted id can be used with FeatureState API. 
* Enable instant  transitions for data-driven paint layer properties 

## Bug fixes 🐞
* Fix volatile tiles disappearing on "not modified" response 
* Prioritize addition of a persistent layer whose id is used for other persistent layer positions 
* Only do line breaking process for point placement labels. And if text-max-width is 0, still do general ideographic beaking checks for point labels. 
* Fix collision box's 'dynamicVerticesExt' updating in placement stage 
* Trigger map redraw when feature state changes 
* Print nothing when DisableAll set for log level


tobrun (approved):
Could you explain why the code changes were required?

pengdev (approved):
Shall we bring gl-native and common change log together in this PR?

Chaoba:
@tobrun For offline part because gl-native changed the logic to send error notification; for Point init values is because we add value range for latitude and longitude.
@pengdev Add change log to pr description

app/src/main/java/com/mapbox/maps/testapp/examples/customlayer/CustomLayerActivity.kt
```diff
@@ -29,7 +29,7 @@ class CustomLayerActivity : AppCompatActivity() {
       Style.MAPBOX_STREETS
     ) {
       mapboxMap.setCamera(
-        CameraOptions.Builder().center(Point.fromLngLat(-243.60947, 39.91448)).zoom(10.0).build()
+        CameraOptions.Builder().center(Point.fromLngLat(-63.60947, 39.91448)).zoom(10.0).build()
```
kiryldz, line 32: It should be just `63` I assume

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -419,9 +422,11 @@ class OfflineTest {
     }
     switchAirplaneMode()
     verifyAirplaneModeEnabled()
+    // Show map with the zoom level larger than download range to trigger loading error
     prepareMapView(
       observer,
-      listOf(MapEvents.MAP_IDLE, MapEvents.MAP_LOADING_ERROR)
+      listOf(MapEvents.MAP_IDLE, MapEvents.MAP_LOADING_ERROR),
+      ZOOM + 3
```
kiryldz, line 429: Why is it 3 here? 1 in some places and 2 in some other places.
Perhaps we could use one number and move it to const.

Chaoba: The download rang is [zoom - 2, zoom + 2], use zoom + 3 here to trigger loading error.

app/src/main/java/com/mapbox/maps/testapp/examples/customlayer/CustomLayerActivity.kt
Chaoba: Should be 116.39053

kiryldz (approved):

ank27 (approved):

Issue #542, opened 2021-08-03
Annotations do not show when 3D Terrain added


moritzzzzz:
## Environment

- Android OS version: Tested on API 29, 30
- Devices affected: 
- Maps SDK Version: tested on com.mapbox.maps:android:10.0.0-rc.3 - com.mapbox.maps:android:10.0.0-rc.5

## Observed behavior and steps to reproduce

Annotations (PointAnnotation, CircleAnnotation, PolyLineAnnotation) do not show when there is 3D Terrain added to the style.

```
mapboxMap.loadStyle(
            styleExtension = style(Style.SATELLITE_STREETS){
                +rasterDemSource(SOURCE) {
                    url(TERRAIN_URL_TILE_RESOURCE)
                    tileSize(512)
                }

            }

        ) 
```

To reproduce, add 3D Terrain as described above and try to create an annotation. (PointAnnotation, PolyLineAnnotation, CircleAnnotation) by:

```
val annotationApi = mapView?.annotations
circleAnnotationManager = annotationApi?.createCircleAnnotationManager(mapView)

val circleAnnotationOptions: CircleAnnotationOptions = CircleAnnotationOptions()
            // Define a geographic coordinate.
            .withPoint(point)
            // Style the circle that will be added to the map.
            .withCircleRadius(5.0)
            .withCircleColor("#ff0000")
            .withCircleStrokeWidth(1.0)
            .withCircleStrokeColor("#ffffff")

// Add the resulting circle to the map.
        circleAnnotationManager?.create(circleAnnotationOptions)
```

When excluding the `+rasterDemSource()` part, annotations show.

Please see the attached sample project that displays the behaviour on an emualted device on API 29 with `com.mapbox.maps:android:10.0.0-rc.5`.

Please note that the example project currently is on `com.mapbox.maps:android:10.0.0-rc.3` which does not show the behaviour on this device, but it shows this behaviour on other physical devices.

## Expected behavior

Annotations should be visible.

## Notes / preliminary analysis

This is related to the 3D Terrain. 

## Additional links and references

Example project that displays this behaviour:

[AnnotationBug3DTerrain.zip](https://github.com/mapbox/mapbox-maps-android/files/6926606/AnnotationBug3DTerrain.zip)

**Steps to run this:**
- add you access token to strings.xml
- run the project
- click somewhere, click again --> a line should be drawn
- change Maps SDK version to rc5
- run the project
- click somewhere
- click again --> nothing is drawn

[...]

PR #630, opened 2021-09-09
No need to check style load state for annotation plugin

Chaoba:
Fixes: #542

## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>o need to check style load state for annotation plugin</changelog>`.

### Summary of changes
In the past, we can only get the style while it is fully loaded.  So if the style is loading, no annotation will be added to map before style load is complete.
But annotation plugin only needs the style, no matter the style is fully loaded or not. This pr add a default param `checkloaded` for `getStyle` API and let the annotation plugin don't check style load state.


Chaoba:
We still need the style fully loaded, otherwise, annotation layer and source may not be added to the style while manipulating annotations and crashes will occur.

Issue #587, opened 2021-08-24
Can't update PointAnnotation's bitmap


akirmse:
If you do this:

```
// This doesn't work
options.withIconImage(bitmap1)
val annotation = pointAnnotationManager.create(options)
annotation.iconImageBitmap = bitmap2
pointAnnotationManager.update(annotation)
```

you will not see bitmap2.  In order to see the change, you also need to update the image string to something with a magic prefix:

```
// This works
options.withIconImage(bitmap1)
val annotation = pointAnnotationManager.create(options)
annotation.iconImageBitmap = bitmap2
annotation.iconImage = "icon_default_name_" + name_of_bitmap
pointAnnotationManager.update(annotation)
```

In the Google SDK, this works by itself:
```
marker.setIcon(BitmapDescriptorFactory.fromBitmap(bitmap))
```

akirmse:
I'm not sure that this is the correct fix.  Now when I change an annotation's bitmap in rc8 using the code I initially posted *all* annotations sharing the original bitmap are changed.

PR #633, opened 2021-09-10
Fix point annotation update bitmap affect others issue.

Chaoba:
Fixes: #587 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix point annotation update bitmap affect others issue.</changelog>`.

### Summary of changes
This pr changes the way to generate `iconImage` value for PointAnnotation and update the check condition for updating bitmap.

https://user-images.githubusercontent.com/8577318/132826011-d2a99d8f-7597-4224-a263-48df7efe6795.mp4

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
@@ -65,8 +65,9 @@ class PointAnnotation(
     set(value) {
       value?.let {
         field = it
-        if (iconImage == null) {
-          iconImage = ICON_DEFAULT_NAME_PREFIX + it.generationId
+        if (iconImage == null || iconImage!!.startsWith(ICON_DEFAULT_NAME_PREFIX)) {
+          // User does not set iconImage, update iconImage to this new bitmap
+          iconImage = ICON_DEFAULT_NAME_PREFIX + it.hashCode()
         }
       }
     }
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PointAnnotationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/PointAnnotationActivity.kt
@@ -1,6 +1,7 @@
 package com.mapbox.maps.testapp.examples.markersandcallouts
 
 import android.animation.ValueAnimator
+import android.graphics.Bitmap
 import android.graphics.Color
 import android.os.Bundle
 import android.view.Menu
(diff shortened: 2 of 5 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
```diff
@@ -65,8 +65,9 @@ class PointAnnotation(
     set(value) {
       value?.let {
         field = it
-        if (iconImage == null) {
-          iconImage = ICON_DEFAULT_NAME_PREFIX + it.generationId
+        if (iconImage == null || iconImage!!.startsWith(ICON_DEFAULT_NAME_PREFIX)) {
+          // User does not set iconImage, update iconImage to this new bitmap
+          iconImage = ICON_DEFAULT_NAME_PREFIX + it.hashCode()
```
kiryldz, line 70: Shouldn't we revisit how `generationId` is generated instead of not using it here and introducing standard hash function?

Chaoba: From the official docs: `generationId` changes whenever the bitmap is modified. This can be used as an efficient way to check if a bitmap has changed. Here we want to distinguish different bitmap, so use `hashCode` would be appropriate.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/generated/PointAnnotation.kt
```diff
@@ -65,8 +65,9 @@ class PointAnnotation(
     set(value) {
       value?.let {
         field = it
-        if (iconImage == null) {
-          iconImage = ICON_DEFAULT_NAME_PREFIX + it.generationId
+        if (iconImage == null || iconImage!!.startsWith(ICON_DEFAULT_NAME_PREFIX)) {
```
kiryldz, line 68: Why was second condition added?

Chaoba: If users set `iconImageBitmap` before, `iconImage` would be set to a name start with `ICON_DEFAULT_NAME_PREFIX`. And then users set a new `iconImageBitmap`, we need to update `iconImage` to the new bitmap, otherwise, it will still show the previous bitmap.

kiryldz: I see

kiryldz (approved):

alexander-kulikovskii (approved):

pengdev (approved):

PR #636, opened 2021-09-13
Add support for promoteId

ank27:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add support for PromoteId to be used with Feature state API.</changelog>`.

### Summary of changes

This PR adds support for `promoteId` which allows a developer to promote a particular string / source-property to be the feature state id in features.

'promoteId' is a feature for geojson and vector sources. The feature allows to promote feature's property to a feature id, so that promoted id can be used with FeatureState API.

### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/VectorSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/VectorSource.kt
@@ -6,6 +6,7 @@ import com.mapbox.maps.StyleManager
 import com.mapbox.maps.extension.style.layers.properties.PropertyValue
 import com.mapbox.maps.extension.style.sources.Source
 import com.mapbox.maps.extension.style.sources.TileSet
+import com.mapbox.maps.extension.style.types.PromoteId
 import com.mapbox.maps.extension.style.types.SourceDsl
 import com.mapbox.maps.extension.style.utils.TypeUtils
 import com.mapbox.maps.extension.style.utils.silentUnwrap
@@ -144,6 +145,25 @@ class VectorSource(builder: Builder) : Source(builder.sourceId) {
      * @return String
      */
     get() = getPropertyValue("attribution")
+
+  /**
+   * A property to use as a feature id (for feature state). Either a property name, or
+   * an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile
+   * source, the same property is used across all its source layers.
+   */
+  val promoteId: PromoteId?
+    /**
+     * Get the PromoteId property
+     *
+     * @return PromoteId
+     */
+    get() {
+      val propertyValue = getPropertyValue<Any>("promoteId")
+      propertyValue?.let {
+        return PromoteId.fromProperty(it)
+      }
+      return null
+    }
 
   /**
    * A setting to determine whether a source's tiles are cached locally.
@@ -302,6 +322,16 @@ class VectorSource(builder: Builder) : Source(builder.sourceId) {
     }
 
     /**
+     * A property to use as a feature id (for feature state). Either a property name, or
+     * an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile
+     * source, the same property is used across all its source layers.
+     */
+    fun promoteId(value: PromoteId) = apply {
+      val propertyValue = PropertyValue("promoteId", value.toValue())
+      properties[propertyValue.propertyName] = propertyValue
+    }
+
+    /**
      * A setting to determine whether a source's tiles are cached locally.
      */
     fun volatile(value: Boolean = false) = apply {
@@ -417,6 +447,25 @@ class VectorSource(builder: Builder) : Source(builder.sourceId) {
        * @return Long
        */
       get() = StyleManager.getStyleSourcePropertyDefaultValue("vector", "maxzoom").silentUnwrap()
+
+    /**
+     * A property to use as a feature id (for feature state). Either a property name, or
+     * an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile
+     * source, the same property is used across all its source layers.
+     */
+    val defaultPromoteId: PromoteId?
+      /**
+       * Get the PromoteId property
+       *
+       * @return PromoteId
+       */
+      get() {
+        val propertyValue = StyleManager.getStyleSourcePropertyDefaultValue("vector", "promoteId").silentUnwrap<Any>()
+        propertyValue?.let {
+          return PromoteId.fromProperty(it)
+        }
+        return null
+      }
 
     /**
      * A setting to determine whether a source's tiles are cached locally.
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
@@ -1,0 +1,61 @@
+package com.mapbox.maps.extension.style.types
+
+import androidx.annotation.Keep
+import com.mapbox.bindgen.Value
+import com.mapbox.maps.extension.style.sources.generated.GeoJsonSource
+import com.mapbox.maps.extension.style.sources.generated.VectorSource
+import java.lang.UnsupportedOperationException
+
+/**
+ * Holds a property type to promote a specific feature for feature state API.
+ *
+ * @param sourceId source layer id of the feature, either source [GeoJsonSource] or [VectorSource].
+ * @param propertyName feature property name.
+ *
+ * For more information see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
+ */
+@Keep
+data class PromoteId @JvmOverloads constructor(
+  val propertyName: String,
+  val sourceId: String? = null
+) {
+  internal fun toValue(): Value {
+    sourceId?.let {
+      return Value(hashMapOf(sourceId to Value(propertyName)))
+    }
+    return Value(propertyName)
+  }
+
+  /**
+   * Static variables and methods.
+   */
+  companion object {
+    /**
+     * Construct a [PromoteId] object from a Property returned from the core.
+     *
+     * @param propertyName feature property name.
+     * Can be either of type [String] or [HashMap] of <String,String>
+     * Throws [RuntimeException] exception if couldn't construct to [PromoteId].
+     */
+    internal fun fromProperty(propertyName: Any) = when (propertyName) {
+      is String -> {
+        PromoteId(propertyName)
+      }
+      is HashMap<*, *> -> {
+        @Suppress("UNCHECKED_CAST")
+        try {
+          val propertyMap = propertyName as HashMap<String, String>
+          if (!propertyMap.keys.isNullOrEmpty()) {
+            val key = propertyMap.keys.iterator().next()
+            PromoteId(propertyMap[key] ?: "", key)
+          } else {
+            PromoteId("")
+          }
+        } catch (e: RuntimeException) {
+          throw RuntimeException("$propertyName must be in the format HashMap<String,String>")
+        }
+      }
+      else -> throw UnsupportedOperationException("Wrapping ${propertyName::class.java.simpleName} to PromoteId is not supported.")
+    }
+  }
+}
--- a/extension-style/src/test/java/com/mapbox/maps/extension/style/utils/PromoteIdTest.kt
+++ b/extension-style/src/test/java/com/mapbox/maps/extension/style/utils/PromoteIdTest.kt
@@ -1,0 +1,64 @@
+package com.mapbox.maps.extension.style.utils
+
+import com.mapbox.maps.extension.style.types.PromoteId
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class PromoteIdTest {
+
+  @Test
+  fun fromProperty_String() {
+    val expected = PromoteId("property")
+    assertEquals(expected, PromoteId.fromProperty("property"))
+  }
+
+  @Test
+  fun fromProperty_EmptyString() {
+    val expected = PromoteId("")
+    assertEquals(expected, PromoteId.fromProperty(""))
+  }
+
+  @Test
+  fun fromProperty_HashMap() {
+    val expected = PromoteId("abc", "source")
+    val actual = PromoteId.fromProperty(hashMapOf("source" to "abc"))
+    assertEquals(expected, actual)
+  }
+
+  @Test
+  fun fromProperty_EmptyHashMap() {
+    val expected = PromoteId("")
+    val actual = PromoteId.fromProperty(hashMapOf<String, String>())
+    assertEquals(expected, actual)
+  }
+
+  @Test
+  fun fromProperty_NullSource() {
+    val expected = PromoteId("abc", null)
+    assertEquals(expected, PromoteId.fromProperty("abc"))
+  }
+
+  @Test
+  fun fromProperty_EmptyPropertyAndValidSource() {
+    val expected = PromoteId("", "source")
+    val actual = PromoteId.fromProperty(hashMapOf("source" to ""))
+    assertEquals(expected, actual)
+  }
+
+  @Test(expected = UnsupportedOperationException::class)
+  fun fromProperty_UnsupportedType() {
+    PromoteId.fromProperty(intArrayOf(1, 2, 3))
+  }
+
+  @Test(expected = UnsupportedOperationException::class)
+  fun fromProperty_UnsupportedDataClassType() {
+    PromoteId.fromProperty(MockType("source", "abc"))
+  }
+
+  @Test(expected = RuntimeException::class)
+  fun fromProperty_InvalidHashMapType() {
+    PromoteId.fromProperty(hashMapOf(1 to 2))
+  }
+
+  private data class MockType(val a: String, val b: String)
+}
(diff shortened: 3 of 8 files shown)


extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun fromProperty(propertyName: Any): PromoteId {
+      return when (propertyName) {
+        is String -> {
+          PromoteId(propertyName = propertyName)
+        }
+        is HashMap<*, *> -> {
+          @Suppress("UNCHECKED_CAST")
+          val propertyMap = propertyName as HashMap<String, String>
+          val key = propertyMap.keys.iterator().next()
+          PromoteId(key, propertyMap[key]!!)
```
Chaoba, line 42: Can we avoid using `!!` here?

Chaoba (approved):
LGTM

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+ * Holds a property type to promote a specific feature for feature state API.
+ *
+ * @param sourceId source layer id of the feature, either source [GeoJsonSource] or [VectorSource].
+ * @param propertyName feature property name.
+ *
+ * For more information see [The online documentation]https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
+ */
+@Keep
+data class PromoteId @JvmOverloads constructor(
+  val sourceId: String? = null,
```
kiryldz, line 18: if this param is optional - shouldn't it be actually second one? guess it will make usage easier

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Holds a property type to promote a specific feature for feature state API.
+ *
+ * @param sourceId source layer id of the feature, either source [GeoJsonSource] or [VectorSource].
+ * @param propertyName feature property name.
+ *
+ * For more information see [The online documentation]https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
+ */
+@Keep
```
kiryldz, line 16: Are we using keep for other similar classes? just curious about it

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+     * Can be either of type [String] or [HashMap] of <String,String>
+     */
+    fun fromProperty(propertyName: Any): PromoteId {
+      return when (propertyName) {
+        is String -> {
+          PromoteId(propertyName = propertyName)
+        }
+        is HashMap<*, *> -> {
+          @Suppress("UNCHECKED_CAST")
+          val propertyMap = propertyName as HashMap<String, String>
```
kiryldz, line 40: what if it's not `HashMap<String, String>`? Will we have runtime crash? Perhaps use try-catch here to produce better exception text

kiryldz: I'm not sure how it works, could check like:
```
is HashMap<String, String> -> {
```
instead of 
```
is HashMap<*, *> -> {
```
work here?

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+    return Value(propertyName)
+  }
+
+  companion object {
+    /**
+     * Construct a [PromoteId] object from a Property returned from the core.
+     * Can be either of type [String] or [HashMap] of <String,String>
+     */
+    fun fromProperty(propertyName: Any): PromoteId {
+      return when (propertyName) {
```
kiryldz, line 34:
```suggestion
    fun fromProperty(propertyName: Any) = when (propertyName) {
```

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+ * @param propertyName feature property name.
+ *
+ * For more information see [The online documentation]https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
+ */
+@Keep
+data class PromoteId @JvmOverloads constructor(
+  val sourceId: String? = null,
+  val propertyName: String
+) {
+  internal fun toValue(): Value {
```
kiryldz, line 21: Is it used in tests only?
if yes - do we need this function at all here? we could add it as extension function in tests or do smth similar

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.mapbox.maps.extension.style.sources.generated.GeoJsonSource
+import com.mapbox.maps.extension.style.sources.generated.VectorSource
+
+/**
+ * Holds a property type to promote a specific feature for feature state API.
+ *
+ * @param sourceId source layer id of the feature, either source [GeoJsonSource] or [VectorSource].
+ * @param propertyName feature property name.
+ *
+ * For more information see [The online documentation]https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
```
kiryldz, line 14:
```suggestion
 * For more information see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#promoteId).
```
And also not sure this will work as expected in real generated docs

kiryldz: Also link does not lead me to promote id...

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import androidx.annotation.Keep
+import com.mapbox.bindgen.Value
+import com.mapbox.maps.extension.style.sources.generated.GeoJsonSource
+import com.mapbox.maps.extension.style.sources.generated.VectorSource
+
+/**
+ * Holds a property type to promote a specific feature for feature state API.
+ *
+ * @param sourceId source layer id of the feature, either source [GeoJsonSource] or [VectorSource].
```
kiryldz, line 11: Do we have a check for that?

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+     */
+    fun fromProperty(propertyName: Any): PromoteId {
+      return when (propertyName) {
+        is String -> {
+          PromoteId(propertyName = propertyName)
+        }
+        is HashMap<*, *> -> {
+          @Suppress("UNCHECKED_CAST")
+          val propertyMap = propertyName as HashMap<String, String>
+          val key = propertyMap.keys.iterator().next()
```
kiryldz, line 41: what if we pass an empty hash map? 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,48 @@
...
+    sourceId?.let {
+      return Value(hashMapOf(sourceId to Value(propertyName)))
+    }
+    return Value(propertyName)
+  }
+
+  companion object {
+    /**
+     * Construct a [PromoteId] object from a Property returned from the core.
+     * Can be either of type [String] or [HashMap] of <String,String>
```
kiryldz, line 31: Specify `@throws` as well as `@param`

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: i did it the same way initially, but then aligned it with the docs here 
https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#vector-promoteId. 
`Either a property name, or an object of the form {<sourceLayer>: <propertyName>}`

But definitely its better to use optional param as last args.

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: Yes we are. :D

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: Using HashMap<String,String> gives compile time error.
`Cannot check for instance of erased type: kotlin.collections.HashMap<String, String> /* = java.util.HashMap<String, String>`

ank27: I will add more try{} catch{} for hashmap, but the args for this function returned from core in either string or map<string,string>.

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: Yeah, the link is broken from gl-js side, I have asked @alexshalamov if that will be fixed or i will remove from here :)

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: fair point, will add a check here. :D

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: `promoteId` is only available for GeoJson or vector source on [builder method](https://github.com/mapbox/mapbox-maps-android/pull/636/files#diff-f5e000d4ded37a7719c6f0253cb88496cf0820543ec87717a2eed34dfd169eadR586) so we don't need to check here.

But if you are asking that if we initialize promoteId with the wrong sourceId and provide that to geojson builder -> IMO it should be handled by core itself, and shouldn't be applied to feature state, but I will test it.

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
ank27: it also used in [Builder methods](https://github.com/mapbox/mapbox-maps-android/blob/fd16ddb63b0f065b178fea7fa3b681a925620c70/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt#L586-L587) of source in order to add the properties to source.

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,47 @@
...
+     */
+    fun fromProperty(propertyName: Any) = when (propertyName) {
```
alexander-kulikovskii, line 33: Could you add tests for this method and cover all scenarios?

extension-style/src/test/java/com/mapbox/maps/extension/style/utils/PromoteIdTest.kt
```diff
@@ -0,0 +1,54 @@
...
+  fun fromProperty_UnsupportedDataClassType() {
+    PromoteId.fromProperty(MockType("source", "abc"))
+  }
+
+  @Test(expected = RuntimeException::class)
+  fun fromProperty_InvalidHashMapType() {
+    PromoteId.fromProperty(hashMapOf(1 to 2))
+  }
+
+  internal data class MockType(val a: String, val b: String)
```
alexander-kulikovskii, line 53:
```suggestion
  private data class MockType(val a: String, val b: String)
```

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,55 @@
...
+            val key = propertyMap.keys.iterator().next()
+            PromoteId(propertyMap[key] ?: "", key)
+          } else {
+            PromoteId("")
+          }
+        } catch (e: RuntimeException) {
+          throw RuntimeException("$propertyName must be in the format HashMap<String,String>")
+        }
+      }
+      else -> throw RuntimeException("Wrapping ${propertyName::class.java.simpleName} to PromoteId is not supported.")
```
alexander-kulikovskii, line 52:
```suggestion
      else -> throw UnsupportedTypeException("Wrapping ${propertyName::class.java.simpleName} to PromoteId is not supported.")
```

extension-style/src/test/java/com/mapbox/maps/extension/style/utils/PromoteIdTest.kt
```diff
@@ -0,0 +1,54 @@
+package com.mapbox.maps.extension.style.utils
+
+import com.mapbox.maps.extension.style.types.PromoteId
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
alexander-kulikovskii, line 9: do really need this runner?

extension-style/src/test/java/com/mapbox/maps/extension/style/utils/PromoteIdTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+  @Test
+  fun fromProperty_EmptyString() {
+    val expected = PromoteId("")
+    assertEquals(expected, PromoteId.fromProperty(""))
+  }
+
+  @Test
+  fun fromProperty_HashMap() {
+    val expected = PromoteId("abc", "source")
```
alexander-kulikovskii, line 26: nice to add one more test `PromoteId("abc", null)`

alexander-kulikovskii: and `PromoteId("", "source")`

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
```diff
@@ -0,0 +1,56 @@
...
+            val key = propertyMap.keys.iterator().next()
+            PromoteId(propertyMap[key] ?: "", key)
+          } else {
+            PromoteId("")
+          }
+        } catch (e: RuntimeException) {
+          throw RuntimeException("$propertyName must be in the format HashMap<String,String>")
+        }
+      }
+      else -> throw UnsupportedOperationException("Wrapping ${propertyName::class.java.simpleName} to PromoteId is not supported.")
```
alexander-kulikovskii, line 53: can we somehow check this exception?

kiryldz (approved):

alexander-kulikovskii (approved):

extension-style/src/main/java/com/mapbox/maps/extension/style/types/PromoteId.kt
pengdev: I think we could split this function to two functions to make the API signature strongly typed, since we only support two types of data.  i.e. `fromProperty(propertyName: String)` and `fromProperty(propertyMap: Map<String, String>)`. 

Thinking again, why would we need to have this method? The promoteId could either a property name, or an object of the form {<sourceLayer>: <propertyName>}, so the constructor should be well enough for this purpose.

ank27: > I think we could split this function to two functions to make the API signature strongly typed, since we only support two types of data. i.e. `fromProperty(propertyName: String)` and `fromProperty(propertyMap: Map<String, String>)`.
> 
> Thinking again, why would we need to have this method? The promoteId could either a property name, or an object of the form {: }, so the constructor should be well enough for this purpose. cc @ank27

It was made this way to handle the errors in case core provides other type formats. 
hmm, I see that we can use `Value` as it is already wrapper for type.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/VectorSource.kt
```diff
@@ -301,6 +321,16 @@ class VectorSource(builder: Builder) : Source(builder.sourceId) {
       properties[propertyValue.propertyName] = propertyValue
     }
 
+    /**
+     * A property to use as a feature id (for feature state). Either a property name, or
+     * an object of the form `{<sourceLayer>: <propertyName>}`. If specified as a string for a vector tile
+     * source, the same property is used across all its source layers.
+     */
+    fun promoteId(value: PromoteId) = apply {
+      val propertyValue = PropertyValue("promoteId", TypeUtils.wrapToValue(value.toValue()))
```
pengdev, line 330: Nit: do we still need wrapToValue since toValue() should have been Value type already?

pengdev (approved):
LGTM with a nit

PR #639, opened 2021-09-14
Fix drag annotation blink issue

Chaoba:

## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix drag annotation blink issue</changelog>`.

### Summary of changes
In this pr, we will remove the drag annotation from the drag source only when the map is idle and 1 sec later, so there will be no blink issue after drag.

Before:

https://user-images.githubusercontent.com/8577318/130190854-a89298ce-b63d-4b77-a6d6-29d754885c5e.mp4

After:

https://user-images.githubusercontent.com/8577318/133201118-aa534c68-1026-412a-be75-d981fb66223b.mp4

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -1,6 +1,8 @@
 package com.mapbox.maps.plugin.annotation
 
 import android.graphics.PointF
+import android.os.Handler
+import android.os.Looper
 import android.view.View
 import androidx.annotation.VisibleForTesting
 import com.mapbox.android.gestures.MoveGestureDetector
@@ -34,10 +36,8 @@ import com.mapbox.maps.extension.style.sources.generated.geoJsonSource
 import com.mapbox.maps.plugin.InvalidPluginConfigurationException
 import com.mapbox.maps.plugin.Plugin.Companion.MAPBOX_GESTURES_PLUGIN_ID
 import com.mapbox.maps.plugin.annotation.generated.PointAnnotation
-import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
-import com.mapbox.maps.plugin.delegates.MapDelegateProvider
-import com.mapbox.maps.plugin.delegates.MapFeatureQueryDelegate
-import com.mapbox.maps.plugin.delegates.MapStyleStateDelegate
+import com.mapbox.maps.plugin.delegates.*
+import com.mapbox.maps.plugin.delegates.listeners.OnMapIdleListener
 import com.mapbox.maps.plugin.gestures.GesturesPlugin
 import com.mapbox.maps.plugin.gestures.OnMapClickListener
 import com.mapbox.maps.plugin.gestures.OnMapLongClickListener
@@ -61,6 +61,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private var mapFeatureQueryDelegate: MapFeatureQueryDelegate =
     delegateProvider.mapFeatureQueryDelegate
   private var styleStateDelegate: MapStyleStateDelegate = delegateProvider.styleStateDelegate
+  private var mapListenerDelegate: MapListenerDelegate = delegateProvider.mapListenerDelegate
   protected val dataDrivenPropertyUsageMap: MutableMap<String, Boolean> = HashMap()
   private var currentId = 0L
   private var width = 0
@@ -76,6 +77,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   protected abstract val sourceId: String
   protected abstract val dragLayerId: String
   protected abstract val dragSourceId: String
+  private val handler = Handler(Looper.getMainLooper())
 
   @Suppress("UNCHECKED_CAST")
   private var gesturesPlugin: GesturesPlugin = delegateProvider.mapPluginProviderDelegate.getPlugin(
@@ -663,15 +665,28 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
       return false
     }
 
+    private val onMapIdleListener: OnMapIdleListener = object : OnMapIdleListener {
+      override fun onMapIdle() {
+        mapListenerDelegate.removeOnMapIdleListener(this)
+        // Remove dragging annotation from drag layer
+        handler.postDelayed(
+          {
+            // Delay 1 second to avoid blink issue.
+            delegateProvider.getStyle { style ->
+              updateDragSource(style)
+            }
+          },
+          UPDATE_DELAY
+        )
+      }
+    }
+
     private fun stopDragging() {
       draggingAnnotation?.let { annotation ->
         dragListeners.forEach { it.onAnnotationDragFinished(annotation) }
         draggingAnnotation = null
+        mapListenerDelegate.addOnMapIdleListener(onMapIdleListener)
         addAnnotation(annotation)
-      }
-      // Remove dragging annotation from drag layer
-      delegateProvider.getStyle { style ->
-        updateDragSource(style)
       }
     }
   }
@@ -763,5 +778,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     private const val QUERY_WAIT_TIME = 2L
     private const val CLUSTER_TEXT_LAYER_ID = "mapbox-android-cluster-text-layer"
     private val DEFAULT_TEXT_FIELD = get("point_count")
+    /** The delay time when updating drag source */
+    private const val UPDATE_DELAY = 1000L
   }
 }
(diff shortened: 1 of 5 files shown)


plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -663,16 +664,29 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
+        mapListenerDelegate.removeOnMapIdleListener(this)
+        // Remove dragging annotation from drag layer
+        Handler(Looper.getMainLooper()).postDelayed(
+          {
+            // Delay 1 second to avoid blink issue.
+            delegateProvider.getStyle { style ->
+              updateDragSource(style)
+            }
+          },
+          1000
```
kiryldz, line 678: move to const

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -663,16 +664,29 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
       return false
     }
 
+    private val onMapIdleListener: OnMapIdleListener = object : OnMapIdleListener {
+      override fun onMapIdle() {
+        mapListenerDelegate.removeOnMapIdleListener(this)
+        // Remove dragging annotation from drag layer
+        Handler(Looper.getMainLooper()).postDelayed(
```
kiryldz, line 671: could create handler only once and store it as a val

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: and why 1000? isn't it too much?

kiryldz:
@Chaoba `Fix drag annotation blink; Apply properties to drag layer` is pretty confusing imho.
Please better describe how it was fixed stating that we introduce delay when drag is finished before removing drag source.
Also could you please clarify if all those tests updates are related to drag blink issue? if not - can we move it to a separate PR?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: Users can't notice this delay, we need this buffer to cover the update delay, if too short, the blink may come back when there are too many annotations.

kiryldz:
Also when I look at the second video when you start dragging annotation again (sec 4-5) it seems that drag start is not behaving as expected, annotation jumps to some position instead of smooth dragging.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: also what if user starts dragging again while we're waiting for one second here?
shouldn't we do `handler.removeCallbackAndMessages(null)` when drag starts?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: Well, short is a pretty common word 😄 We don't know what device / how many annotations user is adding. If it's the only way to go do we need to make that delay configurable by adding getter / setter from the user side?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: If users drag again, `draggingAnnotation` will not be null, that's simply what we do while dragging. You can see from the attached video, the time between two drags is obviously less than 1 second and it works well.

kiryldz: The problem that it does not 😄 
https://github.com/mapbox/mapbox-maps-android/pull/639#issuecomment-918855680

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: IMO, we should expose this config to users unless they have a strong request on it. Would love to get enough feedback before that step.

Chaoba:
> Also when I look at the second video when you start dragging annotation again (sec 4-5) it seems that drag start is not behaving as expected, annotation jumps to some position instead of smooth dragging.

We still need to remove the dragging annotation from original source, which will trigger the update for all annotations. We can't skip this step in the current architecture. What we currently do is making it better but not the best.

kiryldz:
> > Also when I look at the second video when you start dragging annotation again (sec 4-5) it seems that drag start is not behaving as expected, annotation jumps to some position instead of smooth dragging.
> 
> We still need to remove the dragging annotation from original source, which will trigger the update for all annotations. We can't skip this step in the current architecture. What we currently do is making it better but not the best.

Why is there no such problem when the very first dragging happens?

pengdev:
> > > Also when I look at the second video when you start dragging annotation again (sec 4-5) it seems that drag start is not behaving as expected, annotation jumps to some position instead of smooth dragging.
> > 
> > 
> > We still need to remove the dragging annotation from original source, which will trigger the update for all annotations. We can't skip this step in the current architecture. What we currently do is making it better but not the best.
> 
> Why is there no such problem when the very first dragging happens?

I took a closer look at the first video, I see the annotation being dragged also jumped at the beginning, instead of smooth transition, so in this regard I don't think the behaviour is a regression compared to main branch.

Chaoba:
> I took a closer look at the first video, I see the annotation being dragged also jumped at the beginning, instead of smooth transition, so in this regard I don't think the behaviour is a regression compared to main branch.

Exactly. We have to update all annotations in order to remove the dragging annotation from the original source at the beginning of the drag gesture. Currently we don't have a solution for it.

pengdev:
@Chaoba one thought, can we keep both layers during the transition? for example, start merging the dragged annotation to the annotation source, and listen to the MapIdle event, when the MapIdle event is triggered, then remove the dragged layer, wdyt?

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -663,16 +665,29 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
...
     }
 
+    private val onMapIdleListener: OnMapIdleListener = object : OnMapIdleListener {
+      override fun onMapIdle() {
+        mapListenerDelegate.removeOnMapIdleListener(this)
+        // Remove dragging annotation from drag layer
+        handler.postDelayed(
+          {
+            // Delay 1 second to avoid blink issue.
+            delegateProvider.getStyle { style ->
```
alexander-kulikovskii, line 675: can we get crash when remove (set to null) annotation manager before this postDelayed happen?
should we remove it from handler?

Chaoba: Not see a crash when test this scenario locally. But need remove it in `onDestroy`.

Chaoba:
> @Chaoba one thought, can we keep both layers during the transition? for example, start merging the dragged annotation to the annotation source, and listen to the MapIdle event, when the MapIdle event is triggered, then remove the dragged layer, wdyt?

That's what we are doing now. But we always keep two layers.

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -663,16 +666,29 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
       return false
     }
 
+    private val onMapIdleListener: OnMapIdleListener = object : OnMapIdleListener {
+      override fun onMapIdle() {
+        mapListenerDelegate.removeOnMapIdleListener(this)
+        // Remove dragging annotation from drag layer
+        handler.postDelayed(
+          {
+            // Delay 1 second to avoid blink issue.
```
kiryldz, line 675: Delay is now 3 seconds isn't it?

kiryldz (approved):

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
Chaoba: This is used to test the handler after onDestroy. Still need to keep it 1 second.

alexander-kulikovskii (approved):

Issue #634, opened 2021-09-11
MapboxMap#getStyle returns null after adding a source


mfazekas:
## Environment

- Maps SDK Version: v10.0.0-rc.8

## Observed behavior and steps to reproduce

`MapboxMap#getStyle` is `null` after adding something (style or layer)

```kotlin
map.getStyle().addSource(source)
assert(map.getStyle() != null)
map.getStyle().addLayer(layer)
```

The assertion in above code fails.

## Expected behavior

IMO `MapboxMap#getStyle` should  return the style even if we've added something, and it's not fully loaded yet. I'd expect `getStyle()` to return null initially, but one it's non-null it should remain non `null`. 

We do have interface that waits for the style to finish loading, but in many cases we can add layers/sources to the style even if something still pending on the style. 

Of course we can work around the current behaviour of `getStyle` in a lot of ways. But current behaviour is a bit surprising IMO.

## Additional links and references

See also #583

tobrun:
@Chaoba you have been looking into a similar related issue. Could you look more into this one?

Issue #583, opened 2021-08-20
Cannot access map style if source fails to load


jpayne79:
## Environment

- Android OS version:
- Devices affected:
- Maps SDK Version: 10.0.0-r.6

## Observed behavior and steps to reproduce

When adding a new source to the map style, if the source fails to load, `map.getStyle` will return null since the internal check to `style.isStyleLoaded` is false. This seems problematic as it would be ideal if we could access the style object to remove the problematic source.

Steps to repro:
 - Get `MapboxMap` and load a default style
 - Once the style is loaded, add a source (vector/raster) that will return a 400/500
 - `map.getStyle` is now null since the style isn't fully loaded due to the source error. 

## Expected behavior

Preferably we'd be able to access `map.style` regardless of whether it's fully loaded or not as callers can check it themselves. The only other workaround that is apparent to me would be to store a reference to the style after the initial load. However, I'd prefer to leave the style reference management to `MapboxMap` and only access it through the map object instead of having to store a reference on our side.

Issue #614, opened 2021-09-05
Location component location retrieval speed and reliability decreased with 10.0.0-rc.4


moritzzzzz:
## Environment

- Android OS version: tested on API 29 / 30
- Devices affected: Tested on Xiaomi Mi MIX 2s, Emulator Pixel 4 API 29
- Maps SDK Version: 10.0.0-rc.4 and above

## Observed behavior and steps to reproduce

Please find attached an example project that showcases the behaviour. I have built in a timer that will pop open a alter window showing the time elapsed since the style has loaded until the `locationComponent` found the position:

[locationComponentBug.zip](https://github.com/mapbox/mapbox-maps-android/files/7112309/locationComponentBug.zip)

Test out the change in behaviour by changing the dependency in the app level build.gradle from `    implementation 'com.mapbox.maps:android:10.0.0-rc.3'`  to `    implementation 'com.mapbox.maps:android:10.0.0-rc.4'`

## Expected behavior

The behaviour experienced in `10.0.0-rc.3` is expected.
LocationComponent should retrieve position ~50ms after style was loaded.

## Notes / preliminary analysis

When running simultaneously a normal location request using GMS services, the location is returned within milliseconds. 
When the locationComponent does not fetch a location at all, it helps to update the `lastKnownLocation` by finding the location in a different application.

--> I assume a location update from the `FusedLocationProvider` might be requested with the wrong priority. Can you please check that you are requesting with highest priority?

## Additional links and references

Please see attached project.

kiryldz:
Hey @moritzzzzz please add correct dependency in your `build.gradle` file explicitly like done [here](https://github.com/mapbox/mapbox-maps-android/blob/6aca5e170d8a077f82a851d0756f3d87484452df/app/build.gradle.kts#L89).

moritzzzzz:
hi @kiryldz  adding `implementation 'com.google.android.gms:play-services-location:18.0.0'` does not change/improve the behaviour.

moritzzzzz:
@kiryldz i made another observation:

When starting the application (see example project) the `onMapReady()` function is called, the style is loaded, the `onStyleLoaded` callback is invoked and the `onIndicatorPositionchangedListener` is attached.

However the `locationComponent` does not fetch the location and display it.

BUT when putting the application in background and returning to it, the `locationComponent` will find and the display the position.

pengdev:
[...]

PR #643, opened 2021-09-15
Fix getStyle returns null after adding a source

Chaoba:
Fixes: #634 
Fixes: https://github.com/mapbox/mapbox-maps-android/issues/583

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix getStyle returns null after adding a source</changelog>`.

### Summary of changes
`onStyleLoaded` will only be triggered once for one style, so if users add a new source to style after style loaded, `onStyleLoaded` will never be invoked. And during the loading time for this new source, `getStyle` will return null because `style.isStyleLoaded` will be false.
This pr add a property `styleLoaded` to record wheter the style is loaded or not no matter how many sources or layers are added after style loaded.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -52,6 +52,7 @@ class MapboxMap internal constructor(
   internal var isStyleLoadInitiated = false
   private val styleObserver = StyleObserver(this, nativeMapWeakRef, nativeObserver, pixelRatio)
   internal var renderHandler: Handler? = null
+  internal var styleLoaded = false
 
   /**
    * Represents current camera state.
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -52,6 +52,7 @@ class MapboxMap internal constructor(
   internal var isStyleLoadInitiated = false
   private val styleObserver = StyleObserver(this, nativeMapWeakRef, nativeObserver, pixelRatio)
   internal var renderHandler: Handler? = null
+  internal var styleLoaded = false
```
alexander-kulikovskii, line 55: should it be with private setter?
because in Java `internal` converted into `public`?

Chaoba: We need set it in test cases.

kiryldz:
@Chaoba please link an issue that gets fixed by this PR.
Also we should have a new test for it that was failing before but should pass now.

pengdev:
This PR reverts @tobrun's PR in https://github.com/mapbox/mapbox-maps-android/pull/332, where we used native `isStyleLoaded` instead of the platform side cached value. Would be good to get a clear understanding of the issue we experienced and check with gl-native team if we want to modify the behaviour of `isStyleLoaded` on the native side.

pengdev:
Could be related to https://github.com/mapbox/mapbox-maps-android/issues/614 as well

pengdev:
@Chaoba could you build a snapshot using this branch?

tobrun:
> check with gl-native team if we want to modify the behaviour of isStyleLoaded on the native side.

This was already confirmed in the annotation related ticket that it's expected to return false when something new is loading. This is conform to gl-js implemenation. For our own internal style logic, we need to roll our own setup as proposed here.

pengdev (approved):

kiryldz (approved):

Issue #649, opened 2021-09-17
Mapbox crashes on setCamera


SamuelBrucksch:
## Environment

- Android OS version: 11
- Devices affected: Pixel3
- Maps SDK Version: 10 RC8

Hi,

i get this stacktrace reported on a crash that randomly appears on android auto:
```
java.lang.Error: std::exception
        at com.mapbox.maps.CameraManager.setCamera(CameraManager.java:-2)
        at com.mapbox.maps.NativeMapImpl.setCamera(NativeMapImpl.kt:35)
        at com.mapbox.maps.MapboxMap$setCamera$1.invoke(MapboxMap.kt:280)
        at com.mapbox.maps.MapboxMap$setCamera$1.invoke(MapboxMap.kt:37)
        at com.mapbox.maps.UtilsKt.call(Utils.kt:9)
        at com.mapbox.maps.MapboxMap.setCamera(MapboxMap.kt:280)
        at com.mapbox.maps.plugin.animation.CameraAnimationsPluginImpl.performMapJump(CameraAnimationsPluginImpl.kt:185)
        at com.mapbox.maps.plugin.animation.CameraAnimationsPluginImpl.access$performMapJump(CameraAnimationsPluginImpl.kt:36)
        at com.mapbox.maps.plugin.animation.CameraAnimationsPluginImpl$commitChangesRunnable$1.run(CameraAnimationsPluginImpl.kt:62)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7664)
        at java.lang.reflect.Method.invoke(Method.java:-2)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```
Unfortunately it does not contain any information, which part of my code was executed before this. I would assume it was caused by a call of setCamera but later in the stack trace i also see the Camera Animator Plugin. So i'm not really sure where to look for.

Is there any way to find the problem here? What could typically be the problem here?

kiryldz:
Hi @SamuelBrucksch!
Could you please as maximum info as possible:
 - how often does this happen?
 - are there any preconditions, does it happen during app minimize / maximize or during regular map update with no visible preconditions?
[...]

PR #652, opened 2021-09-20
Fix set camera crash

ank27:

## Pull request checklist:

### Summary of changes
Added try-catch block in order to catch `java.lang.Error` and  other `Exceptions` that could occur during `setCamera` call with animating the camera. Reported in #649.


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -182,10 +182,22 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+      mapCameraManagerDelegate.setCamera(cameraOptions)
+      // notify listeners with actual values
+      notifyListeners(mapCameraManagerDelegate.cameraState)
+      lastCameraOptions = cameraOptions
+    } catch (e: Exception) {
+      Logger.e(
+        TAG,
+        "Exception while setting camera options : ${e.message} CameraOptions = $cameraOptions"
+      )
+    } catch (error: Error) {
```
kiryldz, line 195: can we add a unit test making sure that when exception happens we do not update  `lastCameraOptions` and do not notify listeners?

SamuelBrucksch:
Hi,

thanks for adding this fix, i'm the one who reported this via https://github.com/mapbox/mapbox-maps-android/issues/649. This might fix the crash, but we are not able to see the reson in a running app. We use Bugsnag, which helped us in getting the Error at all, as it rarely happens and we never saw it while running in the emulator.

If it is now catched, we will never see the reason and it might happen, that the map does not move anymore at all, if the subsequent calls still contain the wrong camera options.

ank27:
@SamuelBrucksch thanks for the feedback. As @kiryldz mentioned in #649 that this crash happens in native, and happens **rarely**. we are going with this solution as we saw that native code can recover from such exceptions and map can resume from the same camera state, rather than not move at all. 
But we would need further testing on this as well. Once this is merged, could you be able to test it again and provide some feedback?

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
```diff
@@ -972,6 +974,60 @@ class CameraAnimationsPluginImplTest {
     unmockkStatic(ShadowLogger::class)
   }
 
+  @Test
+  fun catchErrorOnSetCameraTest() {
```
alexander-kulikovskii, line 978: I guess that the differences between this test and `catchExceptionOnSetCameraTest` is only throwing different type of error. 
so maybe they should be merged?

SamuelBrucksch:
> But we would need further testing on this as well. Once this is merged, could you be able to test it again and provide some feedback?

Sure. I will test it.

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
ank27: Yeah, the only difference is the type of error, but let's keep it just to be sure that both cases pass.

kiryldz: guess @alexander-kulikovskii was talking more about introducing private function and reusing the code and not duplicating it.

kiryldz:
@ank27 please update changelog in PR description so that script will automatically pick up changelog entry.

ank27:
> @ank27 please update changelog in PR description so that script will automatically pick up changelog entry.

done.

kiryldz (approved):
LGTM

alexander-kulikovskii (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -177,15 +178,28 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     handler.removeCallbacks(commitChangesRunnable)
   }
 
-  private fun performMapJump(cameraOptions: CameraOptions) {
+  @VisibleForTesting(otherwise = PRIVATE)
+  fun performMapJump(cameraOptions: CameraOptions) {
```
pengdev, line 182: nit: internal fun

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -130,7 +130,8 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     }
   }
 
-  private var lastCameraOptions: CameraOptions? = null
+  @VisibleForTesting(otherwise = PRIVATE)
+  var lastCameraOptions: CameraOptions? = null
```
pengdev, line 134: nit: internal var

pengdev (approved):
LGTM % nits

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
```diff
@@ -972,6 +974,43 @@ class CameraAnimationsPluginImplTest {
...
+      .zoom(5.0).build()
+    cameraAnimationsPluginImpl.lastCameraOptions = lastCameraOptions
+    every { mapCameraManagerDelegate.setCamera(any<CameraOptions>()) } throws error
+
+    cameraAnimationsPluginImpl.performMapJump(targetCameraOptions)
+    // assert camera options did not change
+    assertNotEquals(targetCameraOptions, cameraAnimationsPluginImpl.lastCameraOptions)
+
+    // assert camera states
+    assertNotNull(cameraAnimationsPluginImpl.lastCameraOptions)
```
axti, line 1002: Instead of assert each field I think we can just assert _lastCameraOption_ and _cameraAnimationsPluginImpl.lastCameraOptions_ since CameraOptions has all checks in **equals** method

axti (approved):
LGTM

Issue #381, opened 2021-06-01
Do not apply geojson async setting data in specific scenario


kiryldz:
## Environment

- Android OS version: any
- Devices affected: any
- Maps SDK Version: v10.0.0-beta.21

## Observed behavior and steps to reproduce

As stated [here](https://github.com/mapbox/mapbox-maps-android/pull/380#issuecomment-851997325):

Imagine we're setting some data to `GeoJsonSource` using `feature`, `featureCollection` or `geometry`.
Issue that I'm noticing is wrt the cancellation policy - let's imagine these 3 calls made one by one:
1. U1 async
2. U2 sync
3. U3 async

The cancellation policy is solely based on the `ignoreParsedGeoJson` flag, which means that `U2` will set it, but `U3` will immediately unset it. This will result in this order of updates being rendered:
1. U2
2. U1
3. U3

Since there was a sync update (`U2`) scheduled _after_ an async update (`U1`), that async update should in all cases be dismissed and never rendered, but in the current setup, I see this risk of it appearing anyway.

## Expected behavior

Result should be:
1. U2
2. U3

tobrun:
landed

kiryldz:
@tobrun exactly that did not land 😄

tobrun:
![giphy (1)](https://user-images.githubusercontent.com/2151639/124832428-f4be8f00-df7c-11eb-974f-17d28d8f87ec.gif)

PR #655, opened 2021-09-22
Change the logic of handling geojson update

Chaoba:
Fixes: #381

## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Change the logic of handling geojson update</changelog>`.

### Summary of changes
Introduced a HashSet to record the data that need to update for each GeojsonSource to avoid the issue that aync data update reset `ignoreParsedGeoJsonRegistry` tag.

### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -39,14 +39,15 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
     onGeoJsonParsed: OnGeoJsonParsed
   ) : this(builder) {
     rawGeoJson?.let {
-      updateParsedGeoJsonRegistry.add(sourceId + it.hashCode())
+      val dataKey = sourceId + it.hashCode()
+      updateParsedGeoJsonRegistry.add(dataKey)
       onGeoJsonParsedListenerList.add(onGeoJsonParsed)
       workerHandler.post {
         val property = it.toPropertyValue()
         mainHandler.post {
           geoJsonParsed = true
           // we set parsed data when sync setter was not called during background work
-          if (updateParsedGeoJsonRegistry.remove(sourceId + it.hashCode())) {
+          if (updateParsedGeoJsonRegistry.remove(dataKey)) {
             setProperty(property, throwRuntimeException = false)
             onGeoJsonParsedListenerList.forEach { onGeoJsonParsed ->
               onGeoJsonParsed.onGeoJsonParsed(this)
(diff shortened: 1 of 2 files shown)


kiryldz:
@Chaoba please add a test for this scenario

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -39,17 +39,17 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
     onGeoJsonParsed: OnGeoJsonParsed
   ) : this(builder) {
     rawGeoJson?.let {
-      ignoreParsedGeoJsonRegistry[sourceId] = false
+      updateParsedGeoJsonRegistry.add(sourceId + it.hashCode())
```
alexander-kulikovskii, line 42: maybe save `sourceId + it.hashCode()` to some variable and not calculate it twice

same below

Chaoba:
@kiryldz Add a test case for this scenario. We can't verify the sync set data in this test case, however, we have other cases to verify sync set data, so IMO we can skip this part.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -39,17 +39,18 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
     onGeoJsonParsed: OnGeoJsonParsed
   ) : this(builder) {
     rawGeoJson?.let {
-      ignoreParsedGeoJsonRegistry[sourceId] = false
+      val dataKey = sourceId + it.hashCode()
```
kiryldz, line 42: why can't it be simply hash code?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -97,7 +98,8 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
    * apply when data is parsed.
    */
   fun data(value: String) = apply {
-    ignoreParsedGeoJsonRegistry[sourceId] = true
+    // Clear all asynchronous data
+    updateParsedGeoJsonRegistry.clear()
```
kiryldz, line 102: shouldn't actually it be the whole fix?
why do we need a set instead of the hash map?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: If set the same data to two different sources, they will affect each other.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
Chaoba: The set contains those data that need to be updated to source, no need additional true or false values for them. In other words, those data that are not in the set will not be updated to the source

kiryldz (changes requested):
@Chaoba this PR will become redundant when upcoming gl-native release will land as parsing geojson synchronously will be removed.
Let's leave it open for now though.

kiryldz:
not relevant as https://github.com/mapbox/mapbox-maps-android/pull/699 landed.

PR #657, opened 2021-09-22
Fix memory leak in renderer destroy

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix memory leak in renderer destroy.</changelog>`.

### Summary of changes

Fix memory leak in renderer destroy.
Previously in surface mode we did not call releasing native renderer as `Activity.onDestroy` call is happening after Android system was sending `SurfaceHolder.Callback.surfaceDestroyed` when we have already stopped render thread.

Diagram from the profiler after the fix and several opening / closing activity with `MapView`

image

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/terrain3D/SantaCatalinaActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/terrain3D/SantaCatalinaActivity.kt
@@ -132,14 +132,15 @@ class SantaCatalinaActivity : AppCompatActivity() {
         response: Response<DirectionsResponse>
       ) {
         response.body()?.let { body ->
-          body.routes()[0].geometry()?.let {
-            animateRoute(
-              LineString.fromPolyline(it, PRECISION_6)
-            )
-            return
+          if (body.routes().isNotEmpty()) {
+            body.routes()[0].geometry()?.let {
+              animateRoute(
+                LineString.fromPolyline(it, PRECISION_6)
+              )
+              return
+            }
           }
-        }
-        throw RuntimeException("Not able to retrieve a directions route")
+        } ?: throw RuntimeException("Not able to retrieve a directions route")
       }
 
       override fun onFailure(call: Call<DirectionsResponse>, t: Throwable) {}
--- a/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
@@ -31,8 +31,8 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   private val eglCore: EGLCore
 
   private val lock = ReentrantLock()
-  private val destroyCondition = lock.newCondition()
   private val createCondition = lock.newCondition()
+  private val destroyCondition = lock.newCondition()
 
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal val renderEventQueue = CopyOnWriteArrayList<Runnable>()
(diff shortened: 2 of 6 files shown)


tobrun (approved):

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -254,21 +260,15 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
         shouldExit = true
         lock.withLock {
           // TODO https://github.com/mapbox/mapbox-maps-android/issues/607
-          if (mapboxRenderer.needDestroy || mapboxRenderer is MapboxTextureViewRenderer) {
-            mapboxRenderer.onSurfaceDestroyed()
-            releaseEgl()
-            surface?.release()
+          if (mapboxRenderer is MapboxTextureViewRenderer) {
```
alexander-kulikovskii, line 263: just a question
should we check `eglPrepared` like in `destroy` method?

alexander-kulikovskii (approved):

pengdev (approved):

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: basically I'm using this flag to check if `releaseAll` was called, I think we actually need another flag for it as EGL could be released in some other cases as well and that could lead us to a trouble.

kiryldz: after given another thought let's leave as is, in this particular case current way is the correct one. Adding new flag just brings more complexity, we already have a bunch of those.

sdk/src/main/java/com/mapbox/maps/renderer/WorkerHandlerThread.kt
```diff
@@ -12,6 +12,8 @@ internal class WorkerHandlerThread {
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal var handler: Handler? = null
 
+  internal var started = false
```
alexander-kulikovskii, line 15: maybe add private set?

kiryldz: 👍

sdk/src/main/java/com/mapbox/maps/renderer/WorkerHandlerThread.kt
```diff
@@ -12,6 +12,9 @@ internal class WorkerHandlerThread {
...
```
Chaoba, line 15: Instead of recording the state, we can add a `get()` and return `handlerThread.isAlive`

kiryldz: nice idea but a really small one
if you don't mind, will leave as is for now, I have a bit of plans to revisit rendering logic a bit in future, I'll do that change in another PR

kiryldz: as was pushing some changes also applied your suggestion.

app/src/main/java/com/mapbox/maps/testapp/examples/terrain3D/SantaCatalinaActivity.kt
```diff
@@ -132,14 +132,15 @@ class SantaCatalinaActivity : AppCompatActivity() {
         response: Response<DirectionsResponse>
       ) {
         response.body()?.let { body ->
-          body.routes()[0].geometry()?.let {
-            animateRoute(
-              LineString.fromPolyline(it, PRECISION_6)
-            )
-            return
+          if (body.routes().isNotEmpty()) {
```
axti, line 135: Or just `body.routes().firstOrNull()?.geometry()?.let {`

kiryldz: I've spent too much time already waiting for CI, let's leave as is, we can revisit it later during some kind of example cleanup.

PR #664, opened 2021-09-23
Add JvmOverloads for Compass plugin

axti:

## Pull request checklist:

### Summary of changes
- Add JvmOverloads for Compass View constructor to simplify the code
- Move compass view size to dimensions.

### DIFF
--- a/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
+++ b/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
@@ -23,7 +23,8 @@ open class CompassViewImpl
 @JvmOverloads constructor(
   context: Context,
   attrs: AttributeSet? = null,
-  defStyleAttr: Int = 0): CompassView, AppCompatImageView(context, attrs, defStyleAttr) {
+  defStyleAttr: Int = 0
+) : CompassView, AppCompatImageView(context, attrs, defStyleAttr) {
 
   private lateinit var presenter: CompassPlugin
 
--- a/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
+++ b/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
@@ -11,36 +11,21 @@ import androidx.core.content.res.ResourcesCompat
 /**
  * Concrete implementation of CompassView.
  */
-open class CompassViewImpl : CompassView, AppCompatImageView {
+open class CompassViewImpl
+/**
+ * Constructor with context, attribute set and defStyleAttr.
+ *
+ * @param context
+ * @param attrs The attribution set.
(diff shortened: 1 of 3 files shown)


Chaoba:
IMO, we don't need to add `JvmOverloads` for plugin classes because we create the plugin in our SDK, users don't need to create a plugin instance.

plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
```diff
@@ -50,8 +36,8 @@ open class CompassViewImpl : CompassView, AppCompatImageView {
     )?.let { compassImage = it }
 
     // Layout params
-    val screenDensity = context.resources.displayMetrics.density
-    layoutParams = FrameLayout.LayoutParams((48 * screenDensity).toInt(), (48 * screenDensity).toInt())
+    val compassViewSize = resources.getDimensionPixelSize(R.dimen.compass_view_size)
+    layoutParams = FrameLayout.LayoutParams(compassViewSize, compassViewSize)
```
tobrun, line 40: :+1:

plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt
```diff
@@ -48,38 +49,39 @@ open class CompassViewPlugin(
...
-    )
-    fadeAnimator.addUpdateListener {
-      val value = it.animatedValue as Float
-      if (value < internalSettings.opacity) {
-        compassView.setCompassAlpha(value)
+      addUpdateListener {
+        val value = it.animatedValue as Float
+        if (value < internalSettings.opacity) {
+          compassView.setCompassAlpha(value)
+        }
```
tobrun, line 62: :broom:

tobrun (approved):

tobrun:
> IMO, we don't need to add JvmOverloads for plugin classes because we create the plugin in our SDK, users don't need to create a plugin instance.

Maybe not a pure requirement but as with all the changes made in this PR it's cleaner and also correct.

plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewImpl.kt
axti: If some of customers would like to change compass size (for example for automotive maybe it is good to have it bigger) they can override this dimension and we don't need to update or modify our code.

alexander-kulikovskii: good point
Should we think about min size? If I set size to `0` I'll not see this element. Of course it's my fault as a developer, but maybe we should print some alert in console?

kiryldz (approved):
@axti is it a breaking change? for Java users?
Please update changelog entry as well

axti:
> @axti is it a breaking change? for Java users? Please update changelog entry as well

There is no breaking changes, just cleaning and applying @JvmOverload

PR #673, opened 2021-09-24
fix getLayerAs return optional Layer type.

ank27:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Update getLayerAs function to return nullable Layer type.</changelog>`.

### Summary of changes

Currently `StyleMangerInterface.getLayerAs` has return type which is non-nullable.
It should be a nullable type.

### User impact (optional)
Breaking change to get layer from style using `getLayerAs` function.

```
fun <T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T {
  @Suppress("UNCHECKED_CAST")
  return getLayer(layerId) as T
}
```

Changed to 
```
inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
  val layer = getLayer(layerId)
  if (layer !is T) {
    Logger.e("StyleLayerPlugin", "Given layerId = $layerId is not requested type in Layer")
    return null
  }
  return layer
}
```

### DIFF
--- a/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
+++ b/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
@@ -29,13 +29,15 @@ internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: Lis
         }
         .forEach { layer ->
           val symbolLayer = style.getLayerAs<SymbolLayer>(layer.id)
-          symbolLayer.textFieldAsExpression?.let { textFieldExpression ->
-            if (BuildConfig.DEBUG) {
-              Logger.i(TAG, "Localize layer id: ${symbolLayer.layerId}")
+          symbolLayer?.let {
+            it.textFieldAsExpression?.let { textFieldExpression ->
+              if (BuildConfig.DEBUG) {
+                Logger.i(TAG, "Localize layer id: ${it.layerId}")
+              }
+              val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
+              else getLanguageNameV7(locale)
+              convertExpression(language, it, textFieldExpression)
             }
-            val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
-            else getLanguageNameV7(locale)
-            convertExpression(language, symbolLayer, textFieldExpression)
           }
         }
     }
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
@@ -40,14 +40,18 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
 }
 
 /**
- * Tries to cast the Layer to T, throws ClassCastException if it's another type.
+ * Tries to cast the Layer to T.
  *
  * @param layerId the layer id
- * @return T
+ * @return T if layer is T, otherwise null
  */
-fun <T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T {
-  @Suppress("UNCHECKED_CAST")
-  return getLayer(layerId) as T
+inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
+  val layer = getLayer(layerId)
+  if (layer !is T) {
+    Logger.e("StyleLayerPlugin", "Given layerId = $layerId is not requested type in Layer")
+    return null
+  }
+  return layer
 }
 
 /**
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
@@ -32,7 +32,7 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
       "raster" -> RasterLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
       "symbol" -> SymbolLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
       else -> {
-        Logger.e("StyleLayerPlugin", "Layer type: $type unknown.")
+        Logger.e(TAG, "Layer type: $type unknown.")
         null
       }
     }
@@ -46,9 +46,9 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
  * @return T if layer is T, otherwise null
  */
 inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
-  val layer = getLayer(layerId)
-  if (layer !is T) {
-    Logger.e("StyleLayerPlugin", "Given layerId = $layerId is not requested type in Layer")
+  val layer = getLayer(layerId) as? T
+  if (layer == null) {
+    Logger.e(TAG, "Given layerId = $layerId is not requested type in Layer")
     return null
   }
   return layer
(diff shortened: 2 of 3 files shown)


extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -40,14 +40,18 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
...
+ * Tries to cast the Layer to T.
  *
  * @param layerId the layer id
  * @return T
  */
-fun <T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T {
-  @Suppress("UNCHECKED_CAST")
-  return getLayer(layerId) as T
+inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
+  val layer = getLayer(layerId)
```
kiryldz, line 49: imho it will already crash here if layer is not `T` but I might be wrong.
perhaps:
```suggestion
  val layer = getLayer(layerId) as? T
  if (layer == null) {
    Logger.e("StyleLayerPlugin", "Given layerId = $layerId is not requested type in Layer")
    return null
  }
  return layer
```

kiryldz: but maybe it won't work my way, I did not check it 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -40,14 +40,18 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
...
- * @return T
+ * @return T if layer is T, otherwise null
  */
-fun <T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T {
-  @Suppress("UNCHECKED_CAST")
-  return getLayer(layerId) as T
+inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
+  val layer = getLayer(layerId)
+  if (layer !is T) {
+    Logger.e("StyleLayerPlugin", "Given layerId = $layerId is not requested type in Layer")
```
tobrun, line 51: nit: introduce TAG in companion object and use that as first parameter of Logger.e (I do think we had a convention of prefixing with mbgl_ but better to check the codebase).

tobrun: `Given layerId = $layerId is not requested type in Layer` the error statement is a bit confusing. 
The layer itself could also just be null

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
ank27: Yeah, it will work as well.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
ank27: Any suggestion for error message ? :D

ank27: > nit: introduce TAG in companion object and use that as first parameter of Logger.e (I do think we had a convention of prefixing with mbgl_ but better to check the codebase).

actually, this is not a class, it's an extension function file.

tobrun: Could we do something as:

```
private const val TAG = "mgbl-LayerUtils"
```
at bottom or top of file?

LayerUtils naming comes from the java name of the LayerExt file.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -145,4 +149,6 @@ fun StyleInterface.addPersistentLayer(layer: Layer, position: LayerPosition? = n
 @MapboxExperimental
 fun Layer.isPersistent(): Boolean? {
   return delegate?.isStyleLayerPersistent(layerId)?.value
-}
\ No newline at end of file
+}
+
+const val TAG = "Mbgl-LayerUtils"
```
tobrun, line 154: could be made private? We might be leaking this tag if not?

ank27: `getLayerAs` is a public inline function, which cannot use Private / internal declarations.
Although I can use @PublishedApi annotation to make TAG internal and let the function use it. but basically, that would be public TAG in a way.

ank27: Fixed by annotating TAG with @PublishedApi in order to use it in public inline function `getLayerAs`.

kiryldz:
@ank27 how does it all look from Java side?

ank27:
> @ank27 how does it all look from Java side?
we can't use `inline` functions from java code directly. For that we have `getLayer` function already. Its the same with `SourceExt` as well.

kiryldz (approved):
LGTM

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -29,13 +29,15 @@ internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: Lis
...
-              Logger.i(TAG, "Localize layer id: ${symbolLayer.layerId}")
+          symbolLayer?.let {
+            it.textFieldAsExpression?.let { textFieldExpression ->
+              if (BuildConfig.DEBUG) {
+                Logger.i(TAG, "Localize layer id: ${it.layerId}")
+              }
+              val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
+              else getLanguageNameV7(locale)
+              convertExpression(language, it, textFieldExpression)
             }
```
pengdev, line 40: nit: can be simplified
```suggestion
          symbolLayer?.textFieldAsExpression?.let { textFieldExpression ->
              if (BuildConfig.DEBUG) {
                Logger.i(TAG, "Localize layer id: ${it.layerId}")
              }
              val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
              else getLanguageNameV7(locale)
              convertExpression(language, it, textFieldExpression)
            }
```

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -32,22 +32,26 @@ fun StyleManagerInterface.getLayer(layerId: String): Layer? {
...
- * @return T
+ * @return T if layer is T, otherwise null
  */
-fun <T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T {
-  @Suppress("UNCHECKED_CAST")
-  return getLayer(layerId) as T
+inline fun <reified T : Layer> StyleManagerInterface.getLayerAs(layerId: String): T? {
+  val layer = getLayer(layerId) as? T
+  if (layer == null) {
+    Logger.e(TAG, "Given layerId = $layerId is not requested type in Layer")
```
pengdev, line 51: nit: shall we add a check to log if the layer id exists in currently style?

ank27: there is no need for an additional check, this will return `null` if the layer doesn't exist.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
ank27: do you think this will work? in `convertExpression(language, it, textFieldExpression)` `it` is basically a `symbolLayer`, which will give an unresolved reference if we use it like this. let me check that too :D

Chaoba (approved):

Issue #635, opened 2021-09-13
"Mapbox" word overlaps blue click button in RTL


akirmse:
Put the phone in RTL mode (e.g. Settings -> Developer options -> Force RTL layout direction)
The "Mapbox" map overlay overlaps the little blue button you can click on.  See the lower right corner of this screenshot.

![Screenshot_20210912-194016](https://user-images.githubusercontent.com/2266354/133016123-eb2b667a-8d92-40ba-bb56-b8d5b3ad0b30.jpg)

This is in rc8.

ZiZasaurus:
@akirmse thank you for bringing this to our attention. We are working on a fix [here](https://github.com/mapbox/mapbox-maps-android/pull/674), which should be available in our next release.

PR #674, opened 2021-09-24
Fix attribution/logo jumble when RTL layout is configured

ZiZasaurus:
Created to resolve https://github.com/mapbox/mapbox-maps-android/issues/635

Before: 
Screen Shot 2021-09-27 at 9 19 44 AM

After:
Screen Shot 2021-09-27 at 9 20 16 AM

### DIFF
--- a/plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
+++ b/plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
@@ -71,7 +71,12 @@ class AttributionViewImpl @JvmOverloads constructor(
    * @param bottom Margin to the bottom in pixel
    */
   override fun setAttributionMargins(left: Int, top: Int, right: Int, bottom: Int) {
-    (layoutParams as FrameLayout.LayoutParams).setMargins(left, top, right, bottom)
+    (layoutParams as FrameLayout.LayoutParams).apply{
+      marginStart = left
+      setTop(top)
+      marginEnd = right
+      setBottom(bottom)
+    }
   }
 
   /**
--- a/plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
+++ b/plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
@@ -71,11 +71,11 @@ class AttributionViewImpl @JvmOverloads constructor(
    * @param bottom Margin to the bottom in pixel
    */
   override fun setAttributionMargins(left: Int, top: Int, right: Int, bottom: Int) {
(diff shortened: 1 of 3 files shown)


Chaoba:
@ZiZasaurus Did you verify this fix could take effect?

tobrun (commented):
> @ZiZasaurus Did you verify this fix could take effect?

was looking together with @ZiZasaurus and it did.
It would be great though to also include before/after screenshots.

Currently two CI jobs are failing:
 - verify-code: this is a ktlint formatting issue: run `make fix` to patch this up automatically
 - run-unit-test: we changed some "business logic" so expected that we also need update some tests:

```
com.mapbox.maps.plugin.attribution.AttributionViewImplTest > setMargin FAILED
    java.lang.AssertionError at AttributionViewImplTest.kt:57
```

plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
```diff
@@ -71,7 +71,12 @@ class AttributionViewImpl @JvmOverloads constructor(
    * @param bottom Margin to the bottom in pixel
    */
   override fun setAttributionMargins(left: Int, top: Int, right: Int, bottom: Int) {
-    (layoutParams as FrameLayout.LayoutParams).setMargins(left, top, right, bottom)
+    (layoutParams as FrameLayout.LayoutParams).apply{
+      marginStart = left
+      setTop(top)
```
ank27, line 76: we might want to use `marginTop` instead of `setTop` as we are setting the attributionMargins in this function. same : `marginBottom` instead of `setBottom`

ank27 (approved):

kiryldz (approved):

plugin-attribution/src/main/java/com/mapbox/maps/plugin/attribution/AttributionViewImpl.kt
```diff
@@ -71,7 +71,12 @@ class AttributionViewImpl @JvmOverloads constructor(
    * @param bottom Margin to the bottom in pixel
    */
   override fun setAttributionMargins(left: Int, top: Int, right: Int, bottom: Int) {
-    (layoutParams as FrameLayout.LayoutParams).setMargins(left, top, right, bottom)
+    (layoutParams as FrameLayout.LayoutParams).apply {
+      marginStart = left
```
Chaoba, line 75: Just curious, why not use `leftMargin` and `rightMargin` to be consistent with others?

ank27: in order to support for RTL configuration as well, we use `startMargin` and `endMargin`, rather than `leftMargin` and `rightMargin`. Android system will ignore `left` and `right` in rtl configuration.

Chaoba (approved):

Issue #641, opened 2021-09-14
Location puck is not enabled if style loading is triggered by xml attributes


zeac:
## Environment

- Android OS version: Any
- Devices affected: All
- Maps SDK Version: android-v10.0.0-rc.8

## Observed behavior and steps to reproduce

1. Inflate MapView into your Activity from Activity.onCreate.
2. Enable location puck plugin.
3. The puck is not shown.

The puck will appear if the Activity passed through a stop-start cycle.

This behaviour can be reproduced with a small patch to the demo app:
```
diff --git a/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt b/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
index dbc7ccc9..103a7897 100644
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/LocationTrackingActivity.kt
@@ -7,7 +7,6 @@ import androidx.appcompat.content.res.AppCompatResources
 import com.mapbox.android.gestures.MoveGestureDetector
 import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.MapView
-import com.mapbox.maps.Style
 import com.mapbox.maps.extension.style.expressions.dsl.generated.interpolate
 import com.mapbox.maps.plugin.LocationPuck2D
 import com.mapbox.maps.plugin.gestures.OnMoveListener
@@ -49,6 +48,7 @@ class LocationTrackingActivity : AppCompatActivity() {
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
+
     mapView = MapView(this)
     setContentView(mapView)
     locationPermissionHelper = LocationPermissionHelper(this)
@@ -63,12 +63,8 @@ class LocationTrackingActivity : AppCompatActivity() {
         .zoom(14.0)
         .build()
     )
-    mapView.getMapboxMap().loadStyleUri(
-      Style.MAPBOX_STREETS
-    ) {
-      initLocationComponent()
-      setupGesturesListener()
-    }
+    initLocationComponent()
+    setupGesturesListener()
   }
 
   private fun setupGesturesListener() {
```

## Expected behavior

The puck is visible.

## Notes / preliminary analysis

[...]

PR #680, opened 2021-09-27
Call plugin onStart after start loading style

axti:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix initialisation location puck when no style loaded from code by changing Plugin#onStart() call after style loaded</changelog>`.

### Summary of changes
In this PR has been changed order of methods called on lifecycle event onStart. Now plugins will receive onStart after style has been loaded. This reorder prevent of loosing callbacks from StyleObserver when callback were added before style starts loading.

### User impact (optional)
User with styleUri attribute in XML will have all plugins started correctly after style has been loaded. Before in this case it was the issue reported here: #641

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -49,7 +49,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   private val pluginRegistry: MapPluginRegistry
   private val onStyleDataLoadedListener: OnStyleDataLoadedListener
   private val onCameraChangedListener: OnCameraChangeListener
-  private var lifecycleState: LifecycleState = LifecycleState.STATE_STOPPED
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal var lifecycleState: LifecycleState = LifecycleState.STATE_STOPPED
 
   constructor(
     renderer: MapboxRenderer,
@@ -314,12 +315,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     pluginRegistry.onAttachedToWindow(mapView)
   }
 
-  @VisibleForTesting(otherwise = VisibleForTesting.NONE)
-  internal fun simulateStartedState() {
-    lifecycleState = LifecycleState.STATE_STARTED
-  }
-
-  private enum class LifecycleState {
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal enum class LifecycleState {
     STATE_STOPPED,
     STATE_STARTED,
     STATE_DESTROYED
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -2,6 +2,7 @@ package com.mapbox.maps
 
 import android.content.Context
 import android.view.MotionEvent
+import androidx.annotation.VisibleForTesting
 import com.mapbox.annotation.module.MapboxModuleType
 import com.mapbox.common.Logger
 import com.mapbox.common.module.provider.MapboxModuleProvider
@@ -86,6 +87,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
   }
 
+  @VisibleForTesting(otherwise = VisibleForTesting.NONE)
   constructor(
     renderer: MapboxRenderer,
     nativeObserver: NativeObserver,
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -123,13 +123,13 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
     }
     renderer.onStart()
-    pluginRegistry.onStart()
     if (!mapboxMap.isStyleLoadInitiated) {
       // Load the style in mapInitOptions if no style has loaded yet.
       mapInitOptions.styleUri?.let {
         mapboxMap.loadStyleUri(it)
       }
     }
+    pluginRegistry.onStart()
```
tobrun, line 132: On internal convo, you mentioned:
```
What if we will change here:
    pluginRegistry.onStart()
    if (!mapboxMap.isStyleLoadInitiated) {
      // Load the style in mapInitOptions if no style has loaded yet.
      mapInitOptions.styleUri?.let {
        mapboxMap.loadStyleUri(it)
      }
    }
to do like this:
    if (!mapboxMap.isStyleLoadInitiated) {
      // Load the style in mapInitOptions if no style has loaded yet.
      mapInitOptions.styleUri?.let {
        mapboxMap.loadStyleUri(it)
        pluginRegistry.onStart()
      }
    }
```

Proposed change here is actually not part of the style loading callback?

pengdev: I think we should call `onStart` inside the Activity's `onStart` event, regardless of the style loading status. 
and I remember we do have a `onStyleChanged` listener, we could hook to this listener inside the plugin if the style is need for certain action, wdyt?

axti: @pengdev you are right, this `onStart` is initially called from Activity/Fragment and implementation is inside *MapController*, It is entry point.

pengdev (approved):
Great work @axti , LGTM, please add tests as well 🚀

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: > Proposed change here is actually not part of the style loading callback?


I think the point of this change is that the new style load will clear the style load listeners that previous `getStyle() { style -> } ` sets. 
In some plugins we rely on the async `getStyle()` call to do the initialisation in the plugin's onStart(). And when the new style is loaded after the plugin's onStart, it will result in the plugins' getStyle() callback never being fired.
This fix intentionally moved the pluginRegistry.onStart() after the `loadStyleUri` call, so that we made sure the styleLoaded listener set by plugins' getStyle() is not cleared.

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -8,250 +8,310 @@ import com.mapbox.maps.loader.MapboxMapStaticInitializer
...
-    mapController.onStart()
-    mapController.onStop()
-    verify { renderer.onStop() }
-    verify { pluginRegistry.onStop() }
+    every { mockPluginRegistry.onStop() } just Runs
+    every { mockRenderer.onStop() } just Runs
+    every { mockNativeObserver.removeOnCameraChangeListener(any()) } just Runs
+    every { mockNativeObserver.removeOnStyleDataLoadedListener(any()) } just Runs
+
+    testMapController.simulateStartedState()
```
pengdev, line 107: can we use `onStart` directly instead of using `simulateStartState`?

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -8,250 +8,310 @@ import com.mapbox.maps.loader.MapboxMapStaticInitializer
...
   fun onStart() {
-    mapController.onStart()
-    verify { renderer.onStart() }
-    verify { pluginRegistry.onStart() }
-    verify { mapboxMap.loadStyleUri(Style.MAPBOX_STREETS) }
+    every { mockPluginRegistry.onStart() } just Runs
+    every { mockMapboxMap.loadStyleUri(Style.MAPBOX_STREETS) } just Runs
+    every { mockNativeObserver.addOnCameraChangeListener(any()) } just Runs
+    every { mockNativeObserver.addOnStyleDataLoadedListener(any()) } just Runs
+    every { mockRenderer.onStart() } just Runs
```
pengdev, line 71: nit: we can use relaxed mockk to avoid adding `every` for these functions.

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
kiryldz: +1 for that and remove `simulateStartState` at all. imho there's no possibility in reality to set `lifecycleState = LifecycleState.STATE_STARTED` and __not__ call `onStart` at the same time

kiryldz: > I disagree with relaxed, I'm trying to follow Arrange-Action-Assert pattern to write a good tests.

Is this comment related to https://github.com/mapbox/mapbox-maps-android/pull/680#discussion_r721203936 actually?

kiryldz (approved):

PR #683, opened 2021-09-27
Update titles/descriptions of style/expression examples

colleenmcginnis:
This PR suggests some ways to make the descriptions more specific. I think including a generic title and a specific description (mentioning specific classes/methods/etc) will help with discoverability when we publish examples to docs.mapbox.com pages (I'm pulling these titles and descriptions programmatically into the metadata for each example page).

The thinking is that examples will show up more prominently in search results if a developer is using a query like "switchCase" if they know they want to use that expression, but want to see an example in the larger context of an activity.

@sbma44 @pengdev @tobrun I did my best to summarize the classes/methods used for the primary purpose of each example, but feel free to make direct edits if any of my interpretations of the example code are incorrect!


tobrun (approved):

pengdev (approved):
Thanks for the work!

app/src/main/res/values/example_titles.xml
```diff
@@ -17,7 +17,7 @@
     <string name="activity_map_viewpager">ViewPager</string>
     <string name="activity_map_multi">Multi map fragments</string>
     <string name="activity_debug_mode">Debug mode</string>
-    <string name="activity_fill_extrusion">Display building extrusions</string>
+    <string name="activity_fill_extrusion">Display 3D buildings</string>
```
colleenmcginnis, line 20: Aligns with iOS and IMO is easier to understand at first glance. (The term "extrusion" is still used in the description to tie back to the technical language.)

app/src/main/res/values/example_titles.xml
```diff
@@ -47,11 +47,11 @@
     <string name="activity_lab_moving_icon_with_trailing_line_title">Line behind moving icon</string>
     <string name="activity_labs_snaking_directions_route_title">Animate a line</string>
     <string name="activity_animate_marker_position">Animate Marker Position</string>
-    <string name="activity_styles_show_hide_layers_title">Show/hide layers</string>
+    <string name="activity_styles_show_hide_layers_title">Toggle the visibility of a layer</string>
```
colleenmcginnis, line 50: Align with iOS: 

```suggestion
    <string name="activity_styles_show_hide_layers_title">Show and hide layers</string>
```

ank27 (approved):

PR #701, opened 2021-09-30
Use correct enum for setting render thread priority

kkaefer:
In https://github.com/mapbox/mapbox-maps-android/blob/ecd6f73471dc4d1d049415134c35a3b7b2c89632/sdk/src/main/java/com/mapbox/maps/renderer/WorkerHandlerThread.kt#L26, we’re setting the rendering thread’s `HandlerThread`'s priority to `Thread.MAX_PRIORITY` with the intention that the rendering thread should be prioritized over other threads, e.g. worker threads. Unfortunately, Android uses plain integer instead of enums, and the compiler doesn’t catch that this is the wrong kind of constant. The [Android docs say this about `HandlerThread`'s constructor](https://developer.android.com/reference/android/os/HandlerThread):

> The value supplied must be from `Process` and not from java.lang.Thread.

Instead, we should use [ `Process.THREAD_PRIORITY_DISPLAY` ](https://developer.android.com/reference/android/os/Process#THREAD_PRIORITY_DISPLAY) (or maybe `THREAD_PRIORITY_FOREGROUND`). The docs also say that “Applications can not normally change to this priority”, but in my limited testing, changing a `HandlerThread`’s priority to this worked fine.

`<changelog>Correctly set the render thread priority</changelog>`


pengdev (approved):
Thanks for catching it 🚀

axti:
In Android we have two ways to set priority to threads : `Thread.setPriority()` and `Process.setThreadPriority()`.
Thread.setPriority() contains a value from MIN_PRIORITY(1) to MAX_PRIORITY(10)
The second way to set priorities is to call Process.setThreadPriority(). It supports value from -20 to 19.
Here is the mapping from Thread class priority to Process class Priority:
image
As you can see THREAD_PRIORITY_DISPLAY = -4 it is equals to Thread.priority = 7. 

In your PR you change priority wrongly, because when you do `priority = ` you actually change Thread.priority and your code will not work, it should trigger exception: 
```
if (newPriority > MAX_PRIORITY || newPriority < MIN_PRIORITY) {
// Android-changed: Improve exception message when the new priority is out of bounds.
throw new IllegalArgumentException("Priority out of range: " + newPriority);
}
```
To us `Process.THREAD_PRIORITY_DISPLAY` you should put it in constructor:
`handlerThread = HandlerThread(HANDLE_THREAD_NAME, Process.THREAD_PRIORITY_DISPLAY).apply {`

If you want to verify any thread priority you can run command in `adb shell` : `ps -T -l -p <pid>` it will show all thread for the current pid.
this is how it looks before change:
1 R 10146 20281 20358   325 30 19   **0**   - 3855022 0    ?        00:00:02 MapboxRenderThr
this is how it looks after change(but put correct priority in the constructor)
1 S 10146 19979 20073   325 15 23  **-4**   - 3797753 0    ?        00:00:27 MapboxRenderThr

sdk/src/main/java/com/mapbox/maps/renderer/WorkerHandlerThread.kt
```diff
@@ -23,7 +24,7 @@ internal class WorkerHandlerThread {
 
   fun start() {
     handlerThread = HandlerThread(HANDLE_THREAD_NAME).apply {
```
axti, line 26:
```suggestion
    handlerThread = HandlerThread(HANDLE_THREAD_NAME, Process.THREAD_PRIORITY_DISPLAY).apply {
```

axti:
I have seen that our Worker threads are more prioritised:
1 S 10146 20281 20360   325 1  29 **-10**   - 3855022 0    ?        00:00:00 Worker 1
1 S 10146 20281 20361   325 1  29 **-10**   - 3855022 0    ?        00:00:00 Worker 2
1 S 10146 20281 20362   325 1  29 **-10**   - 3855022 0    ?        00:00:00 Worker 3
1 S 10146 20281 20363   325 0  29 **-10**   - 3855022 0    ?        00:00:00 Worker 4

kiryldz:
@kkaefer thanks for catching this! However some more tailwork and verification needs to be done in this PR.

@axti please take a deeper look at this PR in terms of fixing tests, verifying that it indeed works, that this will not break regular Android devices etc, make sure that native worker threads will still work as expected etc. Also would be nice to use our standard Android PR template, add more visuals from the profiler if there's any difference at all etc.

kkaefer:
@axti thanks for digging in here as well. I realized that this PR is incorrect as well last night, but didn’t get around to posting my findings here. I think we should change to passing `Process.THREAD_PRIORITY_DISPLAY` in the constructor, like you suggested.

> I have seen that our Worker threads are more prioritised

I noticed that too, and I’m currently trying to find out why that is the case. The opposite should be true.

tobrun (approved):

kiryldz (changes requested):
@axti btw this thread is used only for rendering, initially I thought we could re-use it for some other purposes but this fix actually makes it render only.
Can you rename this class to `RenderHandlerThread`?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -25,7 +25,7 @@ import kotlin.math.pow
  */
 internal class MapboxRenderThread : Choreographer.FrameCallback {
 
-  internal val handlerThread: WorkerHandlerThread
+  internal val handlerThread: RenderHandlerThread
```
alexander-kulikovskii, line 28: maybe choose one name for instance of this object?
I found three different names: `handlerThread`, `renderThread`, `renderHandlerThread`

I vote for `renderHandlerThread`

alexander-kulikovskii (approved):

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
axti: Done

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRenderThreadTest.kt
```diff
@@ -25,27 +25,27 @@ class MapboxRenderThreadTest {
...
-  private lateinit var workerThread: WorkerHandlerThread
+  private lateinit var renderThread: RenderHandlerThread
   private val waitTime = 200L
 
   @Before
   fun setUp() {
     mapboxRenderer = mockk(relaxUnitFun = true)
     eglCore = mockk(relaxed = true)
-    workerThread = WorkerHandlerThread()
+    renderThread = RenderHandlerThread()
```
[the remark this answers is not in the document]
axti: Done

kiryldz (approved):
LGTM

PR #706, opened 2021-10-04
Init asset file thread only once

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix spawning multiple assert file threads by initing asset file thread only once per application lifecycle.</changelog>`.

### Summary of changes

Fix spawning multiple assert file threads by initing asset file thread only once per application lifecycle.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -56,7 +56,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   ) {
     this.renderer = renderer
     this.mapInitOptions = mapInitOptions
-    AssetManagerProvider().initialize(mapInitOptions.context.assets)
+    if (assetsManagerNeedInit) {
+      assetsManagerNeedInit = false
+      AssetManagerProvider().initialize(mapInitOptions.context.applicationContext.assets)
+    }
     this.nativeMap = MapProvider.getNativeMap(
       mapInitOptions,
       renderer,
@@ -325,6 +328,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       "Add %s plugin dependency to the classpath take automatically load the plugin implementation."
     private const val VIEW_HIERARCHY_MISSING_TEMPLATE =
       "%s plugin requires a View hierarchy to be injected, plugin is ignored."
+    private var assetsManagerNeedInit = true
 
     init {
       MapboxMapStaticInitializer.loadMapboxMapNativeLib()


sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -325,6 +328,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       "Add %s plugin dependency to the classpath take automatically load the plugin implementation."
     private const val VIEW_HIERARCHY_MISSING_TEMPLATE =
       "%s plugin requires a View hierarchy to be injected, plugin is ignored."
+    private var assetsManagerNeedInit = true
```
pozdnyakov, line 331: should `assetsManagerNeedInit` be static?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -56,7 +56,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   ) {
     this.renderer = renderer
     this.mapInitOptions = mapInitOptions
-    AssetManagerProvider().initialize(mapInitOptions.context.assets)
+    if (assetsManagerNeedInit) {
```
pengdev, line 59: shall we keep the assetManager static instead of using a static boolean?

pengdev: like using a static initialiser

kiryldz: wouldn't we leek context in this case?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -56,7 +56,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+      assetsManagerNeedInit = false
+      AssetManagerProvider().initialize(mapInitOptions.context.applicationContext.assets)
```
alexander-kulikovskii, line 61: don't forget about tests 🙂

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: when such bool variable is checked only once I think it looks better when we check it for `true` and not like 
`if (!assetsManagerInitialized) {...}`

alexander-kulikovskii: `assetsManagerNotInitialized` 😄

ok, up to you

kiryldz:
I don't think we need it for now, gl-native should fix-up spawning worker threads. Let's leave current setup as is and not use singleton 😄

PR #718, opened 2021-10-05
Refactor MapEvents listeners, each listener will include an event data property.

pengdev:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Refactor MapEvents listeners, so that each listener will include one event data property.</changelog>`.

### Summary of changes

This PR refactors the MapEventListeners, so that each listener will include one event data property. This unblocks us to add properties to the event data incrementally without breaking our public API.

List of affected listeners:
* OnCameraChangeListener
* OnMapIdleListener
* OnMapLoadedListener
* OnMapLoadErrorListener
* OnRenderFrameFinishedListener
* OnRenderFrameStartedListener
* OnSourceAddedListener
* OnSourceDataLoadedListener
* OnSourceRemovedListener
* OnStyleDataLoadedListener
* OnStyleImageMissingListener
* OnStyleImageUnusedListener
* OnStyleLoadedListener

This PR also refactors the package name for Map event's data models to `com.mapbox.maps.extension.observable` inside the `sdk-base` module, so it can be reused across different plugins.


sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/CameraChangedEventData.kt
```diff
@@ -0,0 +1,6 @@
+package com.mapbox.maps.extension.observable.eventdata
+
+/**
+ * The data class for camera-changed event data in Observer
+ */
+class CameraChangedEventData
```
pengdev, line 6: Added this stub class to hold the camera changed event data, currently there's nothing in this event data, but we could add the data in the future without breaking our API.

We do have the startTime property for all event data, we didn't expose them at the moment.

cc @mapbox/maps-android

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/listeners/OnCameraChangeListener.kt
```diff
@@ -14,6 +15,8 @@ fun interface OnCameraChangeListener {
    *
    * For more precise information about actual map camera animations
    * please check [CameraAnimationsPlugin], [CameraAnimationsLifecycleListener] and [CameraAnimatorChangeListener]
+   *
+   * @param eventData CameraChangedEventData
    */
-  fun onCameraChanged()
+  fun onCameraChanged(eventData: CameraChangedEventData)
```
kiryldz, line 21: just curious: will it break all existing Java code using our SDK?

kiryldz:
overall LGTM but let's first fix all CI issues

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -329,34 +329,41 @@ fun Event.getSourceDataLoadedEventData(): SourceDataLoadedEventData {
...
+  return Gson().fromJson(json, StyleImageMissingEventData::class.java)
+}
 
 /**
  * Get the parsed event data for style image unused event.
- * @return a parsed IDStringEventData.
+ * @return a parsed StyleImageUnusedEventData.
  */
-fun Event.getStyleImageUnusedEventData(): IDStringEventData = getIDStringEventData()
+fun Event.getStyleImageUnusedEventData(): StyleImageUnusedEventData {
```
ank27, line 345: we could keep `IDStringEventData` class that was defined earlier, it has the same purpose as `StyleImageUnusedEventData.kt`, `StyleImageUnusedEventData.kt`, `StyleImageMissingEventData.kt`, wdty?

pengdev: we need to make them extensible for each event, so let's separate them.

sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/listeners/OnCameraChangeListener.kt
pengdev: yes, it will 🙃

sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/CameraChangedEventData.kt
pengdev: Actually we should add the `begin` and optional `end` timestamp property to all events, as they are stated in the Event class' documentation.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/CameraChangedEventData.kt
```diff
@@ -0,0 +1,19 @@
+package com.mapbox.maps.extension.observable.eventdata
+
+import com.google.gson.annotations.SerializedName
+
+/**
+ * The data class for camera-changed event data in Observer
+ */
+data class CameraChangedEventData(
```
kiryldz, line 8: oh interesting so it does have fields

sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/CameraChangedEventData.kt
```diff
@@ -0,0 +1,19 @@
...
+  /**
+   * Representing timestamp taken at the time of an event creation, in microseconds, since the epoch.
+   */
+  @SerializedName("begin") val begin: Long,
+
+  /**
+   * For an interval events, an optional `end` property will be present that represents timestamp taken at the time
+   * of an event completion.
+   */
+  @SerializedName("end") val end: Long?
```
kiryldz, line 18: why is it nullable? 🤔

pengdev: From Event class' documentation:
```
    /**
     * Generic container for an event's data (Object). By default, event data will contain `begin` key, whose value
     * is a number representing timestamp taken at the time of an event creation, in microseconds, since the epoch.
     * For an interval events, an optional `end` property will be present that represents timestamp taken at the time
     * of an event completion. Additional data properties are docummented by respective events.
     *
     * ``` text
     * Event data format (Object):
     * .
     * ├── begin - Number
     * └── end - optional Number
     * ```
     */
```

sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/CameraChangedEventData.kt
pengdev: yes, all the events have the begin field and optional end field.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -5,8 +5,12 @@ import com.mapbox.maps.Event
 import com.mapbox.maps.MapEvents
 import com.mapbox.maps.ObservableInterface
 import com.mapbox.maps.Observer
-import com.mapbox.maps.extension.observable.model.*
-import com.mapbox.maps.extension.observable.resourcerequest.ResourceEventData
+import com.mapbox.maps.extension.observable.eventdata.*
+import com.mapbox.maps.extension.observable.eventdata.ResourceEventData
+
+private val gson by lazy {
+  Gson()
```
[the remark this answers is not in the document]
kiryldz: however could it technically happen that user is parsing several instances of event data from multiple threads? guess it's too advanced though 😄 so let's keep as is

kiryldz (approved):
@pengdev LGTM 💎

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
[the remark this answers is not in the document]
pengdev: According to Gson 's [api doc](https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.0/com/google/gson/Gson.html)

```
Gson is typically used by first constructing a Gson instance and then invoking toJson(Object) or fromJson(String, Class) methods on it. Gson instances are Thread-safe so you can reuse them freely across multiple threads.
```

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -299,7 +303,16 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
  */
 fun Event.getResourceEventData(): ResourceEventData {
   val json = data.toJson()
-  return Gson().fromJson(json, ResourceEventData::class.java)
+  return gson.fromJson(json, ResourceEventData::class.java)
```
alexander-kulikovskii, line 306: should we add try-catch block?

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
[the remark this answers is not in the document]
alexander-kulikovskii: how about [koltin serialization](https://github.com/Kotlin/kotlinx.serialization)?
maybe not in this PR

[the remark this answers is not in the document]
pengdev: > how about [koltin serialization](https://github.com/Kotlin/kotlinx.serialization)? maybe not in this PR

we were using gson library to serialise/deserialise in our codebase, it would be good to evaluate and replace it across the codebase in a separate PR.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
pengdev: imho if the data format has been changed, it should be a bug and should throw exceptions so that we can fix it in a patch.

ank27 (approved):

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/ResponseErrorReason.kt
```diff
@@ -1,11 +1,23 @@
-package com.mapbox.maps.extension.observable.resourcerequest.eventdata
+package com.mapbox.maps.extension.observable.model
 
 import com.google.gson.annotations.SerializedName
 
 /**
- * Describes type of Error for response object.
+ * Describes the qreason of Error for response object.
```
ank27, line 6:
```suggestion
 * Describes the reason of Error for response object.
```

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -299,7 +303,16 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
+}
+
+/**
+ * Get the parsed event data for map loaded event.
+ * @return a parsed MapLoadedEventData object.
+ */
+fun Event.getMapLoadedEventData(): MapLoadedEventData {
+  val json = data.toJson()
```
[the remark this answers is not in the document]
pengdev: We could generate the `MapLoadedEventData` using the formal spec, but I think we still need to deserialise it using gson or [koltin serialization ](https://github.com/Kotlin/kotlinx.serialization)

> Modification of an internal logic would be a non-breaking change?

Addition of properties inside `MapLoadedEventData` shouldn't be a breaking change, however, we should not remove or modify the existing properties.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/eventdata/MapLoadedEventData.kt
```diff
@@ -0,0 +1,19 @@
+package com.mapbox.maps.extension.observable.eventdata
+
+import com.google.gson.annotations.SerializedName
+
+/**
+ * The data class for map-loaded event data in Observer
+ */
+data class MapLoadedEventData(
```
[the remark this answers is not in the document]
pengdev: Unfortunately Kotlin data classes do not play too well with inheritance. For example, it's known that there's no way to implement equals() correctly in a hierarchy on non-abstract classes.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
alexander-kulikovskii: we can discuss it with the team. probably leave as is for now. but let's think about it

several options for now:

- return nullable
- support some default value and work with it somehow

also we can send telemetry event for such crashes

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/TileID.kt
```diff
@@ -1,4 +1,4 @@
-package com.mapbox.maps.plugin.delegates.listeners.eventdata
+package com.mapbox.maps.extension.observable.model
```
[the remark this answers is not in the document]
pengdev: great catch, replaced `Double` with `Long` in TileID

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
pengdev: Sounds good, let's make it as a follow up 👍

tobrun (approved):

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
tobrun: > yeah, but is it proper behavior for users? is it possible that our users (developers) can release some app in play market and got crash in app?

That statement could be true for every line of code we ship. The program is expected to return this non null data in that specific format, it's part of the contract that gl-native makes with us. Overall, I'm personally leaning more towards a crash-fast approach, we catch issues during development, else they can become unnoticed and broken if not properly exercised. Additionally introducing null, will require developers to handle null which add logical branching to their code. This makes it more complex to handle, while it's not expected program output.

Issue #564, opened 2021-08-12
Map move uses wrong (faster) map offsetting after zoom-in gesture


daniel-j-h:
Here is a reproducible issue with the gesture system.

To reproduce
1. use the two-finger zoom-in gesture (start with your two fingers in the middle of the screen and move them outwards)
2. only release one finger from the screen
3. with the other finger now move the map around
4. you will notice the map no longer follows the finger but moves around way faster (probably using the zoom level before the zoom-in gesture happened)

Please try to reproduce this and let me know if I should make a screen cast if you don't see this.

daniel-j-h:
The same issue happens with the zoom-out gesture, then the map moves slower than it should.

Can you folks reproduce this or do you need more details, since it's a gesture based issue?

PR #738, opened 2021-10-12
Update the logic to invoke dragStart

Chaoba:
Fixes: #564 

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Update the logic to invoke dragStart to fix map move speed faster or slower issue</changelog>`.

### Summary of changes
In the past, we invoke the `dragStart` method while scale gesture starts, which can make the map move speed faster or slower than normal. In this pr, `dragStart` method  will only be invoked when a real drag event starts, so will not have the upper issue.

Before

https://user-images.githubusercontent.com/8577318/136907352-98a29dde-ae4e-4518-ba89-2f3fb26b4f92.mp4

After

https://user-images.githubusercontent.com/8577318/136906722-73cbb262-f29b-4b59-aa62-72b6def05be3.mp4

### DIFF
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
@@ -73,8 +73,8 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
   private var spanSinceLast: Float = 0.0f
   private var screenHeight: Double = 0.0
   private var startZoom: Double = 0.0
-  private var flingInProcess: Boolean = false
   private var scaleCachedAnchor: ScreenCoordinate? = null
+  private var dragInProgress: Boolean = false
 
   // Rotate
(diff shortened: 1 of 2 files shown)


plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1235,10 +1221,18 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
         return true
       }
 
+      if (detector.pointersCount > 1) {
+        return false
+      }
```
pengdev, line 1226: @Chaoba Any reason we need to do this check? Because it removes our capability to do two finger pan gesture.

pengdev: Created a draft PR to enable two-finger pan at https://github.com/mapbox/mapbox-maps-android/pull/1280

PR #742, opened 2021-10-12
Initialize viewport plugin

axti:

## Pull request checklist:

### Summary of changes

This PR is adds the initial version of  viewport plugin:
* Port the viewport camera logic from navigation sdk and add showcase activity.
* Extract the plugin interfaces to sdk-base module.
* Refactor `ViewportCamera` to `ViewportPluginImpl` and hook its initialisation into the sdk's lifecycle.

Remaining work:
* Currently `MapboxViewportDataSource` is directly accessed and customised in the test app, we need to abstract generic public interface from MapboxViewportDataSource and write a MVP tracking modes implementation using these interfaces, and move the custom implementation to the test app.
* Extract viewport API public config options out to serialisation spec, and generate the settings interface.
* Validate the public API with navigation SDK's use case.


plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/ViewportDataSourceProcessor.kt
```diff
@@ -0,0 +1,97 @@
+package com.mapbox.maps.plugin.viewport.data
+
+import com.mapbox.common.Logger
+import com.mapbox.maps.EdgeInsets
+import com.mapbox.maps.ScreenBox
+import com.mapbox.maps.ScreenCoordinate
+import com.mapbox.maps.Size
+
+internal object ViewportDataSourceProcessor {
```
Chaoba, line 9: Add java doc for this object.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportCameraTransition.kt
```diff
@@ -0,0 +1,266 @@
...
+
+    var centerDuration = 0L
+    cameraOptions.center?.let { center ->
+      val screenDistanceFromMapCenterToLocation = screenDistanceFromMapCenterToTarget(
+        cameraManager = cameraManager,
+        currentCenter = currentMapCameraState.center,
+        targetCenter = center
+      )
+
+      val centerAnimationRate = 500L
```
Chaoba, line 38: There are so many magic numbers here.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+    animatorSet: AnimatorSet,
+    instant: Boolean,
+    transitionEndListener: TransitionEndListener? = null,
+  ) {
+    cancelAnimation()
+    if (transitionEndListener != null) {
+      transitionEndListeners.add(transitionEndListener)
+    }
+    animatorSet.childAnimations.forEach {
+      cameraPlugin.registerAnimators(it as ValueAnimator)
```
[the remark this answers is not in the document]
pengdev: we need to cast the Animator to ValueAnimator for each of the child animators, so I think it's fine to use a foreach here.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,275 @@
...
+import com.mapbox.maps.plugin.viewport.transition.normalizeBearing
+import com.mapbox.maps.toCameraOptions
+import java.util.concurrent.CopyOnWriteArraySet
+import kotlin.math.max
+import kotlin.math.min
+
+/**
+ * Mapbox default implementation for ViewportDataSource.
+ */
+@UiThread
```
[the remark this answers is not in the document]
pengdev: The updates should come in the main thread, I guess it's the reason we added it here, cc @axti  for more context?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,275 @@
...
+    viewportDataSourceUpdateObservers.remove(viewportDataSourceUpdateObserver)
+  }
+
+  /**
+   * Call whenever new user location is available.
+   *
+   * @see [evaluate]
+   */
+  fun onLocationChanged(location: Location) {
+    targetLocation = location
```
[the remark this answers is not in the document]
pengdev: I assume so, I have seen the evaluate called after the `onLocationChanged` at the showcase activity.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/ViewportDataSourceProcessor.kt
pengdev: This is an internal class, no documentation is required, but would add it later when we finalise the public interfaces.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,275 @@
...
+import com.mapbox.maps.plugin.viewport.data.ViewportProperty.PitchProperty
+import com.mapbox.maps.plugin.viewport.data.ViewportProperty.ZoomProperty
+import com.mapbox.maps.plugin.viewport.transition.normalizeBearing
+import com.mapbox.maps.toCameraOptions
+import java.util.concurrent.CopyOnWriteArraySet
+import kotlin.math.max
+import kotlin.math.min
+
+/**
+ * Mapbox default implementation for ViewportDataSource.
```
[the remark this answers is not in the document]
pengdev: This class requires some further refactor, as we are likely to extract the required API to a public interface and make this class internal. We will need to revisit the documentation when the public interface is finalised.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,275 @@
...
+            pointsForFollowing,
+            fallbackCameraOptions,
+            screenBox
+          )
+        } else {
+          fallbackCameraOptions
+        }
+      }
+
+      followingCenterProperty.fallback = cameraFrame.center!!
```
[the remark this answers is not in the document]
pengdev: not sure if we want to assert the cameraFrame.center is non-null and throw some exception. Using ?. will silently ignore some errors that might not be easy to debug later.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportCameraTransition.kt
axti: It is a timings for animation. If you would like to proceed the animation faster or slowly please adjust constants.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,275 @@
...
+  private fun updateOverviewData() {
+    targetLocation?.let { nonNullTargetLocation ->
+      val pointsForOverview = listOf(nonNullTargetLocation.toPoint())
+
+      overviewBearingProperty.fallback = normalizeBearing(
+        cameraManager.cameraState.bearing,
+        BEARING_NORTH
+      )
+
+      val cameraFrame = if (pointsForOverview.isNotEmpty()) {
```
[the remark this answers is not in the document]
axti: This is a leftover from copying, yes it should be simplified. My intention was to clear on the next step, when we will adjust logic for our needs.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSourceOptions.kt
```diff
@@ -0,0 +1,307 @@
...
+ * Options that impact the generation of the following frame.
+ */
+class FollowingFrameOptions internal constructor() {
+
+  /**
+   * The default pitch that will be generated for following camera frames.
+   *
+   * Defaults to `45.0` degrees.
+   */
+  var defaultPitch = 45.0
```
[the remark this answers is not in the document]
pengdev: these options should be generated through serialisation spec later on, will leave it as is for now.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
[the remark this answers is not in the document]
pengdev: Let's leave it for now, I think we might want to introduce multiple points for overview in the next steps, we can revisit later.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportCameraTransition.kt
pengdev: let's iterate in follow-up PRs.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,237 @@
...
+import com.mapbox.maps.plugin.locationcomponent.LocationProvider
+import com.mapbox.maps.plugin.locationcomponent.location
+import com.mapbox.maps.plugin.viewport.ViewportPlugin
+import com.mapbox.maps.plugin.viewport.data.MapboxViewportDataSource
+import com.mapbox.maps.plugin.viewport.state.ViewportCameraState
+import com.mapbox.maps.plugin.viewport.viewport
+import com.mapbox.maps.testapp.R
+import com.mapbox.maps.testapp.databinding.ActivityViewportAnimationBinding
+
+class ViewportShowcaseActivity : AppCompatActivity() {
```
tobrun, line 30: nit some class level docs explaining what this example does

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,237 @@
...
+          }
+          addOnIndicatorPositionChangedListener {
+            latestLocation.latitude = it.latitude()
+            latestLocation.longitude = it.longitude()
+          }
+          addOnIndicatorBearingChangedListener {
+            latestLocation.bearing = it.toFloat()
+            viewportDataSource.onLocationChanged(latestLocation)
+            viewportCamera.resetFrame()
+          }
```
tobrun, line 135: We have two callbacks from the SDK when location changes but only in bearing we are updating the viewportdatasource. This is probably due to the 2 independent callbacks. Would it make sense to have an unified callback that both provides location as bearing in one go?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,237 @@
...
+        }
+      }
+    }
+    setupViewportCamera()
+  }
+
+  lateinit var latestLocation: Location
+  lateinit var viewportCamera: ViewportPlugin
+  lateinit var viewportDataSource: MapboxViewportDataSource
```
tobrun, line 144: nit move to the other fields above the onCreate?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+   * the current state to the requested state.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 3500 millis.
+   * @param frameTransitionOptionsBlock options that impact the transition animations between
+   * viewport frames in the selected state.
+   * This refers to camera transition on each [ViewportDataSource] update when [Following] is engaged.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 1000 millis.
+   * @param transitionEndListener invoked when transition ends.
+   */
+  @JvmOverloads
+  override fun requestNavigationCameraToFollowing(
```
tobrun, line 83: `requestNavigationCameraToFollowing` this wording is too navigation specific, would love to be able to generalize that

tobrun: same for all other overrides in this class, see if we can drop navigation naming

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,237 @@
...
+import com.mapbox.maps.testapp.databinding.ActivityViewportAnimationBinding
+
+class ViewportShowcaseActivity : AppCompatActivity() {
+  private var emitCount = 0
+  private var delta = 0f
+  private val handler = Handler(Looper.getMainLooper())
+  private lateinit var mapboxMap: MapboxMap
+  private lateinit var mapView: MapView
+  private lateinit var viewportButton: Button
+  private val pixelDensity = Resources.getSystem().displayMetrics.density
```
kiryldz, line 37: we avoid getting density thru singleton as this does not work for multi-screen use-case

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+import com.mapbox.maps.plugin.viewport.transition.TransitionEndListener
+import com.mapbox.maps.plugin.viewport.transition.ViewportCameraStateTransition
+import com.mapbox.maps.plugin.viewport.transition.ViewportCameraTransition
+import com.mapbox.maps.plugin.viewport.transition.ViewportCameraTransitionOptions
+import java.util.concurrent.CopyOnWriteArraySet
+
+/**
+ * Mapbox default implementation for Viewport plugin.
+ */
+class ViewportPluginImpl : ViewportPlugin {
```
kiryldz, line 35: I assume impl should be `internal`

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+   * @param stateTransitionOptionsBlock options that impact the transition animation from
+   * the current state to the requested state.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 3500 millis.
+   * @param frameTransitionOptionsBlock options that impact the transition animations between
+   * viewport frames in the selected state.
+   * This refers to camera transition on each [ViewportDataSource] update when [Following] is engaged.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 1000 millis.
+   * @param transitionEndListener invoked when transition ends.
+   */
+  @JvmOverloads
```
kiryldz, line 82: why overloads? do not see default parameters

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * Executes a transition to [Following] state. When started, goes to [TransitionToFollowing]
+   * and to the final [Following] when ended. If transition is canceled, state goes to [Idle].
+   *
+   * The target camera position is obtained with [ViewportDataSource.getViewportData].
+   *
+   * @param stateTransitionOptionsBlock options that impact the transition animation from
+   * the current state to the requested state.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 3500 millis.
```
kiryldz, line 75: move 3500 to const and add reference here?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  /**
+   * Executes a transition to [Following] state. When started, goes to [TransitionToFollowing]
+   * and to the final [Following] when ended. If transition is canceled, state goes to [Idle].
+   *
+   * The target camera position is obtained with [ViewportDataSource.getViewportData].
+   *
+   * @param stateTransitionOptions options that impact the transition animation from the current state to the requested state.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 3500 millis.
```
kiryldz, line 102: refer to constant instead of 3500

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+   * @param frameTransitionOptions options that impact the transition animations between viewport frames in the selected state.
+   * This refers to camera transition on each [ViewportDataSource] update when [Following] is engaged.
+   * Defaults to [ViewportCameraTransitionOptions.maxDurationMs] equal to 1000 millis.
```
kiryldz, line 105: refer to constant instead of 1000

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
kiryldz: applies to all methods below

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+      TransitionToFollowing -> transitionEndListener?.let { listener ->
+        transitionEndListeners.add(listener)
+      }
+      Following -> transitionEndListener?.onTransitionEnd(isCanceled = false)
+      Idle, TransitionToOverview, Overview -> {
+        val data = viewportDataSource.getViewportData()
+        startAnimation(
+          stateTransition.transitionToFollowing(data.cameraForFollowing, stateTransitionOptions)
+            .apply {
+              addListener( // todo
```
kiryldz, line 124: add more information about todo

kiryldz: applies to methods below

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
kiryldz: applies to everything below

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+      transitionEndListeners.add(it)
+    }
+    animatorSet.childAnimations.forEach {
+      cameraPlugin.registerAnimators(it as ValueAnimator)
+    }
+    if (instant) {
+      animatorSet.duration = 0
+    }
+
+    // workaround for https://github.com/mapbox/mapbox-maps-android/issues/277
```
kiryldz, line 299: this issues is closed so guess could remove this workaround

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,389 @@
...
+    ) ?: throw InvalidPluginConfigurationException(
+      "Can't look up an instance of location component plugin, " +
+        "is it available on the clazz path and loaded through the map?"
+    )
+  }
+
+  /**
+   * Called when the plugin is first added to the map.
+   */
+  override fun initialize() {
```
kiryldz, line 377: are we sure nothing should be cleared inside overloaded `cleanup`?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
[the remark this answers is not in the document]
kiryldz: imho if we're using annotations - let's use them everywhere
my impression though is that every class related to the viewport is accessed from main thread so I see no sense in marking some specific class
those annotations are pretty useful when some class is meant to be thread-safe (e.g. our renderer)

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,272 @@
...
+import com.mapbox.maps.toCameraOptions
+import java.util.concurrent.CopyOnWriteArraySet
+import kotlin.math.max
+import kotlin.math.min
+
+/**
+ * Mapbox default implementation for ViewportDataSource.
+ */
+@UiThread
+class MapboxViewportDataSource(
```
kiryldz, line 27: guess should be internal

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,272 @@
...
+          pointsForOverview,
+          overviewPadding,
+          overviewBearingProperty.get(),
+          overviewPitchProperty.get()
+        )
+      } else {
+        cameraManager.cameraState.toCameraOptions()
+      }
+
+      overviewCenterProperty.fallback = cameraFrame.center!!
```
kiryldz, line 246: I'd avoid using `!!` inside SDK files, let's handle those errors somehow (use `RuntimeException` with some text). Applies to other things below

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
```diff
@@ -0,0 +1,272 @@
...
+    options.overviewFrameOptions.run {
+      val cameraState = cameraManager.cameraState
+      overviewBearingProperty.fallback = cameraState.bearing
+      overviewPitchProperty.fallback = cameraState.pitch
+      overviewCenterProperty.fallback = cameraState.center
+      overviewZoomProperty.fallback = min(cameraState.zoom, maxZoom)
+    }
+  }
+
+  private fun Location.toPoint(): Point = Point.fromLngLat(longitude, latitude)
```
kiryldz, line 263: If there are more things like that - move to Util class?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSourceOptions.kt
```diff
@@ -0,0 +1,307 @@
+package com.mapbox.maps.plugin.viewport.data
+
+import android.location.Location
+
+/**
+ * Options that impact generation of frames.
+ */
+class MapboxViewportDataSourceOptions internal constructor() {
```
kiryldz, line 8: should be internal I guess

kiryldz: applies to almost all the files in the PR, imho we should expose only user-facing interfaces, everything else should be hidden

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/ViewportDataSourceProcessor.kt
```diff
@@ -0,0 +1,97 @@
...
+      return 0.0
+    }
+    val rawAngleDiff = angle - anchorAngle
+    return wrap(rawAngleDiff, -180.0, 180.0)
+  }
+
+  private fun wrap(angle: Double, min: Double, max: Double): Double {
+    val d = max - min
+    return ((((angle - min) % d) + d) % d) + min
+  }
```
kiryldz, line 96: remember smth similar in location-component plugin, perhaps we could re-use it then

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/ViewportPlugin.kt
```diff
@@ -0,0 +1,159 @@
...
+import com.mapbox.maps.plugin.viewport.transition.TransitionEndListener
+import com.mapbox.maps.plugin.viewport.transition.ViewportCameraTransition
+import com.mapbox.maps.plugin.viewport.transition.ViewportCameraTransitionOptions
+
+/**
+ * The Viewport plugin allows to track objects on a map.
+ *
+ * Most common use-case is the ability to track the user location and animate the location updates
+ * to create a navigation experience. Next to tracking a single object, the API allows tracking
+ * multiple objects. The viewport needs to automatically optimize the camera position to take in
```
kiryldz, line 21: it's hard to say for me - is it true that we could already track multiple objects with given API?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: There are cases where these two updates might come independently, for example, if we want to update the bearing according to the magnetic sensor on the phone instead of the one from GPS location.

However I think we should update viewPortDataSource in both listeners.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: yes, this PR is meant to be the initial PR to drop in as much as navigation logic and iterate the public API in follow up PRs, but will rename these interfaces in this PR.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: It have to be public since we now use concrete class constructor instead of reflection to initialise the plugin instances.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: The story is that these methods had default value, and these default values is moved to the interface during the plugin interface extraction, and interface can not use `JvmOverloads` annotation 🙃 

I will remove these annotations for now, and maybe we need to manually overload these functions later.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: We need to revisit this later, as these logic should move to the plugin. And we don't have a valid context at this phase, using the pixelRatio from the context would require bigger refactor.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: @axti could you give more context on what's left?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
pengdev: As mentioned in the PR's description. Currently the activity is accessing this class directly, we need to refactor and abstract the public interface from it in the follow-up PRs.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSourceOptions.kt
pengdev: As mentioned in the PR's description. Currently the activity is accessing this class directly, we need to refactor and abstract the public interface from it in the follow-up PRs.

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/ViewportPlugin.kt
pengdev: The documentation is from the SDD, we will continue working on these features in follow up PR, as this is just the initial PR that ports the nav camera and try to make it a working plugin.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
pengdev: Will revisit it in the follow up PR to refactor the `MapboxViewportDataSource`

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/MapboxViewportDataSource.kt
pengdev: good idea

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/data/ViewportDataSourceProcessor.kt
pengdev: we could probably reuse some of these utilities, will follow up in separate PR.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+            latestLocation.latitude = it.latitude()
+            latestLocation.longitude = it.longitude()
+            viewportDataSource.onLocationChanged(latestLocation)
+            viewportCamera.resetFrame()
+          }
+          addOnIndicatorBearingChangedListener {
+            latestLocation.bearing = it.toFloat()
+            viewportDataSource.onLocationChanged(latestLocation)
+            viewportCamera.resetFrame()
+          }
```
macdrevx, line 145: The data flow here seems a bit manual. Should the viewport camera be able to observe changes to the data source? Similarly, should the data source be supplied with a location producer to which it can subscribe?

macdrevx, line 145: The data flow here seems a bit imperative. Could we make a more declarative API by having the viewport camera observe changes to the data source? Similarly, could the data source be supplied with a location producer to which it can subscribe?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+        }
+      }
+    }
+    setupViewportCamera()
+  }
+
+  @SuppressLint("SetTextI18n")
+  private fun setupViewportCamera() {
+    viewportCamera = mapView.viewport
+    viewportDataSource = MapboxViewportDataSource(mapboxMap, mapboxMap)
```
macdrevx, line 155: Looks kind of silly to need to pass mapboxMap twice.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+          }
+        }
+      }
+    }
+    setupViewportCamera()
+  }
+
+  @SuppressLint("SetTextI18n")
+  private fun setupViewportCamera() {
+    viewportCamera = mapView.viewport
```
macdrevx, line 154: The terminology here feels inconsistent. How do we want developers to refer to this? `viewport` or `viewportCamera`?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+    viewportDataSource = MapboxViewportDataSource(mapboxMap, mapboxMap)
+    viewportCamera.viewportDataSource = viewportDataSource
```
macdrevx, line 156: It seems redundant for this property to be prefixed with the word `viewport` — the association is clear from the context. I'd just make this read `viewportCamera.dataSource = viewportDataSource`

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+    viewportCamera.registerViewportCameraStateChangedObserver { cameraState ->
+      // change title of viewport button depending on the camera state
+      when (cameraState) {
```
macdrevx, line 159: Could we rename `ViewportCameraState` to provide better differentiation from `CameraState`? How about something like `ViewportCameraMode`?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
```diff
@@ -0,0 +1,243 @@
...
+        ViewportCameraState.TransitionToFollowing,
+        ViewportCameraState.Following -> viewportButton.text = OVERVIEW
+
+        ViewportCameraState.TransitionToOverview,
+        ViewportCameraState.Overview,
+        ViewportCameraState.Idle -> viewportButton.text = FOLLOW
```
macdrevx, line 165: I'm wondering whether having these states/modes be hard-coded makes sense. Could we instead come up with a more data-driven approach where we expose API that allows developers to define the rules for a given state and then they can add any number of those states to the viewport camera? Transitions between states could have a default implementation, but could also be modeled explicitly if developers want to fine-tune them.

macdrevx: To make it easy to use, we could provide some pre-defined configurations that correspond to following, overview, and idle.

plugin-viewport/gradle.properties
```diff
@@ -0,0 +1,5 @@
+POM_GROUP_ID=com.mapbox.plugin
+POM_ARTIFACT_ID=maps-viewport
+POM_ARTIFACT_TITLE=The viewport module for the Mapbox Maps SDK Carbon
+POM_DESCRIPTION=The viewport module for the Mapbox Maps SDK Carbon
```
macdrevx, line 4: I think we're trying to avoid Carbon in favor of v10.

macdrevx:
Top feedback here for me is:
* let's explore generalizing the concepts of states and transitions
* let's try making the data flow more declarative

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: This is a left over of my refactoring ViewportCamera to ViewportPlugin. We should keep viewport as the name.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: Yes, as mentioned in the PR description, currently the `MapboxViewportDataSource` is directly accessed and customised in the test app, we need to refactor this class and abstract generic public interface from `MapboxViewportDataSource` and write a MVP tracking modes implementation using these interfaces, and move the custom implementation to the test app.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: agree, we should make the viewport plugin a LocationConsumer and subscribe to the LocationProvider

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: I think maybe we can simply name it to `ViewportState`?

pengdev:
Considering the initial PR has got too big and hard to review, I will merge the initial PR and address the remaining comments in follow-up PRs in the feature branch.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt
pengdev: Addressed in https://github.com/mapbox/mapbox-maps-android/pull/969

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: More class name refactoring is done in https://github.com/mapbox/mapbox-maps-android/pull/969

PR #826, opened 2021-11-05
Add example that uses tilejson through the tileset class.

tobrun:
## Pull request checklist:

### Summary of changes

While validating if something was correctly working with tilejson, I noticed we didn't have any code using the TileSet API in an actual map example. The added example uses OSM based raster tiles through tileset configuration of a RasterSource and visualising it with a RasterLayer. 

image

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1077,6 +1077,18 @@
                 android:value=".ExampleOverviewActivity" />
         </activity>
         <activity
+            android:name=".examples.style.TileJsonActivity"
+            android:description="@string/description_osm"
+            android:exported="true"
+            android:label="@string/activity_osm">
+            <meta-data
+                android:name="@string/category"
+                android:value="@string/category_styles" />
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".ExampleOverviewActivity" />
+        </activity>
+        <activity
             android:name=".TestMapActivity"
             android:exported="true" />
     </application>
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
@@ -1,0 +1,86 @@
+package com.mapbox.maps.testapp.examples.style
+
+import android.os.Bundle
+import android.widget.Toast
+import androidx.appcompat.app.AppCompatActivity
+import com.google.android.material.floatingactionbutton.FloatingActionButton
+import com.mapbox.common.Logger
+import com.mapbox.common.ValueConverter
+import com.mapbox.maps.MapView
+import com.mapbox.maps.MapboxMap
+import com.mapbox.maps.extension.style.layers.addLayer
+import com.mapbox.maps.extension.style.layers.generated.rasterLayer
+import com.mapbox.maps.extension.style.sources.TileSet
+import com.mapbox.maps.extension.style.sources.addSource
+import com.mapbox.maps.extension.style.sources.generated.Scheme
+import com.mapbox.maps.extension.style.sources.generated.rasterSource
+import com.mapbox.maps.testapp.R
+
+/**
+ * Activity showcases usage of TileSet to load a TileJSON compatible configuration as a source.
+ *
+ * This example uses OSM tiles as a raster source and visualises them using a raster layer.
+ */
+class TileJsonActivity : AppCompatActivity() {
+
+  private lateinit var mapboxMap: MapboxMap
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_custom_layer)
+    val mapView: MapView = findViewById(R.id.mapView)
+
+    val tileSet = TileSet.Builder(TILE_JSON_VERSION, listOf(OSM_RASTER_TILE_URL))
+      .name(TILE_JSON_NAME)
+      .description(TILE_JSON_DESCRIPTION)
+      .attribution(TILE_JSON_ATTRIBUTION)
+      .scheme(Scheme.XYZ)
+      .minZoom(TILE_JSON_MIN_ZOOM)
+      .maxZoom(TILE_JSON_MAX_ZOOM)
+      .bounds(MERCATOR_BOUNDS)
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+          val properties = it.getStyleSourceProperties(SOURCE_ID).value!!
+          val propertiesJson = ValueConverter.toJson(properties)
+          Log.v(TAG, propertiesJson)
+          Toast.makeText(this, propertiesJson, Toast.LENGTH_LONG).show()
+        }
+      }
+    }
+  }
+
+  companion object {
```
alexander-kulikovskii, line 69:
```suggestion
  private companion object {
```

app/src/main/AndroidManifest.xml
```diff
@@ -1076,6 +1076,18 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".ExampleOverviewActivity" />
         </activity>
+        <activity
+            android:name=".examples.style.TileJsonActivity"
```
kiryldz, line 1080: Do we need an equivalent for iOS? cc @macdrevx

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+class TileJsonActivity : AppCompatActivity() {
+
+  private lateinit var mapboxMap: MapboxMap
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_custom_layer)
+    val mapView: MapView = findViewById(R.id.mapView)
+
+    val tileSet = TileSet.Builder("2.0.0", listOf("http://tile.openstreetmap.org/{z}/{x}/{y}.png"))
```
pengdev, line 33: nit: move strings to const val

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
```diff
@@ -0,0 +1,74 @@
+package com.mapbox.maps.testapp.examples.style
+
+import android.os.Bundle
+import android.util.Log
```
kiryldz, line 4: Should use common logger instead.

app/src/main/res/values/example_titles.xml
```diff
@@ -84,4 +84,5 @@
     <string name="activity_nine_patch">9-patch image</string>
     <string name="activity_feature_state">Feature state</string>
     <string name="activity_globe">3D globe</string>
+    <string name="activity_osm">Use TileJson</string>
```
kiryldz, line 87: maybe this way it sounds a bit better?
```suggestion
    <string name="activity_osm">TileJson usage</string>
```

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+      it.addLayer(rasterLayer(LAYER_ID, SOURCE_ID) {})
+    }
+
+    // Click on button to print out tile set information
+    findViewById<FloatingActionButton>(R.id.fab).setOnClickListener {
+      if (::mapboxMap.isInitialized) {
+        mapboxMap.getStyle {
+          val properties = it.getStyleSourceProperties(SOURCE_ID).value!!
+          val propertiesJson = ValueConverter.toJson(properties)
+          Log.v(TAG, propertiesJson)
```
kiryldz, line 62: I'd remove log at all or at least make it `info`. Also as stated above we need to use our own logger.

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
```diff
@@ -0,0 +1,74 @@
...
+      .bounds(listOf(-180.0, -85.0, 180.0, 85.0))
+      .center(listOf(11.9, 57.7, 8.0))
+      .build()
+
+    mapboxMap = mapView.getMapboxMap()
+    mapboxMap.loadStyleJson("{}") {
+      it.addSource(
+        rasterSource(SOURCE_ID) {
+          tileSet(tileSet)
+          tileSize(256)
```
kiryldz, line 50: does it mean to be strictly 256? if no perhaps move to constants so that user could play around changing that value

app/src/main/AndroidManifest.xml
tobrun: Yes, atm iOS is missing the tileset model. We will have to add that and back it up with this example

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
alexander-kulikovskii: I can suggest to write own detekt or ktlint rule for checking this

pengdev: Thinking from a customer's perspective, I think we should use `android.util.Log` in the example app, and use common logger in our internal logging in sdk and plugin modules.

app/src/main/AndroidManifest.xml
macdrevx: Please ticket this if needed.

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
tobrun: For now, I'm ok with Log or Logger. I'm not seeing a need for the one or other in the testapp module.
Long term, when we won't emit logs of the common SDK lower as error by default. I would suggest using Log else a user copy/pastes code and it wouldn't see any logs.

app/src/main/java/com/mapbox/maps/testapp/examples/style/TileJsonActivity.kt
tobrun: content might not fully fit the toast, so for convenience also logging it out

kiryldz (approved):

pengdev (approved):

Issue #809, opened 2021-11-03
Scalebar ratio setting not applied correctly


skienzl:
## Environment

- Android OS version: all
- Devices affected: all
- Maps SDK Version: 10.1.0-rc.1

## Observed behavior and steps to reproduce

If you change the ratio of the scalebar plugin the bar and labels get cut off if the ratio is higher than 0.5:

![image](https://user-images.githubusercontent.com/4992387/140040343-98bbaf36-bbc3-4468-a6f6-a95d21c78ba0.png)

## Expected behavior

Apply the correct ratio to the bar and lables

## Notes / preliminary analysis

In the the root case is in `ScaleBarImpl.kt` on line 63 in which the width is calculated with a fixed ratio if 0.5: 

`val width = mapViewWidth / 2`

The possible fix would be to apply the ratio from the settings: 

`val width = mapViewWidth * settings.ratio`

Chaoba:
Thank you @skienzl for reporting this, we will fix this issue soon.

PR #827, opened 2021-11-08
Apply ratio to the width of scalebar

Chaoba:
Fixes: #809 
## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix scalebar ratio setting not applied correctly issue.</changelog>`.

### Summary of changes
Scale bar uses the half of mapview width as its width, so the right part will be cut off if set the ratio bigger than o.5.
This pr applies ratio to scalebar and make the width changes with ratio.
The following picture shows the width of scalebar when setting ratio to 1.0
image

### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
+++ b/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
@@ -60,7 +60,7 @@ class ScaleBarImpl : ScaleBar, View {
     }
 
   override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
-    val width = mapViewWidth / 2
+    val width = mapViewWidth * settings.ratio
     val height = settings.run { textBarMargin + textSize + height + (borderWidth * 2) }
     setMeasuredDimension(width.toInt(), height.toInt())
   }


plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -60,7 +60,7 @@ class ScaleBarImpl : ScaleBar, View {
     }
 
   override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
-    val width = mapViewWidth / 2
+    val width = mapViewWidth * settings.ratio
```
alexander-kulikovskii, line 63: should we add tests?

Chaoba: It's hard to test this part before we have a pixel comparison test.

alexander-kulikovskii: why it's hard to test? mock `settings` and verify that called `setMeasuredDimension` with some values

alexander-kulikovskii (approved):

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
pengdev: we can extract the logic to a method and test that method instead.

pengdev (approved):

PR #834, opened 2021-11-09
Feature: view annotations support

kiryldz:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce view annotation support which allows adding Android views on top of the MapView anchored to geo-point.</changelog>`.

### Summary of changes

Add new feature allowing adding Android [views](https://developer.android.com/reference/android/view/View?hl=en) on top of `MapView` anchored to some geo point.

https://user-images.githubusercontent.com/15800566/142022898-afaf4386-c0ca-4a5e-92d2-fb7755c82a83.mp4

### User impact (optional)

View annotations are controlled with `ViewAnnotationManager` that can be accessed from `MapView`:
```
mapView.viewAnnotationManager
```

Adding / updating / removing view annotation operations are supported.
Various view annotations options are present:
 - bind view annotation to `Point` on the map.
 - relative positioning using `anchor` property which behaves the same as symbol layer `iconAnchor` property.
 - additional X, Y offset when positioning view annotation.
 - allowing view annotations to overlap with other view annotations.
 - selecting view annotation making sure it will be always displayed on top of non-selected view annotations.
 - binding view annotation with some `Feature` by id so that view annotation's visibility is in sync with feature's visibility.


sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -0,0 +1,132 @@
...
+import com.mapbox.geojson.Geometry
+import com.mapbox.geojson.Point
+import com.mapbox.maps.MapView
+import com.mapbox.maps.MapboxExperimental
+import com.mapbox.maps.ViewAnnotationOptions
+
+/**
+ * Manager API to control View Annotations.
+ *
+ * View annotations are Android [View]'s that are drawn on top of the [MapView] and bound to some [Geometry] (only [Point] is supported for now).
```
Chaoba, line 16:
```suggestion
 * View annotations are Android [View]s that are drawn on top of the [MapView] and bound to some [Geometry] (only [Point] is supported for now).
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -0,0 +1,132 @@
...
+   * Find [View] by feature id if it was specified as part of [ViewAnnotationOptions.associatedFeatureId].
+   *
+   * @return [View] if view was found and NULL otherwise.
+   */
+  fun getViewAnnotationByFeatureId(featureId: String): View?
+
+  /**
+   * Find [ViewAnnotationOptions] of view annotation by feature id if it was specified as part of [ViewAnnotationOptions.associatedFeatureId].
+   *
+   * @return [ViewAnnotationOptions] if view annotation was found and NULL otherwise.
```
Chaoba, line 122:
```suggestion
   * @return [ViewAnnotationOptions] if view annotation options was found and NULL otherwise.
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+        "Trying to add view annotation that was already added before! " +
+          "Please consider deleting annotation view ($view) beforehand."
+      )
+    }
+    validateOptions(options)
+    prepareViewAnnotation(view, options)
+  }
+
+  override fun removeViewAnnotation(view: View): Boolean {
+    val id = idLookupMap[view] ?: return false
```
Chaoba, line 77: `remove` method will return the previous value associated with key, or null if there was no mapping for key,
so here we can call remove directly
`val id = idLookupMap.remove(view)`

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+    return true
+  }
+
+  override fun updateViewAnnotation(
+    view: View,
+    options: ViewAnnotationOptions,
+  ): Boolean {
+    val id = idLookupMap[view] ?: return false
+    annotationMap[id]?.let {
+      it.handleVisibilityAutomatically = options.visible == null
```
Chaoba, line 92:
```suggestion
      it.handleVisibilityAutomatically = (options.visible == null)
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+    val (_, options) = findByFeatureId(featureId)
+    return options
+  }
+
+  override fun getViewAnnotationOptionsByView(view: View): ViewAnnotationOptions? {
+    val id = idLookupMap[view] ?: return null
+    val options = mapboxMap.getViewAnnotationOptions(id)
+    if (options.isValue) {
+      return options.value
+    }
```
Chaoba, line 113: Log the error message in else block?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+
+  private fun prepareViewAnnotation(inflatedView: View, options: ViewAnnotationOptions): View {
+    val inflatedViewLayout = inflatedView.layoutParams as FrameLayout.LayoutParams
+    val updatedOptions = options.toBuilder()
+      .width(options.width ?: inflatedViewLayout.width)
+      .height(options.height ?: inflatedViewLayout.height)
+      .build()
+    val viewAnnotation = ViewAnnotation(
+      view = inflatedView,
+      handleVisibilityAutomatically = options.visible == null,
```
Chaoba, line 149:
```suggestion
      handleVisibilityAutomatically = (options.visible == null),
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+      viewLayoutParams = inflatedViewLayout,
+    )
+    val globalLayoutListener = ViewTreeObserver.OnGlobalLayoutListener {
+      if (viewAnnotation.handleVisibilityAutomatically) {
+        val isVisibleNow = inflatedView.visibility == View.VISIBLE
```
Chaoba, line 154:
```suggestion
        val isVisibleNow = (inflatedView.visibility == View.VISIBLE)
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,253 @@
...
+    idsToAddSet.clear()
+    idsToDeleteSet.clear()
+
+    positionDescriptorList.forEachIndexed { i, descriptor ->
+      positionDescriptorMap[descriptor.identifier] =
+        ViewPosition(
+          leftTopCoordinate = descriptor.leftTopCoordinate,
+          width = descriptor.width,
+          height = descriptor.height,
+          zIndex = ViewPlugin.VIEW_PLUGIN_Z_TRANSLATION - 1f + i.toFloat() / positionDescriptorList.size
```
Chaoba, line 198: Can you elaborate on why `zIndex` is calculated in this way?

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,197 @@
...
+        styleExtension = prepareStyle(Style.MAPBOX_STREETS, bitmap)
+      ) {
+        style = it
+        addOnMapClickListener(this@ViewAnnotationShowcaseActivity)
+        addOnMapLongClickListener(this@ViewAnnotationShowcaseActivity)
+        binding.fabStyleToggle.setOnClickListener {
+          when (getStyle()?.styleURI) {
+            Style.MAPBOX_STREETS -> loadStyle(prepareStyle(Style.SATELLITE_STREETS, bitmap))
+            Style.SATELLITE_STREETS -> loadStyle(prepareStyle(Style.MAPBOX_STREETS, bitmap))
+          }
```
alexander-kulikovskii, line 74:
```suggestion
          val newStyleUri = when (getStyle()?.styleURI) {
            Style.MAPBOX_STREETS -> Style.SATELLITE_STREETS
            Style.SATELLITE_STREETS -> Style.MAPBOX_STREETS
          }
          loadStyle(prepareStyle(newStyleUri, bitmap))
```

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,197 @@
...
+    mapboxMap = binding.mapView.getMapboxMap().apply {
+      loadStyle(
+        styleExtension = prepareStyle(Style.MAPBOX_STREETS, bitmap)
+      ) {
+        style = it
+        addOnMapClickListener(this@ViewAnnotationShowcaseActivity)
+        addOnMapLongClickListener(this@ViewAnnotationShowcaseActivity)
+        binding.fabStyleToggle.setOnClickListener {
+          when (getStyle()?.styleURI) {
+            Style.MAPBOX_STREETS -> loadStyle(prepareStyle(Style.SATELLITE_STREETS, bitmap))
```
alexander-kulikovskii, line 72:
```suggestion
            MAPBOX_STREETS -> loadStyle(prepareStyle(SATELLITE_STREETS, bitmap))
```
use import for this styles

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,197 @@
...
+    return true
+  }
+
+  override fun onMapClick(point: Point): Boolean {
+    mapboxMap.queryRenderedFeatures(
+      mapboxMap.pixelForCoordinate(point),
+      RenderedQueryOptions(listOf(LAYER_ID), null)
+    ) {
+      if (it.isValue && it.value?.size!! > 0) {
+        it.value?.get(0)?.feature?.let { feature ->
```
alexander-kulikovskii, line 112:
```suggestion
     getFirstFeatureOrNull?. ...

```
how about small extension?

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,197 @@
...
+    return true
+  }
+
+  private fun reverseViewVisibility(view: View) {
+    if (view.visibility == View.VISIBLE) {
+      view.visibility = View.GONE
+    } else {
+      view.visibility = View.VISIBLE
+    }
+  }
```
alexander-kulikovskii, line 130:
```suggestion
  private fun View.reverseVisibility() {
    visibility = if (visibility == View.VISIBLE) View.GONE else View.VISIBLE
  }
```

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
```diff
@@ -374,6 +375,14 @@ internal class NativeMapImpl(private val map: MapInterface) :
     map.queryRenderedFeatures(pixel, options, callback)
   }
 
+  override fun queryRenderedFeatures(
+    geometry: RenderedQueryGeometry,
+    options: RenderedQueryOptions,
+    callback: QueryFeaturesCallback
```
alexander-kulikovskii, line 381: nit: 

```suggestion
    callback: QueryFeaturesCallback,
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,225 @@
...
+      annotation.view.viewTreeObserver.removeOnGlobalLayoutListener(annotation.globalLayoutListener)
+      mapView.removeView(annotation.view)
+    }
+    currentViewsDrawnMap.clear()
+    annotationMap.clear()
+    idLookupMap.clear()
+  }
+
+  private fun validateOptions(options: ViewAnnotationOptions) {
+    checkNotNull(options.geometry) { "Geometry can not be null!" }
```
endanke, line 123: Should we maybe check that the geometry is a point?

endanke: Discussed on Slack that the error is handled on GL-Native side and we have to propagate it through our APIs.

kiryldz: tracking in https://github.com/mapbox/mapbox-maps-android/pull/850

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
kiryldz: using completely different logic now, PTAL

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
kiryldz: this will require explicit `else` branch, let's leave as is

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
kiryldz: not using it anywhere else in that file, so let's be consistent

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
kiryldz: across the project `Style.` is used, let's be consistent

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
kiryldz: this was revisited

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManagerImpl.kt
kiryldz: thanks

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,201 @@
...
+
+    val bitmap = BitmapFactory.decodeResource(resources, R.drawable.blue_marker_view)
+    markerWidth = bitmap.width
+    markerHeight = bitmap.height
+
+    mapboxMap = binding.mapView.getMapboxMap().apply {
+      loadStyle(
+        styleExtension = prepareStyle(Style.MAPBOX_STREETS, bitmap)
+      ) {
+        style = it
```
pengdev, line 68: I would suggest not to cache style within the activity, instead, use `mapboxMap.getStyle { style -> }` when you need it. And seems there's only 1 place throughout the activity that uses it.

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt
```diff
@@ -0,0 +1,201 @@
...
+    onFeatureClicked: (Feature) -> Unit
+  ) {
+    if (expected.isValue && expected.value?.size!! > 0) {
+      expected.value?.get(0)?.feature?.let { feature ->
+        onFeatureClicked.invoke(feature)
+      }
+    }
+  }
+
+  private fun View.reverseViewVisibility() {
```
pengdev, line 132:
```suggestion
  private fun View.toggleViewVisibility() {
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -0,0 +1,135 @@
...
+   * @param options view annotation options
+   * @param asyncInflater instance of [AsyncLayoutInflater] provided by the user
+   * @param asyncInflateCallback callback triggered when [View] is inflated.
+   *
+   * @throws [RuntimeException] if options did not include geometry.
+   */
+  fun addViewAnnotation(
+    @LayoutRes resId: Int,
+    options: ViewAnnotationOptions,
+    asyncInflater: AsyncLayoutInflater,
```
pengdev, line 72: seems the asyncInflater is required to use the view annotations, why did we use compileOnly dependency on it?

sdk/build.gradle.kts
```diff
@@ -72,6 +72,7 @@ dependencies {
   api(project(":plugin-annotation"))
   api(project(":extension-localization"))
   api(project(":plugin-lifecycle"))
+  compileOnly(Dependencies.asyncInflater)
```
pengdev, line 75: why compileOnly?

kiryldz: thanks for the question!
we want to support in our API ability to inflate views async however at the same time we don't want to add even small binary size to our SDK module.
that's why we enforce users to provide async inflater instance as parameter in our API (which is possible only if they indeed have async layout inflater dependency) but our code recognizes it because of `compileOnly` (but we do not provide dependency impl as part of our SDK).

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
kiryldz: see https://github.com/mapbox/mapbox-maps-android/pull/834#discussion_r750518789

sdk/build.gradle.kts
pengdev: do we have some numbers on how much binary size the async inflater takes?
Imho if our code relies on it to work(and it's not optional in the public interface) we should use implement dependency. It's a bit different from other cases like LocationEngine, where GoogleLocationService is optional and we have a safe fallback. wdyt? cc @mapbox/maps-android

kiryldz: @pengdev I don't agree here.
Although we don't provide impl having API the way it is right now:
```
fun addViewAnnotation(
    @LayoutRes resId: Int,
    options: ViewAnnotationOptions,
    asyncInflater: AsyncLayoutInflater,
    asyncInflateCallback: (View) -> Unit
  )
```
actually _enforces_ user to add this dependency to call this API. There's no even a concept of safe fallback here because this API could not simply be called by user if he does not import dependency.

At the same time it's convenient possibility to move inflating off the main thread for advanced user cases with bunch of view annotations.

tobrun: Cool idea of saving binary size but I do want to make sure we provide a good developer experience. What happens if the developer calls into the codepath but didn't add the depdency on their end? ClassNotFoundException? 
If yes, I do think we can optimize the experience by catching this exception and throwing a new depedency missing exception that clearly indicates what a user can do to resolve this issue.

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -34,7 +35,18 @@ import com.mapbox.maps.renderer.egl.EGLCore
...
+
+  private var isViewAnnotationManagerInitialized = false
+  /**
+   * Get view annotation manager instance to add / update / remove view annotations
+   * represented as Android views.
+   */
+  val viewAnnotationManager: ViewAnnotationManager by lazy {
+    isViewAnnotationManagerInitialized = true
+    ViewAnnotationManagerImpl(this)
+  }
```
tobrun, line 49: Kotlin language feature of lazy don't require you to manage the initalization state yourself.
See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-lazy/is-initialized.html

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,232 @@
...
+import com.mapbox.maps.viewannotation.ViewAnnotation
+import com.mapbox.maps.viewannotation.ViewAnnotationManager
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.collections.HashMap
+
+internal class ViewAnnotationManagerImpl(
+  private val mapView: MapView
+) : ViewAnnotationManager, ViewAnnotationPositionsUpdateListener {
+
+  private val mapboxMap: MapboxMap = mapView.getMapboxMap()
```
tobrun, line 20: mapboxMap seems to be a hidden dependency of ViewAnnotationManagerImpl. 
Should it be a constructor arg instead? Should make testing also easier

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,232 @@
...
+  private val viewPlugins = mapView.mapController.pluginRegistry.viewPlugins
```
tobrun, line 21: same

sdk/build.gradle.kts
yunikkk: IMHO well written javadoc here should prevent the possible confusion
Just need to check that it indeed won't compile without explicit dependency declaration on used end instead of runtime exception

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,232 @@
...
+
+  init {
+    mapView.requestDisallowInterceptTouchEvent(false)
+    mapboxMap.setViewAnnotationPositionsUpdateListener(this)
+  }
+
+  private val annotationMap = ConcurrentHashMap<String, ViewAnnotation>()
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal val idLookupMap = ConcurrentHashMap<View, String>()
+  private val hiddenViewMap = ConcurrentHashMap<View, Float>()
```
tobrun, line 31: - We have a lot of HashMaps here, can we refactor to a single? by introducing a model that encapsulates these states that are managed independently? 
- Why do we have concurrent? is this solution multi threaded?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -0,0 +1,232 @@
...
+      currentViewsDrawnMap[it.identifier] = it.leftTopCoordinate
+    }
+  }
+
+  private inline fun <reified V> getValue(expected: Expected<String, V>): V? {
+    if (expected.isError) {
+      throw RuntimeException(expected.error)
+    }
+    return expected.value
+  }
```
tobrun, line 231: Maybe we need to support this functionality in the base class of Value.
Should we expose an extension function for now? `Value.valueOrThrow()` signature

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotation.kt
```diff
@@ -0,0 +1,41 @@
...
+) {
+  /**
+   * String id needed to call functions from gl-native
+   */
+  val id: String = (VIEW_ANNOTATION_CURRENT_ID++).toString()
+
+  /**
+   * Global layout listener to control positioning on the screen based on view's actual visibility.
+   */
+  var globalLayoutListener: ViewTreeObserver.OnGlobalLayoutListener? = null
```
tobrun, line 36: do we need a GlobalLayoutListener for each annotation independently? 
why not have 1 in the manager that manages all the annotations separately.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotation.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  private companion object {
+    var VIEW_ANNOTATION_CURRENT_ID = 42
```
tobrun, line 39: 42  :sweat_smile:

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
kiryldz: extension functions are sadly not available for Java classes (while `Value` is generated Java class).

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
kiryldz: can't get what hidden dependency is here tbh, in our terms `MapView` almost ~ to `MapboxMap`. Also tests are already written and I did not see this as an issue tbh 😄 
can you add more clarity here?

sdk/src/main/java/com/mapbox/maps/MapView.kt
kiryldz: I was afraid for some reason that calling this method when destroying map view will first create an instance and then destroy it but seems that it's not true, will apply your solution here.

sdk/build.gradle.kts
kiryldz: Discussed offline, will go with current solution.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
kiryldz: same = move to ctor?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
kiryldz: - user operates with actual views so we want O(1) to actually lookup internal structure for view annotation (we use `idLookupMap` for that). While core callbacks operate with ids and not views so we do use `annotationMap` with different key to also go with O(1) complexity. `hiddenViewMap` is a separate instance needed for non-trivial solution about view visibility, see https://github.com/mapbox/mapbox-maps-android/pull/847 for visuals.
 - we use separate thread but collection may be modified while we're iterating so to avoid concurrent modification exception we go with this impl

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotation.kt
kiryldz: that's a valid point, I'll create follow-up task about it, changing it now may be a bit risky

kiryldz: https://github.com/mapbox/mapbox-maps-android/issues/862

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotation.kt
kiryldz: why not?

sdk/src/main/java/com/mapbox/maps/MapView.kt
tobrun: haha same, but you do need to use something as:

```
val messageBrokerDelegate = lazy { MessageBroker() }
val messageBroker by messageBrokerDelegate

if(messageBrokerDelegate.isInitialized())
    ...
```

from https://stackoverflow.com/questions/42522739/kotlin-check-if-lazy-val-has-been-initialised

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
tobrun: :shipit:

yunikkk: IMO no difference - could be mocked via `mapView` instead

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
tobrun: yeah, :shipit:

sdk/src/main/java/com/mapbox/maps/MapView.kt
kiryldz: Applied, thanks, works like a charm!

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -34,7 +35,15 @@ import com.mapbox.maps.renderer.egl.EGLCore
  * and for ensuring your use adheres to the relevant terms of use.
  */
 open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
-  private var mapController: MapController
+  internal var mapController: MapController
+    private set
+
+  private val viewAnnotationManagerDelegate = lazy { ViewAnnotationManagerImpl(this) }
```
Chaoba, line 41: Why need a delegate here?

kiryldz: Why do we need to init `ViewAnnotationManager` at all if user does not use it? 😄

Chaoba (approved):
LGTM

tobrun (approved):

ank27 (approved):

yunikkk (approved):

SickZauerKrautzz89:
T

PR #869, opened 2021-11-18
[gestures] Refactor fling gestures test to be parameterised tests.

pengdev:

## Pull request checklist:
    
### Summary of changes

This PR refactors the fling gesture tests to be parameterised tests.

### DIFF
--- a/plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
+++ b/plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
@@ -19,6 +19,8 @@ import com.mapbox.maps.ScreenCoordinate
 import com.mapbox.maps.plugin.Plugin
 import com.mapbox.maps.plugin.ScrollMode
 import com.mapbox.maps.plugin.animation.CameraAnimationsPlugin
+import com.mapbox.maps.plugin.animation.MapAnimationOptions
+import com.mapbox.maps.plugin.animation.MapAnimationOwnerRegistry
 import com.mapbox.maps.plugin.delegates.*
 import com.mapbox.maps.plugin.gestures.generated.GesturesAttributeParser
 import com.mapbox.maps.plugin.gestures.generated.GesturesSettings
@@ -787,8 +789,9 @@ class GesturesPluginTest {
 class FlingGestureTest(
   private val targetPitch: Double,
   private val targetScrollMode: ScrollMode,
-  private val velocity: Pair<Float, Float>,
-  private val expectedCoordinate: ScreenCoordinate
+  private val targetVelocity: Pair<Float, Float>,
+  private val expectedCoordinate: ScreenCoordinate,
+  private val expectedFlingDuration: Long
 ) {
   private val context: Context = mockk(relaxed = true)
   private val attrs: AttributeSet = mockk(relaxUnitFun = true)
@@ -803,6 +806,9 @@ class FlingGestureTest(
 
   private val typedArray: TypedArray = mockk(relaxed = true)
   private val pack = "com.mapbox.maps"
+
+  private val motionEvent = mockk<MotionEvent>()
+  private val mapAnimationOptionsSlot = slot<MapAnimationOptions>()
 
   @Before
   fun prepare() {
@@ -834,10 +840,7 @@ class FlingGestureTest(
     presenter.bind(context, gesturesManager, attrs, 1f)
     presenter.onDelegateProvider(mapDelegateProvider)
     presenter.initialize()
-  }
 
-  @Test
-  fun testFling() {
     every { mapCameraManagerDelegate.cameraState } returns CameraState(
       Point.fromLngLat(0.0, 0.0),
       EdgeInsets(0.0, 0.0, 0.0, 0.0),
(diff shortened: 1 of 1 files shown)


Chaoba (approved):

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +780,146 @@ class GesturePluginTest {
...
+    presenter = GesturesPluginImpl(context, attrs, mockk(relaxed = true))
+
+    presenter.bind(context, gesturesManager, attrs, 1f)
+    presenter.onDelegateProvider(mapDelegateProvider)
+    presenter.initialize()
+  }
+
+  @Test
+  fun testFling() {
+    every { mapCameraManagerDelegate.cameraState } returns CameraState(
```
alexander-kulikovskii, line 841: can be lines 841-857 moved to setUp function. or maybe move to small private init function for better code reading

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +780,146 @@ class GesturePluginTest {
...
+    every { motionEvent.x } returns 0.0f
+    every { motionEvent.y } returns 0.0f
+    val result = presenter.handleFlingEvent(motionEvent, mockk(), velocity.first, velocity.second)
+    verify {
+      mapCameraManagerDelegate.getDragCameraOptions(
+        ScreenCoordinate(0.0, 0.0),
+        expectedCoordinate
+      )
+    }
+    verify { cameraAnimationsPlugin.easeTo(any(), any()) }
```
alexander-kulikovskii, line 865: do you know exact values for this function? better not to use `any()`

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +780,146 @@ class GesturePluginTest {
...
+        attrs,
+        any()
+      )
+    } returns GesturesSettings()
+
+    every { context.obtainStyledAttributes(any(), any(), 0, 0) } returns typedArray
+    every { context.packageName } returns pack
+    every { typedArray.getString(any()) } returns "pk.token"
+    every { typedArray.getBoolean(any(), any()) } returns true
+    every { typedArray.getDimension(any(), any()) } returns 10.0f
```
alexander-kulikovskii, line 822: do you really need all this `every` calls? it's slow down every test. better to remove useless

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +780,146 @@ class GesturePluginTest {
...
+        ScreenCoordinate(5.435095880747194, 5.095402388200494)
+      ),
+      arrayOf(
+        0.0,
+        ScrollMode.VERTICAL,
+        Pair(FLING_VELOCITY, FLING_VELOCITY),
+        ScreenCoordinate(0.0, FLING_DISPLACEMENT)
+      ),
+      arrayOf(
+        0.0,
```
alexander-kulikovskii, line 916: minimum value `0.0`? what is the maximum potential value?

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
pengdev: these values would depend on the mock value of 
```
getDragCameraOptions(
        ScreenCoordinate(0.0, 0.0),
        expectedCoordinate
      )
```

pengdev: added the mock value to easeTo function and also added check for animation duration.

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
pengdev: The test execution covers these code paths, otherwise the test will fail with unmockked methods.

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
pengdev: This is the pitch value, there's a constant value defined in GesturesConstants.kt:
```
internal const val MINIMUM_PITCH = 0.0
internal const val MAXIMUM_PITCH = 85.0
```

alexander-kulikovskii: but we have 80 as max value
maybe not super important to test it. up to you.
just want double check it :slightly_smiling_face:

pengdev: In the tests we covered no pitch(0) which is most frequent use case, low pitch(<60) and high pitch(between 60 and 85), which matches the internal logic where we applied different pitch factors for, so I think it should be good enough to cover these pitch bands 🙂

alexander-kulikovskii (approved):

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +782,166 @@ class GesturePluginTest {
...
     return MotionEvent.obtain(200, 500, action, 15.0f, 10.0f, 0)
   }
+}
+
+@RunWith(ParameterizedRobolectricTestRunner::class)
+@LooperMode(LooperMode.Mode.PAUSED)
+class FlingGestureTest(
+  private val targetPitch: Double,
+  private val targetScrollMode: ScrollMode,
+  private val velocity: Pair<Float, Float>,
```
kiryldz, line 792:
```suggestion
  private val targerVelocity: Pair<Float, Float>,
```
for consistency?

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +782,166 @@ class GesturePluginTest {
...
+      )
+    } returns CameraOptions.Builder().build()
+    presenter.updateSettings { scrollMode = targetScrollMode }
+    every { motionEvent.x } returns 0.0f
+    every { motionEvent.y } returns 0.0f
+  }
+
+  @Test
+  fun testFling() {
+
```
kiryldz, line 864: nit: empty line

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1017,8 +782,166 @@ class GesturePluginTest {
...
+      ),
+      arrayOf(
+        55.0,
+        ScrollMode.HORIZONTAL_AND_VERTICAL,
+        Pair(800f, 750f),
+        ScreenCoordinate(51.61290322580645, 48.38709677419355),
+        70L
+      ),
+      arrayOf(
+        80.0,
```
kiryldz, line 923: do we need a test for pitch higher than max pitch? or is it impossible?

kiryldz (approved):
Several minors but in general LGTM

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
pengdev: With current gestures implementation, it's impossible to set pitch over 85 using gestures.
And I also double checked if user could set the camera pitch to 85.0 programmatically, and seems the gl-native clamped to the pitch to a valid range(maximum 85.0) as well. So I think it's safe here.

PR #875, opened 2021-11-19
[Gesture] Update default shove gesture angle

ank27:

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix an issue where shove gesture was not detected when angle between touch points are not horizontal.</changelog>`.

### Summary of changes
Update maximum allowed angle between touch points to qualify it as a shove gesture.

  | Before | After |
  | ----- | ----- |
  | <video src="https://user-images.githubusercontent.com/11589497/142613735-83b9b8c5-4c61-4382-9e37-bb512ca6ca17.mp4" width = 250/> | <video src="https://user-images.githubusercontent.com/11589497/142613750-315b186c-8221-44d4-b28c-432c5bc78fce.mp4" width = 250/> |

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/GesturesActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/GesturesActivity.kt
@@ -74,8 +74,7 @@ class GesturesActivity : AppCompatActivity() {
   private fun initializeMap() {
     gesturesPlugin = binding.mapView.gestures
     gesturesManager = gesturesPlugin.getGesturesManager()
-    // enable map pitch at higher touch point angle (default 20.0f).
-    gesturesManager.shoveGestureDetector.maxShoveAngle = 60.0f
+
     val layoutParams = binding.recycler.layoutParams as RelativeLayout.LayoutParams
     layoutParams.height = (binding.mapView.height / 1.75).toInt()
     layoutParams.width = binding.mapView.width / 3
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/GesturesActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/GesturesActivity.kt
@@ -74,7 +74,8 @@ class GesturesActivity : AppCompatActivity() {
   private fun initializeMap() {
     gesturesPlugin = binding.mapView.gestures
     gesturesManager = gesturesPlugin.getGesturesManager()
-
+    // enable map pitch at higher touch point angle (default 20.0f).
+    gesturesManager.shoveGestureDetector.maxShoveAngle = 60.0f
     val layoutParams = binding.recycler.layoutParams as RelativeLayout.LayoutParams
     layoutParams.height = (binding.mapView.height / 1.75).toInt()
     layoutParams.width = binding.mapView.width / 3
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
@@ -231,7 +231,8 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
         scaleLongPressSet
       )
     }
-    gesturesManager.rotateGestureDetector.angleThreshold = 3.0f
+    gesturesManager.rotateGestureDetector.angleThreshold = ROTATION_ANGLE_THRESHOLD
+    gesturesManager.shoveGestureDetector.maxShoveAngle = MAX_SHOVE_ANGLE
     this.gesturesManager = gesturesManager
   }
 
@@ -1642,5 +1643,10 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
   override fun initialize() {
     initializeGesturesManager(gesturesManager, true)
     initializeGestureListeners(context, true)
+  }
+
+  private companion object {
+    const val ROTATION_ANGLE_THRESHOLD = 3.0f
+    const val MAX_SHOVE_ANGLE = 45.0f
   }
 }
--- a/plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
+++ b/plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
@@ -782,6 +782,25 @@ class GesturesPluginTest {
   fun obtainMotionEventActionLater(action: Int): MotionEvent {
     return MotionEvent.obtain(200, 500, action, 15.0f, 10.0f, 0)
   }
+
+  @Test
+  fun verifyDefaultShoveGestureAngle() {
+    verify {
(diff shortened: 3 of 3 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/GesturesActivity.kt
```diff
@@ -74,7 +74,8 @@ class GesturesActivity : AppCompatActivity() {
   private fun initializeMap() {
     gesturesPlugin = binding.mapView.gestures
     gesturesManager = gesturesPlugin.getGesturesManager()
-
+    // enable map pitch at higher touch point angle (default 20.0f).
+    gesturesManager.shoveGestureDetector.maxShoveAngle = 60.0f
```
kiryldz, line 78: Taking into account it's example activity - let's make UI for it and put it under three dots settings (located in top-right screen angle).

tobrun: The default that iOS allows to use is 45 degrees, any reason for using 20?

ank27: Yeah. default comes from upstream. Do we want to enable the default value to 45 degrees for parity? (then I will also update in `GesturePluginImpl` as well).

kiryldz: technically it's a breaking change in terms of behaviour. 
however if I get logic correctly we may consider it a bug and change it.

pengdev: I think it's more like a bug, and I think we can change the default to 45 degrees in GesturesPluginImpl.
As this is already configurable through `gesturesManager.shoveGestureDetector.maxShoveAngle`, would be good to document it in Gestures Guide(gesture customisation section)

ank27: Yeah make sense. If we are changing the default to 45 degrees, then I think we don't need to add UI in `GestureActivity`. we can just update the gesture guide stating that pitch behavior can be changed with `maxShoveAngle`.  wdyt ?

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -232,6 +232,7 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
       )
     }
     gesturesManager.rotateGestureDetector.angleThreshold = 3.0f
+    gesturesManager.shoveGestureDetector.maxShoveAngle = 45.0f
```
pengdev, line 235: could we add a unit test?

pengdev (approved):
Great work!

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: could we move 45.0f and 3.0f to constants, mark them as internal and make unit test easier?

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -879,6 +879,11 @@ class FlingGestureTest(
     assertEquals(expectedFlingDuration, mapAnimationOptionsSlot.captured.duration)
   }
 
+  @Test
```
kiryldz, line 882: Add / update test for `ROTATION_ANGLE_THRESHOLD` constant as well?

ank27: it's not in the scope of this pr, but yeah I can add it. :)

kiryldz (approved):
LGTM!

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -782,6 +782,21 @@ class GesturesPluginTest {
   fun obtainMotionEventActionLater(action: Int): MotionEvent {
     return MotionEvent.obtain(200, 500, action, 15.0f, 10.0f, 0)
   }
+
+  @Test
+  fun verifyDefaultShoveGestureAngle() {
+    verify {
+      gesturesManager.shoveGestureDetector.maxShoveAngle = GesturesPluginImpl.MAX_SHOVE_ANGLE
```
alexander-kulikovskii, line 789: do you want to check values or only that you set ANY value?
better to make `MAX_SHOVE_ANGLE` from GesturesPluginImpl private in companion object
and add same comp object to this test class

```
gesturesManager.shoveGestureDetector.maxShoveAngle = MAX_SHOVE_ANGLE // from test class obj
```

because when will you change any of this values you will not detect such changes in tests.

alexander-kulikovskii: even you can change type of this value and tests will not detect it

ank27: this test is to validate the default angle set in GesturePluginImpl class. that should be a constant. Although, I get your point that this can also have different values (set programatically). That should be handled in different test imo. I can probably add another test to check that condition too.

alexander-kulikovskii: As for me, default value is a business logic. You can try to change it in comp object and tests will passed

ank27: Yeah, true. made the changes 👍

alexander-kulikovskii (approved):

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1643,4 +1644,9 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
     initializeGesturesManager(gesturesManager, true)
     initializeGestureListeners(context, true)
   }
+
+  companion object {
```
kiryldz, line 1648: mark this companion object as private and rebase your PR on latest `main` to make CI pass.

ank27: all the variables inside this is private,  so i hope that shouldn't affect anything. yeah will rebase :)

alexander-kulikovskii: probably it can be visible as empty comp object

Chaoba (approved):

PR #890, opened 2021-11-23
Revert Dokka to 1.4.10.2; only generate docs for stable releases.

Chaoba:

## Pull request checklist:
    
### Summary of changes
This PR revert dokka to 1.4.10.2 so that the generated docs will keep the original structure.
Also update the CI config, now only the stable release can trigger the job of generating docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,7 +181,6 @@ commands:
             if [[ $CIRCLE_TAG == android-v* ]]; then
               python3 scripts/generate-docs-start-pipeline.py --token ${MOBILE_METRICS_TOKEN} --origin-slug mapbox/mapbox-maps-android --target-slug mapbox/mapbox-maps-android-internal --release_tag $CIRCLE_TAG
             fi
-
   track-metrics:
     steps:
       - unless:
--- a/buildSrc/src/main/kotlin/Project.kt
+++ b/buildSrc/src/main/kotlin/Project.kt
@@ -71,7 +71,7 @@ object Versions {
   const val pluginAndroidGradle = "4.0.1"
   const val pluginKotlin = "1.4.0"
   const val pluginLicense = "0.8.5"
-  const val pluginDokka =  "1.4.32"
+  const val pluginDokka =  "1.4.10.2"
   const val pluginJacoco = "0.2"
   const val mapboxAccessToken="0.2.1"
   const val mapboxSdkRegistry="0.4.0"
--- a/scripts/update-android-docs.sh
+++ b/scripts/update-android-docs.sh
@@ -135,15 +135,17 @@ create_pull_request "Add ${MAPS_SDK_VERSION} API documentation." $BRANCH_WITH_DO
 prepare_branch_with_empty $MAPS_SDK_VERSION
 create_pull_request "Trigger ${MAPS_SDK_VERSION} deploy." $BRANCH_WITH_DOCUMENTATION
 
-# Update config files in Android Docs Repo.
-clone_android_docs_repo
-update_constants_and_map_version_numbers $MAPS_SDK_VERSION
-prepare_android_docs_branch $MAPS_SDK_VERSION
-cd $ANDROID_DOCS_DIRECTORY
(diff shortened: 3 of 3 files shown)


.circleci/config.yml
```diff
@@ -178,10 +178,9 @@ commands:
       - run:
           name: Generate android docs and create PR for review.
           command: |
-            if [[ $CIRCLE_TAG == android-v* ]]; then
+            if [[ $CIRCLE_TAG == android-v* ]] && [[$CIRCLE_TAG != *alpha*]] && [[$CIRCLE_TAG != *beta*]] && [[$CIRCLE_TAG != *rc*]]; then
```
kiryldz, line 181: does it mean we won't generate API docs for alpha, beta and rc?
we should generated API docs but we should not bump version in android-docs repo.

Chaoba: I see, we should config it in script not here.

kiryldz (approved):
LGTM, great job @Chaoba!

buildSrc/src/main/kotlin/Project.kt
```diff
@@ -71,7 +71,7 @@ object Versions {
   const val pluginAndroidGradle = "4.0.1"
   const val pluginKotlin = "1.4.0"
   const val pluginLicense = "0.8.5"
-  const val pluginDokka =  "1.4.32"
+  const val pluginDokka =  "1.4.10.2"
```
yunikkk, line 74: BTW what's the issue with the updated Dokka? Is there a bug or structure is changed by intent? 
Anyway - would be nice to add a comment to not update to 1.4.32

pengdev (approved):

scripts/update-android-docs.sh
```diff
@@ -135,15 +135,17 @@ create_pull_request "Add ${MAPS_SDK_VERSION} API documentation." $BRANCH_WITH_DO
...
-clone_android_docs_repo
-update_constants_and_map_version_numbers $MAPS_SDK_VERSION
-prepare_android_docs_branch $MAPS_SDK_VERSION
-cd $ANDROID_DOCS_DIRECTORY
-create_pull_request "Carbon Maps SDK bump to ${MAPS_SDK_VERSION}" $BRANCH_WITH_DOCUMENTATION
-# Rollback the remote url when run the script locally
-git remote set-url origin git@github.com:mapbox/android-docs.git
-cd -
+# Update config files in Android Docs Repo for stable release.
+if [[ $MAPS_SDK_VERSION != *beta* ]] && [[ $MAPS_SDK_VERSION != *alpha* ]] && [[ $MAPS_SDK_VERSION != *rc* ]]; then
```
alexander-kulikovskii, line 139:
```suggestion
if [[ $MAPS_SDK_VERSION !=~ ^(*beta*|*alpha*|*rc*)$ ]]; then
```
but check syntax

buildSrc/src/main/kotlin/Project.kt
Chaoba: The file structure would be changed with 1.4.32 and we can't deploy it. For more detail please refer to the linked ticekt.

scripts/update-android-docs.sh
Chaoba: The script has been tested and worked well, so I would like to keep it.

buildSrc/src/main/kotlin/Project.kt
ank27: Yeah, I checked that there is a breaking change in the dokka task that breaks the structure in [v1.4.30 ](https://github.com/Kotlin/dokka/releases/tag/v1.4.30). we do use `dokkaHtml` to generate the source sets which is changed to `dokkaHtmlPartial`. that might be the reason why 1.4.32 doesn't work properly.

ank27 (approved):

buildSrc/src/main/kotlin/Project.kt
Chaoba: No, we will update it in the future, I have created to ticket for it.

yunikkk: Ok then!

PR #932, opened 2021-12-01
Add startDelay property for MapAnimationOptions.kt

Chaoba:
### Summary of changes
This PR adds startDelay property for MapAnimationOptions.kt and users can delay the high-level animation with it.

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce ability to specify startDelay property as part of mapAnimationOptions for high-level animations.</changelog>`.

### DIFF
--- a/sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
+++ b/sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
@@ -18,6 +18,10 @@ class MapAnimationOptions private constructor(
    */
   val duration: Long?,
   /**
+   * The amount of time, in milliseconds, to delay starting the animation after animation start.
+   */
+  val startDelay: Long?,
+  /**
    * The animation interpolator.
    * If not set explicitly default interpolator will be taken (CameraAnimatorsFactory.DEFAULT_INTERPOLATOR if using plugin-animation).
    */
@@ -44,6 +48,11 @@ class MapAnimationOptions private constructor(
     private var duration: Long? = null
 
     /**
+     * The amount of time, in milliseconds, to delay starting the animation after animation start.
+     */
+    private var startDelay: Long? = null
+
+    /**
      * The animation interpolator.
      */
     private var interpolator: TimeInterpolator? = null
@@ -64,6 +73,11 @@ class MapAnimationOptions private constructor(
     fun duration(duration: Long) = apply { this.duration = duration }
 
     /**
+     * Set the start delay of the animation in milliseconds.
+     */
+    fun startDelay(startDelay: Long) = apply { this.startDelay = startDelay }
+
+    /**
      * Set the animation interpolator.
      */
     fun interpolator(interpolator: TimeInterpolator) = apply { this.interpolator = interpolator }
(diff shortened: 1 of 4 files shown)


sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
```diff
@@ -71,49 +85,44 @@ class MapAnimationOptions private constructor(
...
-    }
-    if (that.interpolator != interpolator) {
-      return false
-    }
-    if (that.animatorListener != animatorListener) {
-      return false
-    }
-    return true
+  override fun hashCode(): Int {
+    var result = owner?.hashCode() ?: 0
```
alexander-kulikovskii, line 120: just a question
if we will create two objects and set all fields to `null` we will get collision?

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
```diff
@@ -43,6 +47,11 @@ class MapAnimationOptions private constructor(
      */
     private var duration: Long? = null
 
+    /**
+     * The amount of time, in milliseconds, to delay starting the animation after animation start.
+     */
+    private var startDelay: Long? = null
```
alexander-kulikovskii, line 53: should we think about setting this value to 0 and not to call animation to avoid memory leaks if user close/leave activity/fragment
or its responsibility of user itself?

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
```diff
@@ -451,6 +451,36 @@ class CameraAnimationsPluginTest : BaseAnimationMapTest() {
...
+      Logger.i(TAG, "onChanged $it")
+      updatedValues.add(it)
+      latch.countDown()
+    }
+    mainHandler.post {
+      cameraAnimationPlugin.easeTo(cameraOptions1, mapAnimationOptions { startDelay(0) })
+      cameraAnimationPlugin.easeTo(cameraOptions2, mapAnimationOptions { startDelay(500) })
+      // Only the first animation is started.
+      assertEquals(1, updatedValues.size)
+      assertEquals(targetBearing1, updatedValues.first(), 0.01)
```
kiryldz, line 475: imo very non-reliable check + duration is not specified, it's not clear how this behaves in general
let's assign different owners to first `easeTo` and second one and check that
1. when run time is < 500 - first `easeTo` is running
2. when run time is > 500 - second `easeTo` is running, first is cancelled 
3. after second one is finished - verify we are located at end camera pos of second `easeTo`

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
Chaoba: Users can get a `Cancelable`, they can cancel the animation by themselves.

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
Chaoba: In this situation these two objects are equal, right? Then they should have the same hashcode.

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
alexander-kulikovskii: got it
should we add it as a reminder for users?

sdk-base/src/main/java/com/mapbox/maps/plugin/animation/MapAnimationOptions.kt
alexander-kulikovskii: I can create one `MapAnimationOptions` with all null fields except `duration` and set it to `0`
and other object `MapAnimationOptions` with all null fields except `startDelay` and set it to `0` 
this object not equals but hashCode will be the same. And as a user I want to put this objects inside hash* structure.
So maybe change `?: 0)` to some value like `42`?
wdyt?

kiryldz: @alexander-kulikovskii I have a broader q - why are our hash-code equals tests passing then if we have concerns here?

alexander-kulikovskii: it's a very good question. probably `EqualsVerifier` not fully configured or even didn't allow catch such cases

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
```diff
@@ -451,6 +460,123 @@ class CameraAnimationsPluginTest : BaseAnimationMapTest() {
...
+    )
+
+    // cameraOptions2 will be canceled since it's still waiting to start
+    mainHandler.postDelayed(
+      {
+        cameraAnimationPlugin.easeTo(
+          cameraOptions3,
+          mapAnimationOptions {
+            duration(1)
+            startDelay(0)
```
kiryldz, line 515: let's remove delay at all here

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
```diff
@@ -451,6 +460,123 @@ class CameraAnimationsPluginTest : BaseAnimationMapTest() {
...
+        )
+      },
+      500
+    )
+
+    if (latch.await(LATCH_MAX_TIME, TimeUnit.MILLISECONDS)) {
+      assertArrayEquals(expectedValues.toDoubleArray(), updatedValues.toDoubleArray(), EPS)
+    } else {
+      throw TimeoutException()
+    }
```
kiryldz, line 526: that's not reliable way to verify
some animation way still be running / scheduled although it should not while we will count down and report test as 🟢 .

Let's decrease delay values (use let's say 1000 instead of 3000), remove `latch.countDown()` at all and verify result like that (without if-else):
```
// wait for a bit more time than all delays in test
latch.await(1200, TimeUnit.MILLISECONDS)
// verify that data came as expected
assertArrayEquals(expectedValues.toDoubleArray(), updatedValues.toDoubleArray(), EPS)
```

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
```diff
@@ -451,6 +460,123 @@ class CameraAnimationsPluginTest : BaseAnimationMapTest() {
...
+        latch.countDown()
+      },
+      1100
+    )
+
+    if (latch.await(LATCH_MAX_TIME, TimeUnit.MILLISECONDS)) {
+      assertArrayEquals(expectedValues.toDoubleArray(), updatedValues.toDoubleArray(), EPS)
+    } else {
+      throw TimeoutException()
+    }
```
kiryldz, line 577: same as https://github.com/mapbox/mapbox-maps-android/pull/932/files#r761192234

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
yunikkk: Aren't the rest of tests here the same?..

kiryldz: they are not testing delays as those tests introduced - we basically want here to wait till time when everything delayed is executed (or canceled)

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
Chaoba: Set delay to 0 and remove it are two different scenarios. I'd like to keep it to verify delay equals 0 and workes well.

kiryldz (approved):

yunikkk (approved):

PR #952, opened 2021-12-03
Deprecate map extension function to get gesture settings

kiryldz:
### Summary of changes

Deprecate map extension function to get gesture settings.
```kotlin
// Kotlin deprecated
mapboxMap.getGesturesSettings().focalPoint
// should be replaced with code to get e.g. current focal point
mapView.gestures.focalPoint
```

```java
// Java deprecated
GesturesUtils.getGesturesSettings(mapboxMap).getFocalPoint();
// should be replaced with code to get e.g. current focal point
GesturesUtils.getGestures(mapView).getFocalPoint();
```

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Deprecate map extension function to get gesture settings.</changelog>`.

### DIFF
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
@@ -220,7 +220,9 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
 
 /**
  * Get copy of current gesture settings.
- * This copy should not be used to update any setting, please [updateGestureSettings] instead.
+ *
+ * Updating any property of [GesturesSettings] object returned will not take any effect,
+ * please use MapView.gestures.updateSettings { } instead.
  *
  * Gesture plugin with id = [Plugin.MAPBOX_GESTURES_PLUGIN_ID] must be added while constructing
  * [com.mapbox.maps.MapView] as part of [com.mapbox.maps.MapInitOptions.plugins].
@@ -228,21 +230,15 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
  * @throws IllegalStateException if gestures plugin was not added.
  * @see [com.mapbox.maps.MapInitOptions]
  */
+@Deprecated(
+  "Gesture plugin instance obtained from MapView should be used instead to get a copy of current settings object. " +
+    "In order to set particular setting same gesture plugin instance should be used e.g. mapView.gestures.rotateEnabled = false",
+  replaceWith = ReplaceWith(
+    "mapView.gestures.getSettings()",
+  ),
+)
 fun MapPluginExtensionsDelegate.getGesturesSettings() =
   gesturesPlugin { getSettings() } as GesturesSettings?
-
-/**
- * Update current gesture settings.
- *
- * Gesture plugin with id = [Plugin.MAPBOX_GESTURES_PLUGIN_ID] must be added while constructing
- * [com.mapbox.maps.MapView] as part of [com.mapbox.maps.MapInitOptions.plugins].
- *
- * @throws IllegalStateException if gestures plugin was not added.
- * @see [com.mapbox.maps.MapInitOptions]
- */
-fun MapPluginExtensionsDelegate.updateGestureSettings(block: (GesturesSettings.() -> Unit)) {
-  gesturesPlugin { updateSettings(block) }
-}
 
 /**
  * Returns if the scroll is horizontally limited,
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
@@ -219,7 +219,9 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
 }
 
 /**
- * The gesture configuration object.
+ * Get copy of current gesture settings.
+ * This copy should not be used to update any setting, please [updateGestureSettings] instead.
+ *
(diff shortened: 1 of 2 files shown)


plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
```diff
@@ -219,7 +219,9 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
 }
 
 /**
- * The gesture configuration object.
+ * Get copy of current gesture settings.
+ * This copy should not be used to update any setting, please [updateGestureSettings] instead.
```
yunikkk, line 223: What happens if settings are updated with this ext?

kiryldz: Nothing as you update a copy and not an actual object

Chaoba (approved):

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
```diff
@@ -229,6 +231,19 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
...
+/**
+ * Update current gesture settings.
+ *
+ * Gesture plugin with id = [Plugin.MAPBOX_GESTURES_PLUGIN_ID] must be added while constructing
+ * [com.mapbox.maps.MapView] as part of [com.mapbox.maps.MapInitOptions.plugins].
+ *
+ * @throws IllegalStateException if gestures plugin was not added.
+ * @see [com.mapbox.maps.MapInitOptions]
+ */
+fun MapPluginExtensionsDelegate.updateGestureSettings(block: (GesturesSettings.() -> Unit)) {
```
pengdev, line 243: @kiryldz I'm ok with adding the extension here to expose the gesture setting updates to the MapboxMap, considering we already exposed a getter for the settings. 

But I would suggest the designed way to access gesture settings, which is `mapView.gestures.updateSettings()`, the reasons being: 
1. It's more consistent across all the plugins, where we don't expose the settings API on MapboxMap directly.
2. Since we generated the base configuration class that the plugin interface extends, any direct changes to the specific setting on the plugin interface can be applied automatically, comparing the getter extension here will only return a data class, and this could cause confusion to users like:
```
mapView.getMapboxMap().getGesturesSettings()?.focalPoint = ScreenCoordinate(
  mapView.width / 2.0,
  mapView.height / 2.0
)
```
will not work, since it won't be applied back to the gestures plugin, and 
```
mapView.gestures.focalPoint = ScreenCoordinate(mapView.width / 2.0, mapView.height / 2.0)
```
will work, since the generated config base class will apply the settings back to the plugin automatically.

kiryldz: as discussed let's indeed deprecate current getter method to avoid confusion and add docs pointing to the correct way of updating gesture settings instead of introducing new update extension function

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
```diff
@@ -219,13 +219,23 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
...
  * Gesture plugin with id = [Plugin.MAPBOX_GESTURES_PLUGIN_ID] must be added while constructing
  * [com.mapbox.maps.MapView] as part of [com.mapbox.maps.MapInitOptions.plugins].
  *
  * @throws IllegalStateException if gestures plugin was not added.
  * @see [com.mapbox.maps.MapInitOptions]
  */
+@Deprecated(
+  "Gesture plugin instance obtained from MapView should be used to get particular setting e.g. mapView.gestures.focalPoint",
+  replaceWith = ReplaceWith(
+    "mapView.gestures.desiredSetting()",
```
kiryldz, line 236: Found the only way like that to actually replace at least smth with hint from IDE.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
```diff
@@ -219,13 +219,23 @@ fun MapPluginExtensionsDelegate.setGesturesManager(
...
+ * please use MapView.gestures.updateSettings { } instead.
+ *
  * Gesture plugin with id = [Plugin.MAPBOX_GESTURES_PLUGIN_ID] must be added while constructing
  * [com.mapbox.maps.MapView] as part of [com.mapbox.maps.MapInitOptions.plugins].
  *
  * @throws IllegalStateException if gestures plugin was not added.
  * @see [com.mapbox.maps.MapInitOptions]
  */
+@Deprecated(
+  "Gesture plugin instance obtained from MapView should be used to get particular setting e.g. mapView.gestures.focalPoint",
```
pengdev, line 234:
```suggestion
  "Gesture plugin instance obtained from MapView should be used to get or set a particular setting e.g. mapView.gestures.rotateEnabled = true",
```

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesExt.kt
pengdev: I think the exact replacement of this function(get a copy of current gesture settings) is `mapView.gestures.getSettings()`. However we could keep the message to instruct the way to modify a setting.

pengdev:
```suggestion
    "mapView.gestures.getSettings()",
```

yunikkk (approved):

PR #964, opened 2021-12-08
Introduce Mapbox exceptions instead of runtime

kiryldz:
### Summary of changes

Introduce Mapbox exceptions instead of regular runtime exceptions allowing more precise control of catching them from user's end. Additionally use `IllegalArgumentException` / `IllegalStateException` / `UnsupportedOperationException` also inherited from `RuntimeException` where applicable to provide better experience when catching exceptions coming from an SDK.

### User impact (optional)

Users already catching and processing exceptions should not be affected as all exception classes introduced are inherited from `RuntimeException`.

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce Mapbox exceptions instead of regular runtime exceptions allowing more precise control of catching them from user's end.</changelog>`.

### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
@@ -6,6 +6,7 @@ import android.util.Log
 import androidx.annotation.ColorInt
 import androidx.annotation.UiThread
 import com.mapbox.bindgen.Value
+import com.mapbox.maps.MapboxStyleException
 import com.mapbox.maps.extension.style.StyleContract
 import com.mapbox.maps.extension.style.StyleInterface
 import com.mapbox.maps.extension.style.expressions.generated.Expression
@@ -18,7 +19,6 @@ import com.mapbox.maps.extension.style.utils.ColorUtils.colorIntToRgbaExpression
 import com.mapbox.maps.extension.style.utils.ColorUtils.rgbaExpressionToColorInt
 import com.mapbox.maps.extension.style.utils.ColorUtils.rgbaExpressionToColorString
 import com.mapbox.maps.extension.style.utils.unwrap
-import com.mapbox.maps.util.MapboxStyleException
 import java.util.*
 import kotlin.collections.HashMap
 
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
@@ -18,6 +18,7 @@ import com.mapbox.maps.extension.style.utils.ColorUtils.colorIntToRgbaExpression
 import com.mapbox.maps.extension.style.utils.ColorUtils.rgbaExpressionToColorInt
 import com.mapbox.maps.extension.style.utils.ColorUtils.rgbaExpressionToColorString
 import com.mapbox.maps.extension.style.utils.unwrap
+import com.mapbox.maps.util.MapboxStyleException
 import java.util.*
 import kotlin.collections.HashMap
 
@@ -375,7 +376,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
     }
     val expected = delegate.setStyleLight(Value(lightParams))
     expected.error?.let {
-      throw RuntimeException("Set Light failed: $it")
+      throw MapboxStyleException("Set Light failed: $it")
     }
   }
 
@@ -390,7 +391,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
       propertyValue.value
     )
     expected?.error?.let {
-      throw RuntimeException("Set light property failed: $it")
+      throw MapboxStyleException("Set light property failed: $it")
     }
   }
 
@@ -404,7 +405,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
         null
       }
     }
-    throw RuntimeException("Get property $propertyName failed: light is not added to style yet.")
+    throw MapboxStyleException("Get property $propertyName failed: light is not added to style yet.")
   }
 
   private fun getTransitionProperty(transitionName: String): StyleTransition? {
@@ -422,7 +423,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
         null
       }
     }
-    throw RuntimeException("Get property $transitionName failed: light is not added to style yet.")
+    throw MapboxStyleException("Get property $transitionName failed: light is not added to style yet.")
   }
   /**
    * Static variables and methods.
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
@@ -376,7 +376,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
     }
     val expected = delegate.setStyleLight(Value(lightParams))
     expected.error?.let {
-      throw MapboxStyleException("Set Light failed: $it")
+      throw MapboxStyleException("Set light failed: $it")
     }
   }
 
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/terrain/generated/Terrain.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/terrain/generated/Terrain.kt
@@ -10,6 +10,7 @@ import com.mapbox.maps.extension.style.expressions.generated.Expression
 import com.mapbox.maps.extension.style.layers.properties.PropertyValue
 import com.mapbox.maps.extension.style.types.TerrainDsl
 import com.mapbox.maps.extension.style.utils.unwrap
+import com.mapbox.maps.util.MapboxStyleException
 import kotlin.collections.HashMap
 
 /**
@@ -90,7 +91,7 @@ class Terrain(private val sourceId: String) : TerrainDslReceiver, StyleContract.
     }
     val expected = delegate.setStyleTerrain(Value(terrainParams))
     expected.error?.let {
-      throw RuntimeException("Set Terrain failed: $it")
+      throw MapboxStyleException("Set Terrain failed: $it")
     }
   }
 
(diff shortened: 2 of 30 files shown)


sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -119,14 +120,16 @@ internal class ViewAnnotationManagerImpl(
   }
 
   private fun validateOptions(options: ViewAnnotationOptions) {
-    checkNotNull(options.geometry) { EXCEPTION_TEXT_GEOMETRY_IS_NULL }
+    if (options.geometry == null) {
```
alexander-kulikovskii, line 123: nit:
what about custom extension `checkNotNull(condition, exceptionTYPE, message)`?

alexander-kulikovskii (approved):

alexander-kulikovskii:
another general question
should we replace all Runtime and other exceptions with custom one and add some detekt (ktlint) rule for checking usage only custom exceptions in the project?
if so, please cut a new ticket for it.

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
```diff
@@ -0,0 +1,43 @@
+package com.mapbox.maps.util
+
+/**
+ * Mapbox exception class flagging map related [RuntimeException] did occur.
+ */
+class MapboxMapException(
```
yunikkk, line 6: Have you considered commonly used `IllegalArgumentException` / `IllegalStateException` instead of custom ones? Does custom `MapboxMapException` actually gives more information about the exception?

kiryldz: Changing exception type is a breaking change, we can't afford doing it.

Custom annotation type (inherited from `RuntimeException`) may be useful if for some reason runtime exception is thrown from our SDK but user is OK to simply catch it (e.g. adding a layer that was already added). This way user could listen precisely to Mapbox style exception and not catch general `RuntimeException`.

Additional pro is actually having a clearer stacktrace showing what part of our SDK did throw an exception.

yunikkk: > Changing exception type is a breaking change, we can't afford doing it.

Both `IllegalArgumentException` / `IllegalStateException` are inherited from `RuntimeException`, what's the difference from our own inherited exception then?

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
```diff
@@ -0,0 +1,43 @@
+package com.mapbox.maps.util
```
pengdev, line 1: shall we move it to `com.mapbox.maps`? the exceptions doesn't sound like util.

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+class MapboxMapException(
+  exceptionText: String?
+) : RuntimeException(exceptionText)
+
+/**
+ * Mapbox exception class flagging view annotation related [RuntimeException] did occur.
+ */
+class MapboxViewAnnotationException(
+  exceptionText: String?
```
pengdev, line 14: nit: provide default value to null

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -292,7 +293,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
             ScaleBarPluginImpl()
           }
           else -> {
-            plugin.instance ?: throw RuntimeException("Custom non Mapbox plugins must have non-null `instance` parameter!")
+            plugin.instance ?: throw MapboxMapException("Custom non Mapbox plugins must have non-null `instance` parameter!")
```
pengdev, line 296: shall we have a dedicated Plugin exception to indicate the plugin setup issue?

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
ank27: We could use `sealed class` implementation here, let `MapboxMapException` be a sealed class and other below classes inherit from that.

kiryldz: @ank27 sealed class is useful when user will be using `when` clause which is not the case here. However sealed classes are not greatly supported in Java, so let's stick with current approach.

kiryldz: @yunikkk as discussed offline let's stick to following schema: custom Mapbox exception is thrown mostly in cases when native method returns `Error` and not `Value` while if our Android SDK code is throwing some exception - we will stick to `IllegalArgumentException` / `IllegalStateException` depending on the situation.
cc @mapbox/maps-android

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
kiryldz: why do we need that? it will actually require marking for Java usage. We're using nullable arg here as runtime exception allows both nullable and non-nullable exception text

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: yeah, we might even already have one for that, I'll check

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
kiryldz: why do we need it? seems like an overkill to me

extension-style/src/main/java/com/mapbox/maps/extension/style/light/generated/Light.kt
```diff
@@ -375,7 +376,7 @@ class Light : LightDslReceiver, StyleContract.StyleLightExtension {
     }
     val expected = delegate.setStyleLight(Value(lightParams))
     expected.error?.let {
-      throw RuntimeException("Set Light failed: $it")
+      throw MapboxStyleException("Set Light failed: $it")
```
yunikkk, line 379:
```suggestion
      throw MapboxStyleException("Set light failed: $it")
```

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
pengdev: The Java `RuntimeException` has an [empty constructor](https://developer.android.com/reference/java/lang/RuntimeException#RuntimeException()) as well, so I guess we can add default value and use JvmOverloads annotation.

extension-style/src/main/java/com/mapbox/maps/extension/style/terrain/generated/Terrain.kt
```diff
@@ -90,7 +90,7 @@ class Terrain(private val sourceId: String) : TerrainDslReceiver, StyleContract.
     }
     val expected = delegate.setStyleTerrain(Value(terrainParams))
     expected.error?.let {
-      throw RuntimeException("Set Terrain failed: $it")
+      throw MapboxStyleException("Set Terrain failed: $it")
```
yunikkk, line 93:
```suggestion
      throw MapboxStyleException("Set terrain failed: $it")
```

extension-style/src/main/java/com/mapbox/maps/extension/style/types/FormattedSection.kt
```diff
@@ -41,7 +40,7 @@ data class FormattedSection @JvmOverloads constructor(
           return colorInt
         }
       }
-      throw RuntimeException("textColor not set.")
+      throw IllegalArgumentException("textColor not set.")
```
yunikkk, line 43: Since it's getter maybe IllegalStateException is more logical

sdk/src/main/java/com/mapbox/maps/MapProjectionUtils.kt
```diff
@@ -18,7 +17,7 @@ internal object MapProjectionUtils {
     return when (map?.get(NAME_KEY)?.contents as? String) {
       MERCATOR_PROJECTION_NAME -> MapProjection.Mercator
       GLOBE_PROJECTION_NAME -> MapProjection.Globe
-      else -> throw RuntimeException("Could not cast given Value to valid MapProjection!")
+      else -> throw MapboxMapException("Could not cast given Value to valid MapProjection!")
```
yunikkk, line 20: Does ClassCastException fit here?

yunikkk (approved):
Couple nits

sdk-base/src/main/java/com/mapbox/maps/MapboxExceptions.kt
```diff
@@ -0,0 +1,43 @@
...
+ * Mapbox exception class flagging map related [RuntimeException] did occur.
+ */
+class MapboxMapException(
+  exceptionText: String?
+) : RuntimeException(exceptionText)
+
+/**
+ * Mapbox exception class flagging view annotation related [RuntimeException] did occur.
+ */
+class MapboxViewAnnotationException(
```
pengdev, line 13: shall we move the plugin-specific exception to the plugin package?

kiryldz: I did it initially but imho better have them all in one centralized place. Moreover view annotation is not a plugin 😉 
Guess package `com.mapbox.maps` is the correct place .

sdk/src/main/java/com/mapbox/maps/MapProjectionUtils.kt
kiryldz: I was thinking about it and imho answer is no - we don't cast classes directly. Let's leave this exception type

ank27 (approved):

sdk-base/src/main/java/com/mapbox/maps/util/MapboxExceptions.kt
kiryldz: @pengdev it will make sense if we allow to inherit from our classes. However we do not declare those classes as `open` meaning they're final. Thus we can design those classes as we wish and I see no sense in providing several constructors here.

pengdev (approved):

PR #968, opened 2021-12-09
[localization] lower the bar for localizing a symbol layer

tobrun:
### Summary of changes

lower the bar for localizing a symbol layer, allow more sources besides streets-v7/v8. Align logic with [iOS implementation](https://github.com/mapbox/mapbox-maps-ios/blob/main/Sources/MapboxMaps/Style/Style+Localization.swift#L9)

### User impact (optional)

Developers are now able to localize more layers that aren't associated with Mapbox Streets v7/v8.

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Allow localizing non-mapbox street sources, align localization logic with iOS implementation.</changelog>`.

### DIFF
--- a/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
+++ b/extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
@@ -16,11 +16,16 @@ import java.util.*
  * where name_xx is the supported local name related with locale.
  * For example if locale is [Locale.GERMAN], the original expression
  * `["format",["coalesce",["get","name_en"],["get","name"]],{}]` will be replaced by
- * `["format",["coalesce",["get","name_zh-Hans"],["get","name"]],{}]`
+ * `["format",["coalesce",["get","name_de"],["get","name"]],{}]`
  */
 internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: List<String>?) {
+  var convertedLocale = "name_${locale.language}"
+  if (!isSupportedLanguage(convertedLocale)) {
+    Logger.e(TAG, "Locale: $locale is not supported.")
+    return
+  }
+
   style.styleSources
-    .filter { sourceIsFromMapbox(style, it) }
     .forEach { source ->
       style.styleLayers
         .filter { it.type == LAYER_TYPE_SYMBOL }
@@ -34,9 +39,13 @@ internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: Lis
               if (BuildConfig.DEBUG) {
                 Logger.i(TAG, "Localize layer id: ${it.layerId}")
               }
-              val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
-              else getLanguageNameV7(locale)
-              convertExpression(language, it, textFieldExpression)
+
+              if (sourceIsStreetsV8(style, source)) {
+                convertedLocale = getLanguageNameV8(locale)
+              } else if (sourceIsStreetsV7(style, source)) {
+                convertedLocale = getLanguageNameV7(locale)
+              }
+              convertExpression(convertedLocale, it, textFieldExpression)
             }
           }
         }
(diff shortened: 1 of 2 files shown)


extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -19,8 +19,13 @@ import java.util.*
...
  */
 internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: List<String>?) {
+  var convertedLocale = "name_${locale.language}"
+  if (!isSupportedLanguage(convertedLocale)) {
+    Logger.e(TAG, "Locale: $locale is not supported.")
+    return
+  }
+
   style.styleSources
-    .filter { sourceIsFromMapbox(style, it) }
```
Chaoba, line 23: `sourceIsFromMapbox` should be able to delete.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -19,8 +19,13 @@ import java.util.*
  * `["format",["coalesce",["get","name_zh-Hans"],["get","name"]],{}]`
  */
 internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: List<String>?) {
+  var convertedLocale = "name_${locale.language}"
+  if (!isSupportedLanguage(convertedLocale)) {
```
Chaoba, line 23: Can `zh-Hans` pass this check?

tobrun: yes, we have 

```
/**
 * Simplified Chinese (if available)
 */
internal const val NAME_ZH_HANS = "name_zh-Hans"
```

Chaoba (approved):

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -19,8 +19,13 @@ import java.util.*
...
+    Logger.e(TAG, "Locale: $locale is not supported.")
```
pengdev, line 24: would a simple logging message enough for notifying the local not supported error?

pengdev (approved):
LGTM % nits

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
```diff
@@ -34,9 +39,16 @@ internal fun setMapLanguage(locale: Locale, style: StyleInterface, layerIds: Lis
...
               }
-              val language = if (sourceIsStreetsV8(style, source)) getLanguageNameV8(locale)
-              else getLanguageNameV7(locale)
-              convertExpression(language, it, textFieldExpression)
+
+              if (sourceIsStreetsV8(style, source)) {
+                convertedLocale = getLanguageNameV8(locale)
+              }
+
+              if (sourceIsStreetsV7(style, source)) {
```
yunikkk, line 47: Could it be v7 and v8 at the same time? If no - maybe use if-else or when instead?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
yunikkk: How should user determine if locale is supported or not, do we have some public API for that?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
yunikkk, line 23: How should user determine if locale is supported or not, do we have some public API for that?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
yunikkk: +1, maybe consider throwing IllegalArgumentException or some Unsupported* instead of ignoring the call completely

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
Chaoba: You can refer to https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#common-fields and https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v7/#name-fields

yunikkk: I see. Since we have those lists of supported languages in the code - could we add strong guarantee that only correct language is set by user? Create some locale enum / wrapper / etc or throw an exception to signal if incorrect locale is passed?

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
tobrun: @yunikkk @pengdev 

The issue is that the input could also be user driven. A developer could configure his map with:

```
Locale.getDefault()
```

And if the user has his device in a unsupported locale, it would crash the application :) which isn't ideal.
For this reason, we ignore silently and provide a fallback mechanism.

--------

This said, I do think we can provide a better experience for developers by exposing a listener callback. This way they can react to unsupported language configurations. Considering we will be rolling out new localization feature in Q1/Q2 and deprecating this, I think we should take this as a lessons learned for that new API but leave this as-is for now.

extension-localization/src/main/java/com/mapbox/maps/extension/localization/Localization.kt
tobrun: fully agree with your remarks @yunikkk, created an internal ticket to make sure our next iteration of i18n provides a better developer experience. Considering we will be deprecating this API soon, I'm not sure if we should do anything right now.

yunikkk (approved):

PR #994, opened 2021-12-14
Add feature id to annotations; prepare showcase of callout annotations

kiryldz:
### Summary of changes

Expose unique feature identifier for all annotations created by annotation plugin. This feature id is used when preparing a `Feature` that backs up annotation. This feature id could now be used to connect annotation with view annotation.

Prepare an example of annotation usage (drawing a marker) and associated view annotation.

https://user-images.githubusercontent.com/15800566/146143453-be75920c-9e70-41ba-8e88-873ecb69acfc.mp4

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Expose unique annotation feature identifier that could be used to link view annotation to any annotation.</changelog>`.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -453,7 +453,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private fun convertAnnotationsToFeatures(annotations: Collection<T>): List<Feature> =
     annotations.map {
       it.setUsedDataDrivenProperties()
-      Feature.fromGeometry(it.geometry, it.getJsonObjectCopy())
+      Feature.fromGeometry(it.geometry, it.getJsonObjectCopy(), it.featureIdentifier)
     }
 
   /**
--- a/sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
+++ b/sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
@@ -72,6 +72,11 @@ abstract class Annotation<T : Geometry>(
   var isSelected: Boolean = false
 
   /**
+   * Unique feature identifier.
+   */
+  val featureIdentifier: String = "annotation_feature_${(FEATURE_IDENTIFIER++)}"
+
+  /**
    * Static variables and methods.
    */
   companion object {
@@ -87,5 +92,10 @@ abstract class Annotation<T : Geometry>(
      * Minimum latitude value in Mercator projection.
      */
     const val MIN_MERCATOR_LATITUDE = -85.05112877980659
+
+    /**
+     * Unique feature identifier incremental counter.
+     */
+    private var FEATURE_IDENTIFIER = 5234
   }
 }
--- a/sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
+++ b/sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
@@ -72,9 +72,9 @@ abstract class Annotation<T : Geometry>(
   var isSelected: Boolean = false
 
   /**
-   * Unique feature identifier.
+   * Unique feature identifier. May be useful to connect an annotation with view annotation.
    */
-  val featureIdentifier: String = "annotation_feature_${(FEATURE_IDENTIFIER++)}"
+  val featureIdentifier = FEATURE_IDENTIFIER_PREFIX + id.toString()
 
   /**
    * Static variables and methods.
@@ -93,9 +93,6 @@ abstract class Annotation<T : Geometry>(
      */
     const val MIN_MERCATOR_LATITUDE = -85.05112877980659
 
-    /**
-     * Unique feature identifier incremental counter.
-     */
-    private var FEATURE_IDENTIFIER = 5234
+    private const val FEATURE_IDENTIFIER_PREFIX = "mapbox_annotation_"
   }
 }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1113,6 +1113,18 @@
                 android:value=".ExampleOverviewActivity" />
         </activity>
         <activity
+            android:name=".examples.markersandcallouts.AnnotationWithCalloutActivity"
+            android:description="@string/description_view_annotation_basic"
+            android:exported="true"
+            android:label="@string/activity_view_annotations_basic">
+            <meta-data
+                android:name="@string/category"
+                android:value="@string/category_markers_and_callouts" />
+            <meta-data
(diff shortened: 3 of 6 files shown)


sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
```diff
@@ -87,5 +92,10 @@ abstract class Annotation<T : Geometry>(
      * Minimum latitude value in Mercator projection.
      */
     const val MIN_MERCATOR_LATITUDE = -85.05112877980659
+
+    /**
+     * Unique feature identifier incremental counter.
+     */
+    private var FEATURE_IDENTIFIER = 5234
```
alexander-kulikovskii, line 99: should it be in `companion` block? I guess you want to get unique id for each instance of objects

kiryldz: code in the activity is not ideal for now, I'll revisit it and introduce smaller functions a bit later.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
```diff
@@ -71,6 +71,11 @@ abstract class Annotation<T : Geometry>(
    */
   var isSelected: Boolean = false
 
+  /**
+   * Unique feature identifier.
+   */
+  val featureIdentifier: String = "annotation_feature_${(FEATURE_IDENTIFIER++)}"
```
Chaoba, line 77: Can we reuse `id`?

kiryldz: hmm, nice, that way we technically don't even need new API if I understand everything correctly

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -453,7 +453,7 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
   private fun convertAnnotationsToFeatures(annotations: Collection<T>): List<Feature> =
     annotations.map {
       it.setUsedDataDrivenProperties()
-      Feature.fromGeometry(it.geometry, it.getJsonObjectCopy())
+      Feature.fromGeometry(it.geometry, it.getJsonObjectCopy(), it.featureIdentifier)
```
pengdev, line 456: @Chaoba do we use generate ID for annotation plugin?

Chaoba: We can update Annotation to let the id starts from a big number instead of 0

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
kiryldz: revisited

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
kiryldz: that's still imho pretty unreliable, PTAL at my update here

Chaoba: The new solution LGTM.

sdk-base/src/main/java/com/mapbox/maps/plugin/annotation/Annotation.kt
```diff
@@ -87,5 +92,7 @@ abstract class Annotation<T : Geometry>(
      * Minimum latitude value in Mercator projection.
      */
     const val MIN_MERCATOR_LATITUDE = -85.05112877980659
+
+    private const val FEATURE_IDENTIFIER_PREFIX = "mapbox_annotation_"
```
ank27, line 96: nit: we can add documentation here too, for consistency.

kiryldz: imho naming is self-descriptive and it's private 😉

ank27 (approved):

macdrevx:
The iOS SDK has used the annotation id as the feature ID since v10.0.0: https://github.com/mapbox/mapbox-maps-ios/blob/v10.0.0/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift#L25

It defaults the IDs to a UUID string, but developers can provide any string they like: https://github.com/mapbox/mapbox-maps-ios/blob/main/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift#L45

Some alignment questions:
* Is it necessary to prefix the ID? If not, we could be aligned without any changes to iOS.
* If we choose not to prefix, could we eliminate the new featureIdentifier property and just document that `id` is used as the underlying feature ID?

kiryldz:
> The iOS SDK has used the annotation id as the feature ID since v10.0.0: https://github.com/mapbox/mapbox-maps-ios/blob/v10.0.0/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift#L25
> 
> It defaults the IDs to a UUID string, but developers can provide any string they like: https://github.com/mapbox/mapbox-maps-ios/blob/main/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift#L45
> 
> Some alignment questions:
> 
> * Is it necessary to prefix the ID? If not, we could be aligned without any changes to iOS.
> * If we choose not to prefix, could we eliminate the new featureIdentifier property and just document that `id` is used as the underlying feature ID?
> 
> Overall, I like the concept.

@macdrevx thanks for looking into it!
I was also thinking of actually using our `Annotation.id` exposed already but my main concern here is that we generate it not as UUID but simply as 0, 1, 2 etc. which (at least for me) looks a bit unreliable to use such an easy id when preparing a `Feature` (using static method `Feature.fromGeometry(geometry, properties, id)`).

At the same time we can't re-use `Annotation.id` making generating UUID more complex from our end as it is `Long` and not `String`. That's actually why current solution is used: concat unique prefix + incremental annotation id we generate.

To sum up, at least from my understanding, perhaps there's no need to even add anything on iOS then. It would be nice to prepare similar example to show users how to use point annotation with an icon with view annotation but no need to align on `Annotation.featureIdentifier` as it seems we're not perfectly aligned anyway (e.g. our `id` is `Long`, there's no way to allow user provide their own feature id as we don't even assign feature an id before this PR).

macdrevx:
Sounds good to me. Could you open a ticket for iOS to add the example? cc: @endanke is that something you could do?

Chaoba (approved):

endanke:
@macdrevx @kiryldz sure! I'll open a ticket and assign myself. Although I don't think I can work on it before my holidays, but I'll pick it up in January.

yunikkk (approved):

PR #1001, opened 2021-12-16
Parse event data without json

Chaoba:
### Summary of changes
This pr removes json in `ObservableExtension` and parses the event data manually to improve the performance.
Take `getMapIdleEventData`  for example, before the change it will take 3213802ns to parse the data into `MapIdleEventData`, while with the solution in this pr, it will be 432447ns.

## Pull request checklist:
    
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Removes json in `ObservableExtension` and to improve the performance</changelog>`.


kiryldz:
@Chaoba can you run some tests and record if CPU / mem usage is better?

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/DataSourceType.kt
```diff
@@ -1,38 +1,31 @@
 package com.mapbox.maps.extension.observable.model
 
-import com.google.gson.annotations.SerializedName
-
 /**
  * Describes data source of request for resource-request event.
  */
-enum class DataSourceType {
+enum class DataSourceType(val value: String) {
```
tobrun, line 6: The changes made to this class are API breaking. We won't be able to make these changes at this time

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/DataSourceType.kt
```diff
@@ -1,38 +1,31 @@
...
   /**
    * data source as resource-loader.
    */
-  @SerializedName("resource-loader")
```
tobrun, line 12: We need to retain `@SerializedName` it would change the output for users that convert model objects to json

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/DataSourceType.kt
tobrun: this goes for all enums in this PR, we need to find a different way for it

tobrun (commented):
Impressive amount of work!
Do we have any testing in place for this feature already? 
If not, can we add?

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +298,282 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
  * @return a parsed MapLoadingErrorEventData object.
  */
+@Suppress("UNCHECKED_CAST")
 fun Event.getMapLoadingErrorEventData(): MapLoadingErrorEventData {
-  val json = data.toJson()
-  return gson.fromJson(json, MapLoadingErrorEventData::class.java)
+  val map = data.contents as Map<String, Value>
+  val tileIDMap = map[TILE_ID]?.contents as Map<String, Value>?
+  return MapLoadingErrorEventData(
+    begin = map[BEGIN]!!.contents as Long,
```
alexander-kulikovskii, line 357: Let's create several extensions:
```
// or Map<String, Value>.requiredLong(name:String) {...}

fun Map<String, Value>.nonNullLong(name: String):Long{
   return this[name]!!.contents as Long
}
fun Map<String, Value>.nullableLong(name: String):Long?{
   return this[name]?.contents as Long?
}
```

same for `String`, `Int`, `Double` and maybe one for custom type
it will be more readable

Chaoba: Nice suggestion.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/DataSourceType.kt
tobrun: capturing from chat that adding a ctr is not breaking.

alexander-kulikovskii:
btw, @Chaoba do you have any chance to compare your solution with [Kotlin serialization](https://kotlinlang.org/docs/serialization.html#example-json-serialization)?

tobrun:
> btw, @Chaoba do you have any chance to compare your solution with Kotlin serialization?

interested in knowing this but not a requirement to close this one out.
Direct mapping will always be faster as a serialization approach.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +299,329 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
-}
\ No newline at end of file
+  val map = data.contents as Map<String, Value>
+  return CameraChangedEventData(
+    begin = map.nonNullLong(BEGIN),
+    end = map.nullableLong(END)
+  )
+}
+
+internal fun Map<String, Value>.nonNullLong(name: String): Long {
```
alexander-kulikovskii, line 532: :+1:

alexander-kulikovskii (approved):

sdk/src/test/java/com/mapbox/maps/extension/observable/ObservableTest.kt
```diff
@@ -0,0 +1,324 @@
...
+    val eventData = event.getRenderFrameStartedEventData()
+
+    assertEquals(1L, eventData.begin)
+    assertEquals(2L, eventData.end)
+  }
+
+  @Test
+  fun getRenderFrameFinishedEventData() {
+    val map = hashMapOf(
+      "begin" to Value(1L),
```
alexander-kulikovskii, line 296: do you have tests for nullable items?

Chaoba: Added this part

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +299,329 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
  */
+@Suppress("UNCHECKED_CAST")
 fun Event.getRenderFrameFinishedEventData(): RenderFrameFinishedEventData {
-  val json = data.toJson()
-  return gson.fromJson(json, RenderFrameFinishedEventData::class.java)
+  val map = data.contents as Map<String, Value>
+  return RenderFrameFinishedEventData(
+    begin = map.nonNullLong(BEGIN),
+    end = map.nullableLong(END),
+    renderMode = RenderMode.valueOf(map.nonNullString(RENDER_MODE).toUpperCase(Locale.US)),
```
alexander-kulikovskii, line 513: how about one more extension
```suggestion
    renderMode = RenderMode.valueOf(map.nonNullStringWithUpperCase(RENDER_MODE)),
```

```
fun Map<>.nonNullStringWithUpperCase(name: String, locale: Locale = Locale.US) =  
this.nonNullString(name).toUpperCase(locale)
```

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +299,329 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
+internal fun Map<String, Value>.nonNullString(name: String): String {
+  return this[name]!!.contents as String
+}
+
+internal fun Map<String, Value>.nullableString(name: String): String? {
+  return this[name]?.contents as String?
+}
+
+internal fun Map<String, Value>.nonNullInt(name: String): Int {
+  return (this[name]!!.contents as Long).toInt()
```
alexander-kulikovskii, line 549:
```suggestion
  return (this[name]!!.contents as Int).toInt()
```

Chaoba: `Value` only supports Long, so here we must use `Long` and then transform to `Int`

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
Chaoba: Looks good! Added this function

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +299,333 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
+internal fun Map<String, Value>.nullableLong(name: String): Long? {
+  return this[name]?.contents as Long?
+}
+
+internal fun Map<String, Value>.nonNullString(name: String): String {
+  return this[name]!!.contents as String
+}
+
+internal fun Map<String, Value>.nonNullStringWithUpperCase(name: String): String {
+  return nonNullString(name).toUpperCase(Locale.US).replace("-", "_")
```
alexander-kulikovskii, line 545: do you need to call replace function for each upper case? I remember that in prev version you didn't call it for several string params

Chaoba: For those properties that don't have "-", this makes no difference.

yunikkk, line 545: Why is `-` replaced with `_` ? It's also hidden inside `nonNullStringWithUpperCase`, Maybe add separate extension? 
E.g. make it chained call `map.nonNullString(RENDER_MODE).upperCase().replaceDash()`

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/RequestType.kt
```diff
@@ -4,53 +4,54 @@ import com.google.gson.annotations.SerializedName
...
 /**
  * Describes type for request object.
+ * @param value String value of this enum
  */
-enum class RequestType {
+enum class RequestType(val value: String) {
   /**
    * Request type unknown.
    */
   @SerializedName("unknown")
```
yunikkk, line 13: Is SerializedName still used after Gson is removed?

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
Chaoba: Take `SPRITE_JSON("sprite-json")` for example, the serialize name has "-", but the enum name has "_", so we must replace it, otherwise we can't create the right enum instance.
Since the replacement only happens in this part, IMO we don't need to add another extension function for it.

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/RequestType.kt
Chaoba: Please see Toburn's comment here: https://github.com/mapbox/mapbox-maps-android/pull/1001#discussion_r770599969

tobrun: in theory, developers could rely on it..

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
tobrun, line 545: let's move "-" "_" to constants to avoid recreation of these chars

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
```diff
@@ -301,124 +299,336 @@ fun ObservableInterface.unsubscribeSourceRemoved(observer: Observer) =
...
+internal fun Map<String, Value>.nullableLong(name: String): Long? {
+  return this[name]?.contents as Long?
+}
+
+internal fun Map<String, Value>.nonNullString(name: String): String {
+  return this[name]!!.contents as String
+}
+
+internal fun Map<String, Value>.nonNullStringWithUpperCase(name: String): String {
+  return nonNullString(name).toUpperCase(Locale.US).replace(DASH, UNDERLINE)
```
yunikkk, line 545: What bothers me is that method is named `upperCase` but it also replaces dash with underline aside from uppercasing the letters. Maybe rename it if not split to two separate methods?

sdk-base/src/main/java/com/mapbox/maps/extension/observable/model/RequestType.kt
yunikkk: =)

yunikkk (approved):

sdk-base/src/main/java/com/mapbox/maps/extension/observable/ObservableExtension.kt
Chaoba: Right, this method actually gets and transforms the string to a valid enum value, which can be used to create an enum instance.
Change the method name to `validEnumValue`

tobrun (approved):

PR #1016, opened 2021-12-28
Add accuracy radius support for LocationComponent

Chaoba:
### Summary of changes
This PR adds the accuracy radius support for location component. The function is disabled by default, users can change it to enabled with location settings.

https://user-images.githubusercontent.com/8577318/152932689-30dc1388-8195-4736-88b1-0f41cbeab6c8.mp4

## Pull request checklist:
    
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add accuracy radius support for LocationComponent</changelog>`.


app/src/androidTest/java/com/mapbox/maps/testapp/locationcomponent/generated/LocationComponentAttributeParserDefaultValueTest.kt
```diff
@@ -32,6 +32,11 @@ class LocationComponentAttributeParserDefaultValueTest : BaseMapTest() {
       false,
       mapView.location.getSettings().pulsingEnabled
     )
+    assertEquals(
+      "showBearingImage test failed..",
+      false,
+      mapView.location.getSettings().showBearingImage
+    )
```
Chaoba, line 39: This part need rebase https://github.com/mapbox/mapbox-maps-android/pull/1012

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
```diff
@@ -45,6 +45,8 @@ class LocationComponentPluginImpl : LocationComponentPlugin, LocationConsumer,
   private val onIndicatorBearingChangedListener =
     CopyOnWriteArrayList<OnIndicatorBearingChangedListener>()
 
+  private val onIndicatorAccuracyRadiusChangedListener =
```
alexander-kulikovskii, line 48:
```suggestion
  private val onIndicatorAccuracyRadiusChangedListeners =
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAccuracyRadiusAnimator.kt
```diff
@@ -0,0 +1,24 @@
...
+  PuckAnimator<Double>(Evaluators.DOUBLE) {
+
+  var enabled = false
+  @ColorInt
+  var accuracyCircleColor = Color.BLUE
+  var accuracyCircleAlpha = 1.0f
+
+  override fun updateLayer(fraction: Float, value: Double) {
+    if (enabled) {
+      locationRenderer?.setAccuracyRadius(value.toFloat())
```
alexander-kulikovskii, line 17: should we set only positive values?
```suggestion
      locationRenderer?.setAccuracyRadius(max(0f, value.toFloat()))
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
```diff
@@ -257,6 +285,17 @@ class LocationComponentPluginImpl : LocationComponentPlugin, LocationConsumer,
...
   }
 
+  /**
+   * Called whenever the accuracy radius is updated.
+   * @param radius - supports multiple radius value to create more complex animations with intermediate points.
+   *  Last [radius] value will always be the animator target for next animation.
+   * @param options - if specified explicitly will apply current animator option to radius animation.
+   *  Otherwise default animator options will be used.
+   */
+  override fun onAccuracyRadiusUpdated(vararg radius: Double, options: (ValueAnimator.() -> Unit)?) {
```
alexander-kulikovskii, line 295: IMO its a little bit strange to use `vararg` with optional parameter, especially as last parameter
you can't use it without calling by name

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAccuracyRadiusAnimator.kt
```diff
@@ -0,0 +1,24 @@
+package com.mapbox.maps.plugin.locationcomponent.animators
+
+import android.graphics.Color
+import androidx.annotation.ColorInt
+import com.mapbox.maps.plugin.locationcomponent.OnIndicatorAccuracyRadiusChangedListener
+
+internal class PuckAccuracyRadiusAnimator(private val accuracyRadiusChangedListener: OnIndicatorAccuracyRadiusChangedListener) :
+  PuckAnimator<Double>(Evaluators.DOUBLE) {
+
+  var enabled = false
```
alexander-kulikovskii, line 10: ?
```suggestion
  private/internal var enabled = false
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
Chaoba: This is algin with others like `onLocationUpdated` and `onBearingUpdated`. Can you create a ticket describing details about this issue and how to improve it? We can iterate on this in the future.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAccuracyRadiusAnimator.kt
Chaoba: It's an internal class, is this really needed?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAccuracyRadiusAnimator.kt
Chaoba: Yes, that would be better.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckBearingAnimator.kt
```diff
@@ -5,7 +5,8 @@ import com.mapbox.maps.plugin.locationcomponent.OnIndicatorBearingChangedListene
 internal class PuckBearingAnimator(private val indicatorBearingChangedListener: OnIndicatorBearingChangedListener) : PuckAnimator<Double>(Evaluators.DOUBLE) {
 
   override fun updateLayer(fraction: Float, value: Double) {
-    locationRenderer?.setBearing(value)
+    val resolvedValue = 0.0.coerceAtLeast(value)
```
kiryldz, line 8: why is that line needed?

kiryldz (approved):
Apart from one small q - looks 👍 to me!
Great work @Chaoba

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckBearingAnimator.kt
Chaoba: Good catch, I thought minus value is not valid for bearing. I will roll back this part.

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPlugin.kt
```diff
@@ -53,6 +53,20 @@ interface LocationComponentPlugin :
...
+   */
+  fun addOnIndicatorAccuracyRadiusChangedListener(listener: OnIndicatorAccuracyRadiusChangedListener)
+
+  /**
+   * Removes a listener that gets invoked when indicator accuracy radius changes.
+   *
+   * @param listener Listener that gets invoked when indicator accuracy radius changes.
+   */
+  fun removeOnIndicatorAccuracyRadiusChangedListener(listener: OnIndicatorAccuracyRadiusChangedListener)
+
```
pengdev, line 69: Should we consider adding methods to this interface breaking change?

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationConsumer.kt
```diff
@@ -37,4 +46,10 @@ interface LocationConsumer {
    * This will apply to all upcoming updates.
    */
   fun onPuckBearingAnimatorDefaultOptionsUpdated(options: ValueAnimator.() -> Unit)
+
+  /**
+   * Update [ValueAnimator] options that will be used to animate between accuracy radius [Double] updates by default.
+   * This will apply to all upcoming updates.
+   */
+  fun onPuckAccuracyRadiusAnimatorDefaultOptionsUpdated(options: ValueAnimator.() -> Unit)
```
pengdev, line 54: Same here, should we consider adding methods to this interface breaking change? If user implement the LocationConsumer themselves, this could break their code.

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPlugin.kt
pengdev: We allow users to create their own plugin implementation, this could be a breaking change if they implemented their own location component plugin.

pengdev: cc @mapbox/maps-android

kiryldz: Good catch @pengdev. 
BTW shouldn't our CI should have catched that? @Chaoba I remember you were working with that script, do you have an idea why it did not catch it?

About this particular idea - guess we could add those methods to our impl class only..

Chaoba: The check output for said it's a patch release, need to revisit what happened.
`Semver-check TAGGED_RELEASE_VERSION:, LAST_VERSION: 10.2.0-beta.1, API_COMPAT_LEVEL: patch
`

tobrun: maybe we can add this to a separate interface specific to accuracy radius functionality?

pengdev: With current design Alternatively we could add extension functions to the interfaces, and within the implementation, cast the interface to actual implementation and use the new function on the implementation class, something like:

```
fun LocationComponent.addAccuracyListener(..) {
  (this as LocationComponentImpl).addAccuracyListener(..)
}
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAccuracyRadiusAnimator.kt
alexander-kulikovskii: this may change in the future.
better to limit visibility as much as we can

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParser.kt
```diff
@@ -0,0 +1,29 @@
...
+  /**
+   * Parse [AttributeSet] to [LocationComponentSettings].
+   *
+   * @param context Context
+   * @param attrs AttributionSet
+   */
+  fun parseLocationComponentAccuracyRadiusSettings(context: Context, attrs: AttributeSet?, pixelRatio: Float = 1.0f): LocationComponentAccuracyRadiusSettings {
+    val typedArray = context.obtainStyledAttributes(attrs, R.styleable.mapbox_MapView, 0, 0)
+    try {
+      return LocationComponentAccuracyRadiusSettings(
```
alexander-kulikovskii, line 20: should we return some default value if smth goes wrong?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
```diff
@@ -10,4 +11,44 @@ import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
...
\ No newline at end of file
+  get() = this.getPlugin(Plugin.MAPBOX_LOCATION_COMPONENT_PLUGIN_ID)!!
+
+/**
+ * Adds a listener that gets invoked when indicator accuracy radius changes.
+ *
+ * @param listener Listener that gets invoked when indicator accuracy radius changes
+ */
+fun LocationComponentPlugin.addOnIndicatorAccuracyRadiusChangedListener(listener: OnIndicatorAccuracyRadiusChangedListener) {
+  (this as LocationComponentPluginImpl).addOnIndicatorAccuracyRadiusChangedListener(listener)
```
yunikkk, line 22: We could introduce separate interface for the `addOnIndicatorAccuracyRadiusChangedListener` / `removeOnIndicatorAccuracyRadiusChangedListener` method, thus users will be able to implement it as well and don't depent on our LocationComponentPluginImpl, described the idea at https://github.com/mapbox/mapbox-maps-android-internal/issues/857#issuecomment-1010950807

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
```diff
@@ -25,7 +24,7 @@ import java.util.concurrent.CopyOnWriteArrayList
  * to the user's current location.
  */
 class LocationComponentPluginImpl : LocationComponentPlugin, LocationConsumer,
-  LocationComponentSettingsBase() {
+  LocationComponentAccuracyRadiusSettingsBase() {
```
kiryldz, line 27: Why do we need that base class if we do not have `override` modifier for `addOnIndicatorAccuracyRadiusChangedListener` / `removeOnIndicatorAccuracyRadiusChangedListener`?

Chaoba: The base class is related to applying settings has no relation with `addOnIndicatorAccuracyRadiusChangedListener / removeOnIndicatorAccuracyRadiusChangedListener`

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParser.kt
Chaoba: In what scenario?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
Chaoba: Whit the extension functions, users can invoke them like this `mapView.location.addOnIndicatorAccuracyRadiusChangedListener()`
If move these functions into a new interface, how can users invoke them?

yunikkk: I mean we can keep the extension function - and at the same time add the interface with `addOnIndicatorAccuracyRadiusChangedListener` / `removeOnIndicatorAccuracyRadiusChangedListener`, and let `LocationComponentPluginImpl` implement this interface. So, this extension function becomes
```
fun LocationComponentPlugin.addOnIndicatorAccuracyRadiusChangedListener(listener: OnIndicatorAccuracyRadiusChangedListener) {
  (this as NewCoolAccuracyInterface).addOnIndicatorAccuracyRadiusChangedListener(listener)
```
This will allow user to implement it as well, if it's needed.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParser.kt
alexander-kulikovskii: but why you add `try` block?
for example some error in parsing color. user can set some strange value.

Chaoba: Guess this try is to take the advantage of `finally` recycling the typedArray. This is the same style in all our parsers. @pengdev Please fix me if I'm wrong.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
Chaoba: Yeah, that make sense.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
```diff
@@ -58,4 +59,44 @@ fun LocationComponentPlugin.createDefault2DPuck(
...
   }
+}
+
+/**
+ * Adds a listener that gets invoked when indicator accuracy radius changes.
+ *
+ * @param listener Listener that gets invoked when indicator accuracy radius changes
+ */
+fun LocationComponentPlugin.addOnIndicatorAccuracyRadiusChangedListener(listener: OnIndicatorAccuracyRadiusChangedListener) {
+  (this as LocationIndicatorAccuracyRadiusHandler).addOnIndicatorAccuracyRadiusChangedListener(listener)
```
yunikkk, line 70: Would we like to do safe cast `(this as? LocationIndicatorAccuracyRadiusHandler)?.addOnIndicatorAccuracyRadiusChangedListener(listener)` instead? 
It may be fine even throwing an exception to user if LocationComponent is NOT our impl but custom one. We just need clear message, explaining why method throws.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
```diff
@@ -39,19 +38,22 @@ class LocationComponentPluginImpl : LocationComponentPlugin, LocationConsumer,
...
 
-  private val onIndicatorPositionChangedListener =
+  private val onIndicatorPositionChangedListeners =
     CopyOnWriteArrayList<OnIndicatorPositionChangedListener>()
 
-  private val onIndicatorBearingChangedListener =
+  private val onIndicatorBearingChangedListeners =
     CopyOnWriteArrayList<OnIndicatorBearingChangedListener>()
 
+  private val onIndicatorAccuracyRadiusChangedListeners =
```
yunikkk, line 47: Any reasons on using `CopyOnWriteArrayList` instead of `CopyOnWriteArraySet`? 
Also same about listeners above

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -81,6 +84,9 @@ internal class LocationProviderImpl(context: Context, var settings: LocationComp
     updateCompass()
   }
 
+  fun updateAccuracyRaidusSettings(accuracyRadiusSettings: LocationComponentAccuracyRadiusSettings) {
```
yunikkk, line 87:
```suggestion
  fun updateAccuracyRadiusSettings(accuracyRadiusSettings: LocationComponentAccuracyRadiusSettings) {
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParser.kt
yunikkk: If exception is not possible here at all - we could just use `.also{}` block, try finally signals that we _could_ face an exception

Chaoba: Created a ticket to track this.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
Chaoba: `LocationIndicatorAccuracyRadiusHandler` is an interface, it will work fine even LocationComponent is a custom one as long as it implements this interface.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt
Chaoba: Not sure, but `CopyOnWriteArraySet` is better, will update this.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
Chaoba: Sharp eyes

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPlugin.kt
```diff
@@ -12,6 +12,7 @@ interface LocationComponentPlugin :
   MapStyleObserverPlugin,
   LifecyclePlugin,
   ContextBinder,
+  LocationComponentAccuracyRadiusSettingsInterface,
```
kiryldz, line 15: isn't that still a breaking change if user implements own version of LocationComponentPlugin?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
yunikkk: Right, but if user does not implement the interface? Seem that can easily happen, many users with updated SDK who has custom plugin implementations (not sure how common is it though) most probably won't implement it. 
Then when this cast crashes it won't give many useful information about the reason. Should we throw an exception with the clear message, like `your custom plugin implementation does not implement the necessary interface` ?

Chaoba: Right, let's add a check for this and this should also fix @kiryldz 's question about API breaking.

yunikkk: Just not sure if this fixes @kiryldz comment, I've had an imression that only LocationPluginImpl should implement the new interface, otherwise it indeed seems a breaking change

heangpov16 (changes requested):
Support me

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPlugin.kt
tobrun: yes, but we should treat plugin interfaces as internal, with v11 we should put them into an internal package. This avoids the breaking change issue while still allowing developers to benefit from our plugin system.

yunikkk: @tobrun I'm missing smth, I guess. You saying that `LocationComponentAccuracyRadiusSettingsInterface` will be internal? And the `LocationComponentPlugin` would still be public?

yunikkk: Here I'm still don't understand why should the LocationComponentPlugin implement the `LocationComponentAccuracyRadiusSettingsInterface`, instead of LocationComponentPluginImpl implementing it, similarly to implementing `LocationIndicatorAccuracyRadiusHandler`

Chaoba: @yunikkk this is your idea https://github.com/mapbox/mapbox-maps-android/pull/1016#discussion_r782999889

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParser.kt
```diff
@@ -0,0 +1,29 @@
+package com.mapbox.maps.plugin.locationcomponent
```
pengdev, line 1: could we still generate this class?

Chaoba: It will need to add new template files and change the generating script in stylegen. Since it is the only parser file like this style, I feel it's not worth generating it.

plugin-locationcomponent/src/test/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParserTest.kt
```diff
@@ -0,0 +1,93 @@
+package com.mapbox.maps.plugin.locationcomponent
```
pengdev, line 1: is this file generated?

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusSettings.kt
```diff
@@ -0,0 +1,22 @@
+package com.mapbox.maps.plugin.locationcomponent
+
+import android.graphics.Color
+/**
+ * Settings for accuracy radius.
+ */
+data class LocationComponentAccuracyRadiusSettings @JvmOverloads constructor(
```
pengdev, line 7: do we want to keep this class as data classes until we revisit usage of data classes in the next iteration?

Chaoba: Yes, I'd like to keep as others just in case introduce other bugs.

plugin-locationcomponent/src/test/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusAttributeParserTest.kt
Chaoba: It's copied from the generated file.

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationAccuracyRadiusConsumer.kt
```diff
@@ -0,0 +1,26 @@
+package com.mapbox.maps.plugin.locationcomponent
+
+import android.animation.ValueAnimator
+
+/**
+ * Defines the interface for LocationAccuracyRadiusConsumer.
+ */
+interface LocationAccuracyRadiusConsumer {
```
pengdev, line 8: could we introduce a `LocationConsumer2` which is declared as 
```
interface LocationConsumer2 : LocationConsumer {
  fun onAccuracyRadiusUpdated(
    vararg radius: Double,
    options: (ValueAnimator.() -> Unit)? = null
  )
  fun onPuckAccuracyRadiusAnimatorDefaultOptionsUpdated(options: ValueAnimator.() -> Unit)
}
```

We have seen similar pattern in the Android lifecycle callback https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/content/ComponentCallbacks2.java, docs [here](https://developer.android.com/reference/android/content/ComponentCallbacks2) 

while the LocationProvider still accepts the LocationConsumer, we could check if it's the extended version in the implementation and call the extended method if it exist? And `LocationConsumer2` is public and could be suggested to end-users if they want to use the extended version with more functions. wdyt? cc @mapbox/maps-android

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentAccuracyRadiusSettingsInterface.kt
```diff
@@ -0,0 +1,35 @@
+package com.mapbox.maps.plugin.locationcomponent
+
+/**
+ * Interface that defines the public accuracy radius settings interface for LocationComponentPlugin.
+ */
+interface LocationComponentAccuracyRadiusSettingsInterface {
```
pengdev, line 6: similarly to above, shall we introduce a `LocationComponentSettingsInterface2` which is declared as 
```
interface LocationComponentSettingsInterface2 : LocationComponentSettingsInterface {
    fun getAccuracyRadiusSettings(): LocationComponentAccuracyRadiusSettings
    fun updateAccuracyRadiusSettings(block: LocationComponentAccuracyRadiusSettings.() -> Unit)
    var accuracyCircleEnabled: Boolean
    var accuracyCircleColor: Int
    var accuracyCircleAlpha: Float
}
```
to extend our existing interface?

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPlugin.kt
pengdev: Another idea is to introduce a `LocationComponentPlugin2` that extends the `LocationComponetPlugin` interface, where our `LocationComponentPluginImpl` will implement `LocationComponentPlugin2`?

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationAccuracyRadiusConsumer.kt
pengdev: This also gives user a signal that we are iterating on these interfaces, and in the next major release, we could deprecate/remove the first version and directly use the second version instead.

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/generated/LocationComponentSettingsInterface2.kt
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Interface that defines the public settings interface for LocationComponentPlugin.
+ */
+interface LocationComponentSettingsInterface2 : LocationComponentSettingsInterface {
+  /**
+   * Get current locationcomponent configuration2.
+   *
+   * @return locationcomponent settings2
+   */
+  fun getSettings2(): LocationComponentSettings2
```
pengdev, line 14: could we customise this name? like getLocationAccuracySettings?

yunikkk (approved):

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentExt.kt
```diff
@@ -12,7 +12,7 @@ import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
 /**
  * Extension val to get the LocationComponentPlugin instance.
  */
-val MapPluginProviderDelegate.location: LocationComponentPlugin
+val MapPluginProviderDelegate.location: LocationComponentPlugin2
```
pengdev, line 15: does our bot complain this is breaking change?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -63,9 +65,16 @@ internal class LocationProviderImpl(context: Context) :
     locationConsumers.forEach { consumer ->
       consumer.onLocationUpdated(Point.fromLngLat(location.longitude, location.latitude))
       consumer.onBearingUpdated(location.bearing.toDouble())
+      if (accuracyRadiusSettings.showAccuracyRing && consumer is LocationConsumer2) {
```
yunikkk, line 68: IMO location provider shouldn't know about settings of the display logic of locations. Then here we could only check `if (consumer is LocationConsumer2) {` and accuracyRadiusSettings can be checked within  location plugin or location puck

Chaoba: Right, we shouldn't handle it in location provider. This is handled in `accuracyRadiusAnimator`

yunikkk (approved):

sdk-base/src/main/java/com/mapbox/maps/plugin/gestures/generated/GesturesSettingsBase.kt
```diff
@@ -12,7 +12,7 @@ import com.mapbox.maps.plugin.ScrollMode
  */
 abstract class GesturesSettingsBase : GesturesSettingsInterface {
   /**
-   * Gesture configuration allows to control the user touch interaction.
+   * The internalSettings that hold all the settings
```
kiryldz, line 15: Why was that changed? Same for non location files above

sdk-base/src/main/java/com/mapbox/maps/plugin/locationcomponent/generated/LocationComponentSettings.kt
```diff
@@ -5,7 +5,7 @@ package com.mapbox.maps.plugin.locationcomponent.generated
 import android.graphics.Color
 import com.mapbox.maps.plugin.LocationPuck
 /**
- * Shows a location puck on the map.
+ * Settings for LocationComponent
```
kiryldz, line 8: Same as https://github.com/mapbox/mapbox-maps-android/pull/1016/files#r801674302

kiryldz:
@Chaoba great work but wonder why generated files for non location component are changed and some javadocs that were pretty well-designed are replaced with some new single one for all the plugins

PR #1033, opened 2022-01-11
[viewport] Add viewport plugin.

pengdev:
### Summary of changes

This PR introduces viewport plugin, allowing to track objects on a map. The viewport plugin provides a structured approach to organising camera management logic into states and transitions between them.

At any given time, the viewport is either:
  - idle (not updating the camera)
  - in a state (camera is being managed by a ViewportState)
  - transitioning (camera is being managed by a ViewportTransition)

Two ViewportState are provided by default that user can create instance from the ViewportPlugin: 
 * `viewport.makeFollowingViewportState(options): FollowingViewportState`
This state allows to follow user's location, and syncs the viewport position with the location puck provided by the LocationComponent.
 * `viewport.makeOverviewViewportState(options): OverviewViewportState`
This state sets the viewport to cover user provided geometry with configurations.

and users are free to add their own ViewportState and use it with the viewport plugin.

Two ViewportTransition are also provided by default that user can create instance from the ViewportPlugin:
* `viewport.makeDefaultViewportTransition(options): DefaultViewportTransition`
The default viewport transition allows user to transition from one state to another state with the default optimised animations.
* `viewport.makeImmediateViewportTransition(): ImmediateViewportTransition`
The immediate viewport transition allows user to transition from one state to another state immediately without animations.

and users are free to add their own ViewportTransition and set to specifically to animate transition from the desired fromState and targetState.

### User impact (optional)

https://user-images.githubusercontent.com/2764714/149017291-845ec676-35d1-4c5e-a551-e5baf6785560.mp4

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Introduce viewport plugin.</changelog>`.


app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
```diff
@@ -0,0 +1,199 @@
...
+      })
+
+    // Let scale gesture to overwrite the followingViewportState's default zoom value.
+    mapView.gestures.addOnScaleListener(object : OnScaleListener {
+      override fun onScaleBegin(detector: StandardScaleGestureDetector) {
+        // no-ops
+      }
+
+      override fun onScale(detector: StandardScaleGestureDetector) {
+        val zoomBy = ln(detector.scaleFactor.toDouble()) / ln(PI / 2) * SCALE_FACTOR
```
kiryldz, line 149: where do those calculations come from?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
```diff
@@ -0,0 +1,199 @@
...
+import com.mapbox.maps.testapp.utils.SimulateRouteLocationProvider
+import kotlin.math.PI
+import kotlin.math.ln
+
+/**
+ * Showcase the use age of viewport plugin with advanced gestures customisation.
+ *
+ * Touch the map to toggle the following and overview mode.
+ */
+class AdvancedViewportGesturesExample : AppCompatActivity() {
```
kiryldz, line 45: Should it also be marked experimental as another activity? 
Or I'd rather not mark activities as experimental at all

app/src/main/java/com/mapbox/maps/testapp/utils/NavigationSimulator.kt
```diff
@@ -42,146 +37,35 @@ class NavigationSimulator(
...
-      }
-    }
-
-    override fun onMove(detector: MoveGestureDetector): Boolean {
-      return false
-    }
-
-    override fun onMoveEnd(detector: MoveGestureDetector) {}
+    viewportPlugin.defaultTransition = viewportPlugin.makeDefaultTransition(
+      DefaultViewportTransitionOptions.Builder().maxDurationMs(2000).build()
```
kiryldz, line 54: perhaps move 2000 to const making it easier to change if needed

app/src/main/java/com/mapbox/maps/testapp/utils/NavigationSimulator.kt
```diff
@@ -42,146 +37,35 @@ class NavigationSimulator(
...
-          CameraAnimatorOptions.cameraAnimatorOptions(camera.zoom!!)
-        ) {
-          duration = EASE_TO_PUCK_DURATION_MS / 3
-          interpolator = AccelerateDecelerateInterpolator()
+    with(viewportPlugin.status) {
+      if (this is ViewportStatus.State) {
+        when (state) {
+          followingViewportState -> viewportPlugin.transitionTo(overviewViewportState)
+          overviewViewportState -> viewportPlugin.transitionTo(followingViewportState)
+          else -> viewportPlugin.transitionTo(followingViewportState)
```
kiryldz, line 64: is `else` needed? it seems to be a sealed class

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportExt.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.mapbox.maps.MapboxExperimental
+import com.mapbox.maps.plugin.Plugin
+import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
+
+/**
+ * Extension val for MapView to get the [ViewportPlugin] instance.
+ */
+@MapboxExperimental
+val MapPluginProviderDelegate.viewport: ViewportPlugin
```
kiryldz, line 13: I remember it was named `experimentalViewport` at some point, just noting it here (although I do not remember why it was done like that but I like current solution more).

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,331 @@
...
+ *
+ * It provides a structured approach to organizing camera management logic into states and transitions between them.
+ *
+ * at any given time, the viewport is either:
+ *  - idle (not updating the camera)
+ *  - in a state (camera is being managed by a ViewportState)
+ *  - transitioning (camera is being managed by a ViewportTransition)
+ */
+@MapboxExperimental
+class ViewportPluginImpl : ViewportPlugin {
```
kiryldz, line 40: shouldn't our impl be `internal`? I guess we are trying to following such logic for other plugins

kiryldz: in that case it's better to mark `ViewportPlugin` interface as experimental and do not explicitly mark it here

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,331 @@
...
+  private val registeredStatusObservers = CopyOnWriteArraySet<ViewportStatusObserver>()
+  private var currentCancelable: Cancelable? = null
+    @Synchronized set
+    @Synchronized get
```
kiryldz, line 44: why synchronized?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,331 @@
...
+      )
+    }
+  }
+
+  /**
+   * DefaultViewportTransition with default options
+   *
+   * This transition is used unless overridden by one of the registered transitions.
+   */
+  override lateinit var defaultTransition: ViewportTransition
```
kiryldz, line 209: couldn't it lead to an easy reproducible crash if this variable gets accessed if it was not initialized?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,331 @@
...
+    cameraPlugin.removeCameraAnimationsLifecycleListener(cameraAnimationsLifecycleListener)
+  }
+
+  internal companion object {
+    /**
+     * Constant used to recognize the owner of transitions initiated by the [ViewportPluginImpl].
+     *
+     * @see CameraAnimator.owner
+     */
+    internal const val VIEWPORT_CAMERA_OWNER = "VIEWPORT_CAMERA_OWNER"
```
kiryldz, line 323: nit: no need for another `internal` here

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,331 @@
...
+  }
+}
+
+internal fun ViewportStatus.getCurrentOrTargetState(): ViewportState? =
```
kiryldz, line 327: I'm seeing it used across this PR for several times - perhaps expose it and re-use once?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
```diff
@@ -0,0 +1,238 @@
...
+import com.mapbox.maps.plugin.viewport.data.FollowingViewportStateOptions
+import com.mapbox.maps.plugin.viewport.transition.MapboxViewportTransitionFactory
+import java.util.concurrent.*
+
+/**
+ * The actual implementation of [FollowingViewportState] that follows user's location.
+ *
+ * Note: [LocationComponentPlugin] should be enabled to use this viewport state.
+ */
+class FollowingViewportStateImpl internal constructor(
```
kiryldz, line 27: same as above - shouldn't we mark our impl as `internal`?

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
pengdev: These are from our gestures plugin.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
```diff
@@ -0,0 +1,238 @@
...
+    dataSourceUpdateObservers.add(viewportStateDataObserver)
+    return Cancelable {
+      dataSourceUpdateObservers.remove(viewportStateDataObserver)
+      removeIndicatorListenerIfNeeded()
+    }
+  }
+
+  private fun checkLocationComponentEnablement() {
+    if (!locationComponent.enabled) {
+      Logger.w(
```
kiryldz, line 129: Perhaps throw runtime exception? Seems like a severe violation to me

app/src/main/java/com/mapbox/maps/testapp/utils/NavigationSimulator.kt
pengdev: state is not a sealed class, and it can be null, but maybe I can combine the overviewViewportState branch with else branch.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
```diff
@@ -0,0 +1,238 @@
...
+            }
+          )
+          doOnEnd { finishAnimation(this) }
+        },
+      instant
+    )
+  }
+
+  private companion object {
+    private const val TAG = "FollowingViewportStateImpl"
```
kiryldz, line 236: nit: no need for another private

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/OverviewViewportStateImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.mapbox.maps.plugin.animation.camera
+import com.mapbox.maps.plugin.delegates.MapDelegateProvider
+import com.mapbox.maps.plugin.viewport.data.OverviewViewportStateOptions
+import com.mapbox.maps.plugin.viewport.transition.MapboxViewportTransitionFactory
+import java.util.concurrent.*
+
+/**
+ * The actual implementation of [OverviewViewportState] that shows the overview of a given geometry.
+ */
+class OverviewViewportStateImpl(
```
kiryldz, line 17: same as above - perhaps mark whole our impl as internal

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportExt.kt
pengdev: experimentalViewport is another package that's used for prototyping from scratch in parallel with the initial version, and has been promoted in #1032

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
kiryldz: btw - why do we even need this? our viewport plugin requires location + camera as implementations, I guess we should rather have a check that location component is _enabled_.
or alternatively similar logic with similar warning should be added for camera plugin dependency as well

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: if you remember, we have a compile dependency to plugins on the sdk module to initialise the plugin with concrete implementation, so we have to keep it public.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: this could be modified on different threads.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/DefaultViewportTransitionImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+import com.mapbox.maps.plugin.animation.camera
+import com.mapbox.maps.plugin.delegates.MapDelegateProvider
+import com.mapbox.maps.plugin.viewport.CompletionListener
+import com.mapbox.maps.plugin.viewport.data.DefaultViewportTransitionOptions
+import com.mapbox.maps.plugin.viewport.state.ViewportState
+
+/**
+ * The implementation of [DefaultViewportTransition] that transitions Viewport from one [ViewportState] to another.
+ */
+class DefaultViewportTransitionImpl(
```
kiryldz, line 18: same as above - perhaps mark whole our impl as internal

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/ImmediateViewportTransition.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.mapbox.maps.plugin.animation.camera
+import com.mapbox.maps.plugin.delegates.MapDelegateProvider
+import com.mapbox.maps.plugin.viewport.CompletionListener
+import com.mapbox.maps.plugin.viewport.ViewportPluginImpl.Companion.VIEWPORT_CAMERA_OWNER
+import com.mapbox.maps.plugin.viewport.state.ViewportState
+
+/**
+ * [ViewportTransition] that transition to target [ViewportState] immediately.
+ */
+class ImmediateViewportTransition internal constructor(delegateProvider: MapDelegateProvider) :
```
kiryldz, line 14: same as above - perhaps mark whole our impl as internal

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
```diff
@@ -0,0 +1,278 @@
...
+  mapDelegateProvider: MapDelegateProvider
+) {
+  private val cameraManager = mapDelegateProvider.mapCameraManagerDelegate
+  private val cameraPlugin = mapDelegateProvider.mapPluginProviderDelegate.camera
+
+  fun transitionFromLowZoomToHighZoom(
+    cameraOptions: CameraOptions,
+    transitionMaxDurationMs: Long
+  ): AnimatorSet {
+    val animators = mutableListOf<Animator>()
```
kiryldz, line 29: I remember we were talking about it - can we use `CameraAnimator`s here?

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
```diff
@@ -0,0 +1,278 @@
...
+    var centerDuration = 0L
+    cameraOptions.center?.let { center ->
+      val screenDistanceFromMapCenterToLocation =
+        screenDistanceFromMapCenterToTarget(
+          cameraManager = cameraManager,
+          currentCenter = currentMapCameraState.center,
+          targetCenter = center
+        )
+
+      val centerAnimationRate = 500L
```
kiryldz, line 41: can we move this (and all similar vals below in this file) to const? why do we need to declare those vals each time

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/transition/ViewportTransition.kt
```diff
@@ -0,0 +1,31 @@
+package com.mapbox.maps.plugin.viewport.transition
+
+import com.mapbox.maps.plugin.animation.Cancelable
+import com.mapbox.maps.plugin.viewport.CompletionListener
+import com.mapbox.maps.plugin.viewport.state.ViewportState
+
+/**
+ * Defines how a [ViewportState] transition to another [ViewportState].
+ */
+interface ViewportTransition {
```
kiryldz, line 10: could be `fun interface`?

kiryldz (commented):
overall comment (as stated in files directly) - I would hide all our Mapbox implementations (mark them internal) and remove javadoc in impl files (as they duplicate documentation in public interfaces)

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
kiryldz: looks not very clear, perhaps drop a comment here as it's an example

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
pengdev: updated

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportExt.kt
pengdev: This should be the final API that users see.

app/src/main/java/com/mapbox/maps/testapp/utils/NavigationSimulator.kt
pengdev: combined the overviewViewportState branch with else branch.

app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt
pengdev: Added some comments here

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: There's no easy way around it, as the initialisation of DefaultViewportTransition need to be async with delegate provider. Also as these plugins are initialised and onDelegateProvider are available already inside the MapView's constructor, when users access plugin from `mapView.viewport`, the defaultTransition should be ready to used.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: It was specifically removed from previous code reviews, because this extension seems unclear or have very limited value on specific use cases. User could implement this logic easily as well.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
pengdev: changed to internal class.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
pengdev: I didn't make it a fatal error as user could decide to enable the location component later, then the viewport will sync with the puck. I see it more as a warning that the following viewport data source has stopped producing data.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/OverviewViewportStateImpl.kt
pengdev: done

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/DefaultViewportTransitionImpl.kt
pengdev: done

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
pengdev: imo, if we don't specifically use the properties defined in CameraAnimator, the generic Animator would be more clean, compared with `mutableListOf<CameraAnimator<*>>`

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/transition/ViewportTransition.kt
pengdev: 🤔 I'm not sure as this interface is not supposed to be used as a listener which SAM would be more useful. It's more used as an interface where user's complex transition logic class conforms, but would like @mapbox/maps-android 's thoughts as well.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
kiryldz: maybe `ValueAnimator` then?

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
```diff
@@ -0,0 +1,444 @@
...
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+@MapboxExperimental
```
kiryldz, line 37: not sure tests should be marked as experimental but up to you

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
```diff
@@ -0,0 +1,444 @@
...
+        testObserver.onViewportStatusChanged(
+          ViewportStatus.Transition(defaultTransition, null, targetState),
+          ViewportStatus.State(null),
+          VIEWPORT_STATUS_OBSERVER_REASON_PROGRAMMATIC
+        )
+      }
+    }
+  }
+
+  private fun runHandlerAndTest(test: () -> Unit) {
```
kiryldz, line 439: very nice

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/transition/TransitionUtilsTest.kt
```diff
@@ -0,0 +1,192 @@
...
+
+    assertEquals(expected, actual)
+  }
+
+  @Test
+  fun `test constraintDurationTo - no adjustments`() {
+    val animators = listOf<Animator>(
+      ValueAnimator.ofFloat().apply {
+        startDelay = 700
+        duration = 1300
```
ank27, line 139: nit : we can move all these `durations` to constant.

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
```diff
@@ -0,0 +1,444 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+@MapboxExperimental
+class ViewportPluginImplTest {
+  private val delegateProvider = mockk<MapDelegateProvider>(relaxed = true)
+  private val mapPluginProviderDelegate = mockk<MapPluginProviderDelegate>(relaxed = true)
+  private val cameraPlugin = mockk<CameraAnimationsPlugin>(relaxed = true)
+  private val locationPlugin = mockk<LocationComponentPlugin>(relaxed = true)
```
kiryldz, line 42: I understand we're a bit short on time but using fully relaxed mocks (and not `relaxedUnitFun`) may hide some issues. Perhaps we could mock a bit more if it's not a big burden and relax only unit functions

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImplTest.kt
```diff
@@ -0,0 +1,243 @@
...
+import io.mockk.verifySequence
+import org.junit.After
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
kiryldz, line 41: Is it required here?

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/state/OverviewViewportStateImplTest.kt
```diff
@@ -0,0 +1,187 @@
...
+import io.mockk.verifySequence
+import org.junit.After
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
kiryldz, line 35: is it required here?

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactoryTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    val valueSlot = slot<CameraAnimatorOptions<Double>>()
+    every {
+      cameraPlugin.createBearingAnimator(capture(valueSlot), any(), any())
+    } returns mockk()
+    transitionsFactory.transitionFromLowZoomToHighZoom(
+      cameraOptions,
+      DEFAULT_STATE_TRANSITION_MAX_DURATION_MS
+    )
+
+    assertEquals(-10.0, valueSlot.captured.targets.last(), 0.0000000001)
```
kiryldz, line 74: move to const EPS

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
```diff
@@ -0,0 +1,278 @@
...
+        ),
+        startDelay = 600L,
+        duration = 1200L
+      )
+    }
+
+    cameraOptions.pitch?.let { pitch ->
+      animators += createPitchAnimator(
+        pitch = pitch,
+        duration = 1000L
```
ank27, line 139: we can use `LINEAR_ANIMATION_DURATION_MS` here, or probably change the name to `ANIMATION_DURATION_MS` so it can be used for all the constants.

ank27: i see @kiryldz already commented above the same, removing it.

kiryldz (approved):
Dropped some questions / nits but nothing that critical. If needed we can iterate later.
Huge amount of work @pengdev, great job!

ank27 (approved):

ank27:
Looks good, awesome work @pengdev 🚀 🦸

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: In v11 I think we could revisit the plugin setup and inject the required delegateProvider etc in the plugin implementation's constructor, as we got rid of the reflections and don't require an empty constructor.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
pengdev: I can note it down and try to address it after rc, although I also don't see much added value, as we only use this animators list to construct AnimatorSet and call `AnimatorSet.playTogether(Collection<Animator> items)`.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
```diff
@@ -0,0 +1,341 @@
...
+ *
+ * It provides a structured approach to organizing camera management logic into states and transitions between them.
+ *
+ * at any given time, the viewport is either:
+ *  - idle (not updating the camera)
+ *  - in a state (camera is being managed by a ViewportState)
+ *  - transitioning (camera is being managed by a ViewportTransition)
+ */
+@MapboxExperimental
+class ViewportPluginImpl(private val handler: Handler = Handler(Looper.getMainLooper())) :
```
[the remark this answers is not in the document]
pengdev: same has been suggested by @alexander-kulikovskii 

I don't think this constructor would be used as public API, as the initialisation of plugin instance is done internally in our sdk(but in another module, so need to keep this class public.)

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
pengdev: Added to avoid lint warnings on all the methods in the test 🙈

pengdev: Will remove when we remove the experimental of viewport API.

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
pengdev: Yeah, it's mainly used to speed up writing these tests to ship in beta, will revisit tests after the beta release.

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImplTest.kt
pengdev: Yes, it is required as we used Android interpolators in the code.

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/state/OverviewViewportStateImplTest.kt
pengdev: same as above.

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactoryTest.kt
pengdev: done

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: The cancellable is updated internally from user started transition and animation end listeners, the `Synchronized` is added due to a previous code review.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/state/FollowingViewportStateImpl.kt
```diff
@@ -0,0 +1,244 @@
...
+    if (isObservingLocationUpdates && dataSourceUpdateObservers.isEmpty() && !isFollowingStateRunning) {
+      locationComponent.removeOnIndicatorPositionChangedListener(indicatorPositionChangedListener)
+      locationComponent.removeOnIndicatorBearingChangedListener(indicatorBearingChangedListener)
+      isObservingLocationUpdates = false
+    }
+  }
+
+  /**
+   * Describes the configuration options of the state.
+   */
```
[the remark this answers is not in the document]
pengdev: imo more documentation is better than no documentation, but yeah, agree the docs could diverge if we don't pay attention, will do the cleanup as follow up after beta.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
pengdev: these factory utilities are ported from nav sdk, although I see the point of keeping them out of constants, as these are very specific value corresponding to specific situations/functions. And most of these values are not reused by other animators, so I'd prefer to leave it as is.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
pengdev: Decide to remove after offline sync with @kiryldz , as users should use our APIs from main thread only.

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/transition/ViewportTransition.kt
pengdev: added it to my todo list between beta and rc. will think more about it.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/transition/MapboxViewportTransitionFactory.kt
kiryldz: let's leave it current way, I don't have that strong opinion about it, was more curious.

plugin-viewport/src/main/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImpl.kt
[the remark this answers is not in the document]
kiryldz: agree

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/ViewportStatusObserver.kt
```diff
@@ -0,0 +1,15 @@
+package com.mapbox.maps.plugin.viewport
+
+/**
+ * Observer that gets notified whenever [ViewportStatus] changes.
+ */
+fun interface ViewportStatusObserver {
+  /**
+   * Called whenever [ViewportStatus] changes.
+   *
+   * @param from The previous [ViewportStatus], null if the previous [ViewportStatus] is IDLE.
```
yunikkk, line 10: nit : null is not possible here, probably `ViewportStatus.State(null)` should be here instead.

If IDLE state would be reworked after beta probably won't be relevant anymore

plugin-viewport/src/test/kotlin/com/mapbox/maps/plugin/viewport/ViewportPluginImplTest.kt
pengdev: resolved in https://github.com/mapbox/mapbox-maps-android/pull/1074

sdk-base/src/main/java/com/mapbox/maps/plugin/viewport/ViewportStatusObserver.kt
pengdev: Resolved in https://github.com/mapbox/mapbox-maps-android/pull/1069

Issue #1044, opened 2022-01-12
Issues with MapView#queueEvent method


kiryldz:
## Environment

- Android OS version: any 
- Devices affected: any
- Maps SDK Version: v10.2.0 and below

## Observed behavior and steps to reproduce

Most common use-case to use `MapView#queueEvent` is basically running something on render thread with GL context (e.g. find out what GPU is used). However currently this API seems unstable especially when calling it somewhere in activity's `onCreate` method. It may lead to crash / not execute user's runnable etc.

## Expected behavior

User's runnable should be executed properly or correct warn messages should be printed to log if render thread is not in valid state.

PR #1068, opened 2022-01-20
Refactor scheduling events in render thread

kiryldz:
### Summary of changes

`MapView#queueEvent` will be executed properly on Mapbox render thread. Previously scheduling such events e.g. when render was not created properly could lead to a crash / be skipped if running OpenGL ES commands there.

Additionally internal render scheduling logic was refactored / fixed:
 - in case of any EGL setup issues rescheduling will happen in small delay and not immediately in order not to overload render thread message queue
 - user events scheduled by `MapView#queueEvent(runnable, needRender = true | false)` will be executed on render thread even if it's not fully prepared. In this case such events will keep to get re-scheduled with small delay (50 ms) until render thread is either ready or killed.
 - any user render events scheduled by `MapView#queueEvent(runnable, needRender = true)` will be executed after map draw calls but before swapping buffers
 - non-render events (both scheduled by `MapClient#scheduleTask` and `MapView#queueEvent(runnable, needRender = false)`) now have a check to verify render thread is fully ready. 
 - extra render call will be always scheduled now when app is brought back to front.
 
### User impact (optional)

Some data collected by `gfxinfo` by running benchmarks changing map camera.

Average values for `easeTo` benchmark (5 runs, infinite moving camera between 3 points for 45 seconds).

|metric|main branch| current refactor| diff |
| ----- | ----- | ----- | ----- |
| jankyFrames                 | 3.0          | 2.75         | -8.3333%     |
| numberSlowUiThread          | 2.5          | 2.0          | -20.0%       |
| numberFrameDeadlineMissed   | 2.5          | 2.25         | -10.0%       |
| jankyFramesPercent          | 0.21         | 0.193        | -8.0952%     |
| totalFramesRendered         | 1406.5       | 1422.5       | +1.1376%     |

Average values for `flyTo` benchmark (5 runs, infinite moving camera between 3 points for 45 seconds).

|metric|main branch| current refactor| diff |
| ----- | ----- | ----- | ----- |
| jankyFrames                 | 3.25         | 3.25         | 0%           |
| numberSlowUiThread          | 2.75         | 2.5          | -9.0909%     |
| numberFrameDeadlineMissed   | 2.75         | 2.75         | 0%           |
| jankyFramesPercent          | 0.245        | 0.242        | -1.2245%     |
| totalFramesRendered         | 1332.5       | 1335.75      | +0.2439%     |

All the metrics look slightly better with refactor introduced in this PR although it's hard to gather very stable ones. Main idea is that code looks cleaner and less redundant actions are performed in rendering logic.

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix dropping / crashing in user events scheduled on a render thread with MapView#queueEvent. Refactor scheduling logic for render thread in general slightly improving rendering performance.</changelog>`.

Fixes: #1044 


sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -120,23 +123,30 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     }
     if (!surface.isValid) {
       Logger.w(TAG, "EGL was configured but surface is not valid.")
-      postPrepareRenderFrame()
+      // give system a bit of time and try rendering again hoping surface will be valid now
+      postPrepareRenderFrame(delayMillis = RETRY_DELAY_MS)
```
yunikkk, line 127: Do we know existing cases when this happens?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -120,23 +123,30 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
     if (creatingSurface) {
       eglCore.makeNothingCurrent()
     }
-    eglSurface = eglCore.createWindowSurface(surface)
-    if (!eglCore.eglStatusSuccess) {
-      // Set EGL Surface as EGL_NO_SURFACE and try recreate it in next iteration.
-      eglSurface = EGL10.EGL_NO_SURFACE
-      postPrepareRenderFrame()
-      return false
+    if (eglSurface == null || eglSurface == EGL10.EGL_NO_SURFACE) {
```
yunikkk, line 135: nit : maybe we don't need nullability since, let EglCore return EGL10.EGL_NO_SURFACE in case of errors? WIll simplify checks here

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -120,23 +123,30 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+      eglSurface = eglCore.createWindowSurface(surface)
+      if (!eglCore.eglStatusSuccess) {
+        // Set EGL Surface as EGL_NO_SURFACE and try recreate it in next iteration.
+        eglSurface = EGL10.EGL_NO_SURFACE
```
yunikkk, line 139: Also this line will go off

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -177,8 +178,21 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       return
     }
     mapboxRenderer.onDrawFrame()
+    // assuming render event queue holds user's runnables with OpenGL ES commands
+    // it makes sense to execute them after drawing a map but before swapping buffers
+    if (renderEventQueue.isNotEmpty()) {
```
yunikkk, line 183: IMHO drainQueue still looked better(

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: no

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: we can not do that as we can't mock properly EGLSurface in tests then

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: we could not re-use one method now for draining queue for snapshot events (Runnable) and draining a queue for custom class (RenderEvent with nullable runnable field)

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -303,10 +316,17 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
-        renderEventQueue.clear()
+        // we clean only Mapbox events to avoid outdated runnables associated with previous EGL context
+        val iterator = renderEventQueue.iterator()
+        while (iterator.hasNext()) {
+          val next = iterator.next()
+          if (next.eventType == EventType.MAPBOX) {
+            iterator.remove()
+          }
+        }
         snapshotQueue.clear()
```
yunikkk, line 327: What about snapshots though? Should we clear them (e.g. if snapshot was called prior to onSurfaceCreated)

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: as we can't re-use it and taking into account this logic is called once in code - I see no sense in leaving that method

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: imo yes
making a snapshot looks to me as an atomic operation - user requests current state of the map. If there's not map atm (surface not created) - we should return null. I'll check this case though, it might not work as expected now I guess (such runnables could be just silently discarded)

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: Ok, haven't realized the difference

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: Seems tests could be tweaked to mock the necessary static fields, e.g. 
```
    mockkStatic(EGL10::class)
    val eglSurface = mockk<EGLSurface>()
    every { EGL10.EGL_NO_SURFACE } returns eglSurface
    Assert.assertEquals(mapboxRenderThread.eglSurface, eglSurface)
```
instead of checking `mapboxRenderThread.eglSurface` for null

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -177,8 +178,21 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
     }
     mapboxRenderer.onDrawFrame()
+    // assuming render event queue holds user's runnables with OpenGL ES commands
+    // it makes sense to execute them after drawing a map but before swapping buffers
+    if (renderEventQueue.isNotEmpty()) {
+      renderEventQueue.forEach {
+        it.runnable?.run()
+      }
+      renderEventQueue.clear()
+    }
```
Chaoba, line 188: Can make `renderEventQueue` the same style as `snapshotQueue`?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -60,14 +60,46 @@ internal abstract class MapboxRenderer : MapClient {
...
   @AnyThread
   override fun scheduleTask(task: Task) {
-    renderThread.queueEvent {
-      task.run()
-    }
+    renderThread.queueRenderEvent(
+      RenderEvent(
+        runnable = { task.run() },
+        needRender = false,
+        eventType = EventType.MAPBOX
```
alexander-kulikovskii, line 78: nit:
maybe replace with `SYSTEM` or `SDK`

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -405,9 +423,10 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     mapboxRenderer.map = null
   }
 
-  companion object {
-    private const val TAG = "Mbgl-RenderThread"
-    private val ONE_SECOND_NS = 10.0.pow(9.0).toLong()
-    private val ONE_MILLISECOND_NS = 10.0.pow(6.0).toLong()
+  private companion object {
+    const val TAG = "Mbgl-RenderThread"
+    const val RETRY_DELAY_MS = 50L
```
alexander-kulikovskii, line 428: will it work for all devices?
in my experience, such delays don't work well on different devices, especially older ones

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: Ok, I'm wrong here, mockk doesn't allow java static fields to be mocked sadly. There's a hack with `mockkObject` which still doesn't work though

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: we don't care about actual time limit. in any case current logic is to re-schedule infinitely. delay is needed to free render message queue a bit

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
kiryldz: nice idea, thanks

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: sadly no as they have different type (Runnable vs RenderEvent)

sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
```diff
@@ -0,0 +1,6 @@
+package com.mapbox.maps.renderer
+
+internal enum class EventType {
```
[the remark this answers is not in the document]
kiryldz: sure, will do

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
axti: Can you explain why you choose **50ms** please? 
IIRC one frame is around 33ms, should have 1-2 frames delay?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -60,14 +60,48 @@ internal abstract class MapboxRenderer : MapClient {
...
+      )
+    )
   }
 
   @AnyThread
   override fun scheduleTask(task: Task) {
-    renderThread.queueEvent {
-      task.run()
-    }
+    Logger.e("KIRYLDD", "scheduleTask!")
```
alexander-kulikovskii, line 75: maybe change tag? :)

kiryldz: I'll take another look at it tomorrow on other devices and then remove 😄 
Will mark PR as draft for now

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: refactored snapshot runnables to be actual render runnables

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: @axti I guess I did answer it in https://github.com/mapbox/mapbox-maps-android/pull/1068#discussion_r788845168 😄 In general 50 ms is just a number out of my head. Sticking to 33 ms makes no sense as:
1. 90 / 120 Hz screens are present today meaning 33 ms is not a strict constant.
2. Frame time is not connected with actual Android SDK timings, we give that delay for system try to recover after some driver errors (e.g. could not obtain EGL surface, set EGL context) or e.g. waiting for Android to provide us new `Surface` object.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -60,14 +60,48 @@ internal abstract class MapboxRenderer : MapClient {
 
   @AnyThread
   override fun scheduleRepaint() {
-    renderThread.requestRender()
+    Logger.e("KIRYLDD", "scheduleRepaint!")
+    renderThread.queueRenderEvent(
+      RenderEvent(
```
yunikkk, line 65: Maybe create single object instance `val repaintRenderEvent = RenderEvent()` since we don't need distinct instances?

kiryldz: there could be multiple instances obviously coming from any places and any threads. In that case it's rather logical to create some pool but I don't want to overcomplicate it

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: @kiryldz please add this explanation as a comment to avoid confusion in future

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
kiryldz: However indeed, in this particular case we can create it once. Thanks for catching!

Chaoba (approved):
LGTM

ank27 (approved):

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
```diff
@@ -47,14 +44,53 @@ internal abstract class MapboxRendererTest {
...
+    var actualEventType: EventType? = null
+    every { renderThread.queueRenderEvent(any()) } answers {
+      actualNeedRender = firstArg<RenderEvent>().needRender
+      actualEventType = firstArg<RenderEvent>().eventType
+    }
     mapboxRenderer.scheduleTask(task)
-    verify { renderThread.queueEvent(any()) }
+    verify {
+      // no idea how to verify Task got transformed in correct lambda so checking call + desired properties
+      renderThread.queueRenderEvent(any())
```
yunikkk, line 71: Could capture runnable and call it, then check if task will be called

yunikkk: e.g.
```
  var captureTo: RenderEvent
  renderThread.queueRenderEvent(capture(captureTo))
  capture.runnable.run()
```

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -291,96 +317,124 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+      // TODO ideally to be fixed in core
+      if (renderEvent.eventType == EventType.SDK) {
+        if (renderThreadPrepared) {
+          postPrepareRenderFrame()
+        }
+      } else {
+        postPrepareRenderFrame()
+      }
+    } else {
+      // no sense in scheduling non-render SDK tasks if thread is not fully ready for render
```
yunikkk, line 389: But render SDK events are NOT added also, right? Maybe simplify / squash both comments to some short and easy one?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -291,96 +317,124 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
-      eventQueue.add(runnable)
-    } else {
-      renderHandlerThread.post {
-        // at the time we start executing surface may be already destroyed
-        if (!shouldExit) {
-          runnable.run()
-        }
-      }
-    }
+  fun queueSnapshot(snapshotTask: Runnable) {
```
yunikkk, line 417: Maybe remove this one too and just post RenderEvent from the calling code?

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
```diff
@@ -39,8 +46,12 @@ internal class RenderHandlerThread {
     }
   }
 
-  fun clearMessageQueue() {
-    handler?.removeCallbacksAndMessages(null)
+  fun clearMessageQueue(clearAll: Boolean = true) {
```
yunikkk, line 49: IMHO separate `clearSdkMessages` / `clearAllMessages` will be more clear, up to you

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -86,67 +84,98 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+    renderHandlerThread.postDelayed(
+      {
+        prepareRenderFrame()
+      },
+      delayMillis
+    )
   }
 
-  private fun checkSurfaceReady(creatingSurface: Boolean): Boolean {
+  private fun isValidRenderingEnvironment(creatingSurface: Boolean): Boolean {
```
yunikkk, line 99: Suggest to rename since it actually inits and not only checks.
`check` / `prepare` probably

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -35,36 +33,35 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  internal val renderTimeNs = AtomicLong(0)
-  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  private var needRenderOnResume = false
+  internal var renderTimeNs = 0L
   private var expectedVsyncWakeTimeNs = 0L
-  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-  internal val awaitingNextVsync = AtomicBoolean(false)
+  private var awaitingNextVsync = false
   private var sizeChanged = false
```
yunikkk, line 50: sizeChanged should be volatile as well? Used from different threads as well

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -86,67 +84,98 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+          }
+          // it's safe to use !! here as we checked surface above
+          val eglSurfaceOk = checkEglSurface(surface!!)
+          if (eglSurfaceOk) {
+            val eglContextOk = checkEglContextCurrent()
+            // finally we can create native renderer if needed or just report OK
+            if (eglContextOk) {
+              if (!nativeRenderCreated) {
+                mapboxRenderer.onSurfaceCreated()
+                mapboxRenderer.onSurfaceChanged(
```
yunikkk, line 118: nit : may just set `surfaceSizeChanged` to true - will be set later in `checkSurfaceSizeChanged`

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -291,96 +317,124 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+        processAndroidSurface(surface, width, height)
       }
       createCondition.await()
     }
   }
 
   @AnyThread
   fun setMaximumFps(fps: Int) {
-    renderTimeNs.set(ONE_SECOND_NS / fps)
+    renderTimeNs = ONE_SECOND_NS / fps
```
yunikkk, line 359: Maybe update it similarly to size - posting runnable to the render thread? No need for volatile then

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -86,67 +84,98 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+      {
+        prepareRenderFrame()
+      },
+      delayMillis
+    )
   }
 
-  private fun checkSurfaceReady(creatingSurface: Boolean): Boolean {
+  private fun isValidRenderingEnvironment(creatingSurface: Boolean): Boolean {
     lock.withLock {
```
yunikkk, line 100: Wonder if we can skip all the checks if for example `renderThreadPrepared` is true here

alexander-kulikovskii (approved):

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: doesn't change anything I guess, could leave as is

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: done

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: imo code is easier with volatile, in general does not matter that much but slightly easier tests now

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: didn't get it fully - although condition is the same but in one case it's `postNonRenderEvent(renderEvent)`, in other `postPrepareRenderFrame()` - I'll need some lambda param then, not sure it's worth it

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
yunikkk: Well capture works https://github.com/mapbox/mapbox-maps-android/pull/1068/files#diff-57f2dab0d3da1120c8042e3b8e54ae41eb8ef84734ba6f77e51dc373ef33ac59R268 right? Wouldn't the same work here?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -408,9 +468,42 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     mapboxRenderer.map = null
   }
 
+  private fun clearQueue(lock: ReentrantLock, queue: LinkedList<RenderEvent>) {
+    // we clean only Mapbox events to avoid outdated runnables associated with previous EGL context
+    lock.withLock {
+      // using iterator to avoid concurrent modification exception
```
yunikkk, line 474: Could just use kotlin ext `MutableList<T>.removeAll(predicate: (T) -> Boolean)`

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: Ok indeed

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -219,43 +260,46 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+        return
+      }
     }
     checkSurfaceSizeChanged()
-    drainQueue(renderEventQueue)
-    // listen to next VSYNC event if not listening already
-    if (awaitingNextVsync.compareAndSet(false, true)) {
-      Choreographer.getInstance().postFrameCallback(this)
-    }
+    Choreographer.getInstance().postFrameCallback(this)
```
yunikkk, line 301: Is it still possible awaitingNextVsync will still be false but frame callback will already fire?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: as said before in other place - requires API = 24

yunikkk: Not really, API 24 is required for java member function 
https://developer.android.com/reference/java/util/Collection#removeIf(java.util.function.Predicate%3C?%20super%20E%3E)
but there's Kotlin extension available for any API lvl
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-all.html

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 Mapbox welcomes participation and contributions from everyone.
 
+# 10.4.0-beta.1
+
+## Features ✨ and improvements 🏁
+* Refactor scheduling logic for render thread in general slightly improving rendering performance. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
+
+## Bug fixes 🐞
+* Fix dropping / crashing in user events scheduled on a render thread with `MapView#queueEvent`. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
```
yunikkk, line 11: nit : `dropping frames` or what is dropping about ?

yunikkk: Ah, it's `dropping events`

yunikkk: Maybe phrase is somehow more easy to understand

yunikkk (approved):

Issue #1076, opened 2022-01-25
Black screen if using emulator with MSAA enabled


kiryldz:
## Environment

- Android OS version: confirmed on 29, but assume that any
- Devices affected: any (?) emulator
- Maps SDK Version: v10.3.0-beta.1

## Observed behavior and steps to reproduce

Run activity on emulator and create `MapView` as:
```
val mapView = MapView(this, MapInitOptions(this, antialiasingSampleCount = 4))
```

No suitable EGL configs found and map is rendered black.

## Expected behavior

Map should still be rendered.  Correct solution will be to disable MSAA, log it as a warning and prepare EGL config without MSAA so that map renders.

PR #1077, opened 2022-01-25
Fix map not rendering on emulators when MSAA is enabled

kiryldz:
### Summary of changes

At least Google simulators seem not to support MSAA option when configuring EGL. This will result in map not being rendered at all. Now logic is introduced to try double down `EGL_SAMPLE` value in a loop until either supported MSAA config will be found (fixing an issue of map not being rendered even on physical device when setting huge MSAA value) or EGL will be configured without MSAA (fixing an issue with emulators).

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix map not rendering on emulators when MSAA is enabled.</changelog>`.

Fixes: #1076 

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
@@ -14,9 +14,9 @@ import javax.microedition.khronos.egl.EGLDisplay
 
 internal class EGLConfigChooser constructor(
   private val translucentSurface: Boolean,
-  private val antialiasingSampleCount: Int,
+  private var antialiasingSampleCount: Int,
 ) {
-  private val antialiasingEnabled = antialiasingSampleCount > DEFAULT_ANTIALIASING_SAMPLE_COUNT
+  private val antialiasingEnabled get() = antialiasingSampleCount > DEFAULT_ANTIALIASING_SAMPLE_COUNT
 
   // Get all configs at least RGB 565 with 16 depth and 8 stencil
   private val configAttributes: IntArray
@@ -57,9 +57,8 @@ internal class EGLConfigChooser constructor(
   private var eglChooserSuccess = true
 
   fun chooseConfig(egl: EGL10, display: EGLDisplay): EGLConfig? {
-    val configAttrs = configAttributes
     // Determine number of possible configurations
-    val numConfigs = getNumberOfConfigurations(egl, display, configAttrs)
+    val numConfigs = getNumberOfConfigurations(egl, display)
     if (!eglChooserSuccess) {
       return null
     }
@@ -71,7 +70,6 @@ internal class EGLConfigChooser constructor(
     val possibleConfigurations = getPossibleConfigurations(
       egl,
       display,
-      configAttrs,
       numConfigs
     )
     if (!eglChooserSuccess) {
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
```diff
@@ -88,28 +86,48 @@ internal class EGLConfigChooser constructor(
...
     val numConfigs = IntArray(1)
-    if (!egl.eglChooseConfig(display, configAttributes, null, 0, numConfigs)) {
-      Logger.e(
-        TAG,
-        String.format(
-          MAPBOX_LOCALE,
-          "eglChooseConfig returned error %d",
-          egl.eglGetError()
+    val initialSampleCount = antialiasingSampleCount
+    while (true) {
```
pengdev, line 93: I'm a bit nervous to see the while(true) 😅

kiryldz: pretty common practice when configuring EGL, e.g. see [here](https://github.com/brackeen/glfm/blob/0581711cc7a1173e5a63d2982d506ec97716b17f/src/glfm_platform_android.c#L633) 😄

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
```diff
@@ -88,28 +86,48 @@ internal class EGLConfigChooser constructor(
...
+        }
+      } else {
+        break
+      }
+    }
+    if (initialSampleCount != antialiasingSampleCount) {
+      if (antialiasingSampleCount == 1) {
+        Logger.w(TAG, "Found EGL configs only with MSAA disabled.")
+      } else {
+        Logger.w(TAG, "Found EGL configs with MSAA enabled, EGL_SAMPLES=$antialiasingSampleCount.")
```
yunikkk, line 121: Don't we want to print supported EGL configs?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
```diff
@@ -88,28 +86,48 @@ internal class EGLConfigChooser constructor(
...
+          )
         )
-      )
-      eglChooserSuccess = false
+        if (antialiasingSampleCount > 1) {
+          Logger.w(TAG, "Reducing sample count in 2 times for MSAA as EGL_SAMPLES=$antialiasingSampleCount is not supported")
+          antialiasingSampleCount /= 2
+        } else {
+          // we did all we could, breaking the loop and indicating EGL config was not found
+          Logger.e(TAG, "No suitable EGL configs were found.")
```
yunikkk, line 109: Don't we want to print supported EGL configs?

pengdev:
@kiryldz could we add some tests to this function? unit test would be useful to cover logic for `getNumberOfConfigurations`  and preferably also some instrumentation tests?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
yunikkk: Yeah, right. Ok, lets keep this short message then

yunikkk (approved):

ank27 (approved):

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
```diff
@@ -88,28 +86,50 @@ internal class EGLConfigChooser constructor(
...
+            MAPBOX_LOCALE,
+            "eglChooseConfig returned error %d",
+            egl.eglGetError()
+          )
         )
-      )
-      eglChooserSuccess = false
+        if (antialiasingSampleCount > 1) {
+          Logger.w(TAG, "Reducing sample count in 2 times for MSAA as EGL_SAMPLES=$antialiasingSampleCount is not supported")
+          antialiasingSampleCount /= 2
```
pengdev, line 107: what if the device's supported sample count is like 4, 8 etc, while the user defined 5? would this algorithm end up picking 0 in the end?

kiryldz: @pengdev 5 / 2 = 2, 2 / 2 = 1 => config with MSAA disabled will be chosen. Dividing any number (> 1) by 2 will end up with 1.

kiryldz: if user defined 5, while device supports 4 or 8 - it depends on the driver, from what I've seen it EGL will select sample count rounded up so in this case it will select 8.
We will write a warning in this case though, this logic is not changed in this PR

PR #1093, opened 2022-01-28
Fix memory leak in location component.

Chaoba:
### Summary of changes
This PR adds a new class `CurrentLocationEngineCallback` to hold `LocationProviderImpl` instance with a weak reference to avoid memory leak issue.
Invoke `removeCompassListener` when there is no `LocationConsumer` registed.

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone.
 
+## main
+## Bug fixes 🐞
+* Fix memory leak in location component. ([#1093](https://github.com/mapbox/mapbox-maps-android/pull/1093))
+
 # 10.4.0-beta.1
 
 ## Features ✨ and improvements 🏁
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
@@ -42,30 +42,8 @@ class DefaultLocationProvider @VisibleForTesting(otherwise = PRIVATE) internal c
   private lateinit var runnable: Runnable
   private var updateDelay = INIT_UPDATE_DELAY
 
-  private val locationEngineCallback = object : LocationEngineCallback<LocationEngineResult> {
-    /**
-     * Invoked when new data available.
-     *
-     * @param result updated data.
-     */
-    override fun onSuccess(result: LocationEngineResult) {
-      result.lastLocation?.let {
-        notifyLocationUpdates(it)
-      }
-    }
-
-    /**
-     * Invoked when engine exception occurs.
-     *
-     * @param exception
-     */
-    override fun onFailure(exception: Exception) {
-      Logger.e(
-        TAG,
-        "Failed to obtain location update: $exception"
-      )
-    }
-  }
+  private val locationEngineCallback: LocationEngineCallback<LocationEngineResult> =
+    CurrentLocationEngineCallback(this)
 
   @VisibleForTesting(otherwise = PRIVATE)
   internal val locationCompassListener =
@@ -185,5 +163,26 @@ class DefaultLocationProvider @VisibleForTesting(otherwise = PRIVATE) internal c
     private const val TAG = "MapboxLocationProvider"
     private const val INIT_UPDATE_DELAY = 100L
     private const val MAX_UPDATE_DELAY = 5000L
+
+    private class CurrentLocationEngineCallback(locationProviderImpl: LocationProviderImpl) :
+      LocationEngineCallback<LocationEngineResult> {
+      private val componentWeakReference: WeakReference<LocationProviderImpl> =
+        WeakReference(locationProviderImpl)
+
+      override fun onSuccess(result: LocationEngineResult) {
+        result.lastLocation?.let { location ->
+          componentWeakReference.get()?.let {
+            it.notifyLocationUpdates(location)
+          }
+        }
+      }
+
+      override fun onFailure(exception: Exception) {
+        Logger.e(
+          TAG,
+          "Failed to obtain location update: $exception"
+        )
+      }
+    }
   }
 }
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
@@ -159,30 +159,30 @@ class DefaultLocationProvider @VisibleForTesting(otherwise = PRIVATE) internal c
     }
   }
 
+  private class CurrentLocationEngineCallback(locationProvider: DefaultLocationProvider) :
+    LocationEngineCallback<LocationEngineResult> {
+    private val locationProviderWeakReference: WeakReference<DefaultLocationProvider> =
+      WeakReference(locationProvider)
+
+    override fun onSuccess(result: LocationEngineResult) {
+      result.lastLocation?.let { location ->
+        locationProviderWeakReference.get()?.let {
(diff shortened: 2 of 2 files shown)


plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -152,14 +137,36 @@ internal class LocationProviderImpl(context: Context, var settings: LocationComp
...
+      locationCompassEngine.removeCompassListener(this)
     }
   }
 
   private companion object {
     private const val TAG = "MapboxLocationProvider"
     private const val INIT_UPDATE_DELAY = 100L
     private const val MAX_UPDATE_DELAY = 5000L
+
+    private class CurrentLocationEngineCallback(locationProviderImpl: LocationProviderImpl) :
```
kiryldz, line 151: class declared inside companion object? is it intended?

Chaoba: Yes. Do you have any concerns about this？

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone.
 
+## main
+## Bug fixes 🐞
+* Fix memory leak in location component. ([#1093](https://github.com/mapbox/mapbox-maps-android/pull/1093))
```
yunikkk, line 7: @Chaoba could you explain how was this reproduced?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -39,6 +41,9 @@ internal class LocationProviderImpl(context: Context, var settings: LocationComp
   private lateinit var runnable: Runnable
   private var updateDelay = INIT_UPDATE_DELAY
 
+  private val locationEngineCallback: LocationEngineCallback<LocationEngineResult> =
```
yunikkk, line 44: Since this variable is member of the class - it seem to still hold an implicit reference to the enclosing object (LocationProviderImpl).

CHANGELOG.md
Chaoba: Open and close LocationComponentActivity several times could reproduce it.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
Chaoba: `LocationProviderImpl` holds the reference to `LocationEngineCallback`. The change in this pr is to avoid `LocationEngineCallback` holds the reference to `LocationProviderImpl`

yunikkk: Ah indeed, seems I've confused with inner classes somehow.

CHANGELOG.md
yunikkk: Yeah I mean - aren't we hiding the actual problem here? Why isn't unRegisterLocationConsumer called here? Shouldn't lifecycle call it always?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
kiryldz: what's the benefit of it? comparing to inner class or similar

alexander-kulikovskii: should we test this class? 
and maybe make it as internal class in this file, not in `private companion`?

CHANGELOG.md
Chaoba: Actually, we do call unRegisterLocationConsumer, but still have this issue.

yunikkk: Hmm, how is this possible at all?

Chaoba: It seems `locationEngine` still holds the reference even we call       `locationEngine.removeLocationUpdates(locationEngineCallback)
`

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
Chaoba: Moved to an internal class. The current test cases could cover this class as it's in the route when a new location is updated.

CHANGELOG.md
yunikkk: Does this mean Android implementation is broken? Seems very suspicious, wouldn't it cause same issues with leaks in every app using location engine?

Chaoba: This can only be reproduced with compass engine, guess locationEngine holds the reference of location plugin as well as sensorManager.

yunikkk: Maybe we don't unsubscribe from compass engine somehow? I'd say compass is also very common component so if memory leaks were reproducible it would leak in many apps out there

Chaoba: The root cause should be `locationEngine` holde the reference the callback during request location update even we remove it. And then location compass is kept and casue the leak.

Only request location update or have location compass can't reproduce this issue, must combile these two together.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationProviderImpl.kt
```diff
@@ -168,6 +147,27 @@ internal class LocationProviderImpl(
     }
   }
 
+  private class CurrentLocationEngineCallback(locationProviderImpl: LocationProviderImpl) :
+    LocationEngineCallback<LocationEngineResult> {
+    private val componentWeakReference: WeakReference<LocationProviderImpl> =
```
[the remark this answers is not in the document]
Chaoba: Good catch, will rename it.

kiryldz (approved):

alexander-kulikovskii (approved):

pengdev (approved):

CHANGELOG.md
yunikkk: I've digged it a bit and it seems found the root cause. Thing is - we use Activity context to create SensorManager instance at [CompassEngine](https://github.com/mapbox/mapbox-maps-android/blob/9d5a59022b14d1ef8a1b30630d4fd5e703db5763/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationCompassEngine.kt#L19). 

And callbacks, that we add / remove to the LocationEngine indeed are retained, even after remove is called. 

So the issue became visible with the CompassEngine addition, since it holds SensorManager, which holds Context, which holds (and is) Activity. 
It seems to me that we should use application context everywhere, so that in future such problems won't be possible at all. 

PS
Context for the issue is interesting as well. It seems that our implementation of LocationEngine in mapbox-android-core lacks certain logic cleaning the callback or otherwise such issues wouldn't be possible as well. 

The issue with LocationEngine keeping its listeners even after removeLocationUpdates has very long story, raised [here](https://github.com/mapbox/mapbox-events-android/issues/289), reworked [here](https://github.com/mapbox/mapbox-events-android/pull/290) and followed with the documented recommendation to avoid leaks using WeakReferences [here](https://github.com/mapbox/android-docs/issues/834) published [here](https://docs.mapbox.com/android/core/guides/#requesting-location-updates). 
I'll cut a ticket in the lib repo to apply the fix as well.

yunikkk: https://github.com/mapbox/mapbox-events-android/issues/562

Issue #1048, opened 2022-01-12
LocationPuck not showing


badoualy:
## Environment

- Android OS version: API 30 (v11)
- Devices affected:
- Maps SDK Version: 10.2.0

## Observed behavior and steps to reproduce

I'm using the SDK in Compose inside an `AndroidView`, and the LocatoinPuck is not showing most of the time (90%)
(see code below)

Note: the pulsing effect is showing properly

## Expected behavior

LocationPuck should always be visible

## Notes / preliminary analysis

Honnestly not sure what's happening, I tried checking the threading.
I tried waiting for the style to be loaded, same thing.

## Additional links and references

```kotlin
AndroidView(
    factory = { 
        MapView(it).apply {
            location.updateSettings {
                enabled = true
                locationPuck = buildLocationPuck()
                pulsingEnabled = true
            }
        
            getMapboxMap().setCamera(
                CameraOptions.Builder()
                    .center(Point.fromLngLat(7.902085, 49.028343))
                    .zoom(8.0)
                    .build()
            )
        }
    },
    modifier = Modifier.fillMaxSize()
)

fun MapView.buildLocationPuck(): LocationPuck2D {
    return LocationPuck2D(
        bearingImage = AppCompatResources.getDrawable(context, R.drawable.ic_location_puck),
        topImage = AppCompatResources.getDrawable(context, R.drawable.ic_location_puck),
        scaleExpression = literal(3.0).toJson()
    )
}
```

Chaoba:
@badoualy Are you able to receive the GPS signal? Try to go outside or close to the window.

badoualy:
@Chaoba Yes, the pulse animation is present, the GPS coordinates are received correctly, the onIndicatorLocationChangedListener is also called

ZiZasaurus:
@badoualy can you let us know if this behavior also happens on physical devices?

yunikkk:
The root cause of the issue is similar to #321. Fix is in progress, should be available in the next release.

PR #1114, opened 2022-02-02
Split callbacks originating on getStyle and loadStyle requests.

yunikkk:
### Summary of changes

Problem this PR addresses : location puck is not displayed after style load #321, #1048.

Background context  : at #300 the logic that clears `style loaded` callbacks added with `mapboxMap.getStyle` / `mapboxMap.loadStyle` was introduced. 

However, for the `getStyle` the behaviour doesn't feel right. `getStyle` name doesn't suggest that it depends on the specific `loadStyle` call, instead it would be natural to return ANY style whenever it is loaded.

Existing `getStyle` implementation also leads to the issues (mentioned above) with our own LocationComponent. If location component was triggered before style load was started (possible if `MapInitOptions.styleUri = null` or style fails to load), it needs style and waits for it to appear at `getStyle` (in couple places, e.g. [here](https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationComponentPluginImpl.kt#L157) and [here](https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt#L94). When new style is loaded afterwards - this callbacks of location component are being cleared so it's never initialized, if not triggered manually again after style loaded.

I've changed behaviour of `mapboxMap.getStyle` callbacks, now they are not cleared anymore after the `loadStyle` call, while callbacks `loadStyle` retain current behaviour.

### User impact (optional)
Aside from fixing bugs with location puck - when user calls `mapboxMap.getStyle` they will receive style from the first style that is loaded successfully in future. 

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Fix location component now shown in certain cases.</changelog>`.

Fixes: #321, #1048

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -1333,6 +1333,12 @@ class MapboxMap internal constructor(
   /**
    * Returns if the style has been fully loaded.
    */
+  @Deprecated(
+    "Use getStyle()?.isStyleLoaded instead.",
+    replaceWith = ReplaceWith(
+      "getStyle()?.isStyleLoaded",
+    )
+  )
   override fun isFullyLoaded(): Boolean {
     return style?.isStyleLoaded ?: false
   }
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -37,7 +37,7 @@ import java.util.*
  * @property style the map style.
  */
 class MapboxMap internal constructor(
-  internal val nativeMapWeakRef: WeakReference<MapInterface>,
+  private val nativeMapWeakRef: WeakReference<MapInterface>,
   private val nativeObserver: NativeObserver,
   pixelRatio: Float
 ) : MapTransformDelegate,
@@ -49,11 +49,15 @@ class MapboxMap internal constructor(
   MapCameraManagerDelegate,
   MapStyleStateDelegate {
 
-  internal lateinit var style: Style
+  internal var style: Style? = null
   internal var isStyleLoadInitiated = false
-  private val styleObserver = StyleObserver(this, nativeObserver, pixelRatio)
+  private val styleObserver = StyleObserver(
+    nativeMapWeakRef,
+    { style -> this.style = style },
+    nativeObserver,
+    pixelRatio
+  )
   internal var renderHandler: Handler? = null
-  internal var styleLoaded = false
 
   /**
    * Represents current camera state.
@@ -102,9 +106,9 @@ class MapboxMap internal constructor(
   ) {
     initializeStyleLoad(onStyleLoaded, onMapLoadErrorListener)
     if (styleUri.isEmpty()) {
-      nativeMapWeakRef.call { (this as StyleManagerInterface).styleJSON = EMPTY_STYLE_JSON }
+      nativeMapWeakRef.call { styleJSON = EMPTY_STYLE_JSON }
     } else {
-      nativeMapWeakRef.call { (this as StyleManagerInterface).styleURI = styleUri }
+      nativeMapWeakRef.call { styleURI = styleUri }
     }
   }
 
@@ -138,7 +142,7 @@ class MapboxMap internal constructor(
   ) {
     initializeStyleLoad(onStyleLoaded, onMapLoadErrorListener)
     nativeMapWeakRef.call {
-      (this as StyleManagerInterface).styleJSON = styleJson
+      styleJSON = styleJson
     }
   }
 
@@ -214,16 +218,12 @@ class MapboxMap internal constructor(
     onStyleLoaded: Style.OnStyleLoaded? = null,
     onMapLoadErrorListener: OnMapLoadErrorListener? = null
   ) {
-    // clear listeners from previous invocation
-    styleObserver.onNewStyleLoad(
-      {
-        styleLoaded = true
-        onStyleLoaded?.onStyleLoaded(it)
-      },
+    style = null
+    styleObserver.setLoadStyleListener(
+      onStyleLoaded,
       onMapLoadErrorListener
     )
     isStyleLoadInitiated = true
-    styleLoaded = false
   }
 
   /**
@@ -232,29 +232,15 @@ class MapboxMap internal constructor(
    * @param onStyleLoaded the callback to be invoked when the style is fully loaded
    */
   fun getStyle(onStyleLoaded: Style.OnStyleLoaded) {
-    if (::style.isInitialized) {
-      if (styleLoaded) {
-        // style has loaded, notify callback immediately
-        onStyleLoaded.onStyleLoaded(style)
-      } else {
-        // style load is occurring now, add callback
-        styleObserver.addOnStyleLoadListener(onStyleLoaded)
-      }
-    } else {
-      // no style has loaded yet, add callback
-      styleObserver.addOnStyleLoadListener(onStyleLoaded)
-    }
+    style?.let(onStyleLoaded::onStyleLoaded)
+      ?: styleObserver.addGetStyleListener(onStyleLoaded)
   }
 
   /**
    * Get the Style of the map synchronously, will return null is style is not loaded yet.
    */
   fun getStyle(): Style? {
-    if (::style.isInitialized && styleLoaded) {
-      // style has loaded, return it immediately
-      return style
-    }
-    return null
+    return style
   }
 
   /**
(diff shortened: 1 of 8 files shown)


yunikkk:
Tests coming

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -49,11 +49,15 @@ class MapboxMap internal constructor(
   MapCameraManagerDelegate,
   MapStyleStateDelegate {
 
-  internal lateinit var style: Style
+  internal var style: Style? = null
   internal var isStyleLoadInitiated = false
-  private val styleObserver = StyleObserver(this, nativeObserver, pixelRatio)
+  private val styleObserver = StyleObserver(
```
yunikkk, line 54: Removed `this` param from StyleObserver, using reference to the object not yet constructed fully could lead to subtle bugs. Instead only passed lambda that inits style directly.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1348,7 +1334,7 @@ class MapboxMap internal constructor(
    * Returns if the style has been fully loaded.
    */
   override fun isFullyLoaded(): Boolean {
-    return style.isStyleLoaded
+    return style?.isStyleLoaded ?: false
```
yunikkk, line 1337: Not sure this method is needed at all, should we deprecate it? Could users just call `style.isStyleLoaded` themselves?

kiryldz:
@yunikkk label and changelog entry are missing, can you add those so CI will pass?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -102,9 +106,9 @@ class MapboxMap internal constructor(
   ) {
     initializeStyleLoad(onStyleLoaded, onMapLoadErrorListener)
     if (styleUri.isEmpty()) {
-      nativeMapWeakRef.call { (this as StyleManagerInterface).styleJSON = EMPTY_STYLE_JSON }
+      nativeMapWeakRef.call { styleJSON = EMPTY_STYLE_JSON }
```
pengdev, line 109: Recommend to use `this` keyword to limit the scope of the function to be within nativeMap, so that not to be confused with  property/functions with the same name in the `MapboxMap` class. It used to cause trouble since we have the exact same functions both inside the `NativeMap` and `MapboxMap`.
```suggestion
      nativeMapWeakRef.call { this.styleJSON = EMPTY_STYLE_JSON }
```

pengdev:
@yunikkk could you verify if the PR fixes https://github.com/mapbox/mapbox-maps-android/issues/321, https://github.com/mapbox/mapbox-maps-android/issues/1048  with a snapshot build?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: Could you explain a bit more? `this.` is implied anyway, 
`nativeMapWeakRef.call { someProperty ->`
should be always the same as 
`nativeMapWeakRef.call { this.someProperty`

yunikkk:
> 

Yes, will check

yunikkk:
> @yunikkk could you verify if the PR fixes #321, #1048 with a snapshot build?

Yes, will check

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
tobrun: this is a remains from when we were using style state of gl-native. That has a slightly different setup which resulted in managing our own state in Style. We need to keep this around for compatibility for now..

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
tobrun: we have had some occurrences where it was calling the function on the outer class versus on the native map. Not much of an issue here though

tobrun (approved):

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: Ah, got it, it seems. 
IMO for such situations when `this.` context might be confused and outer class may hold similar or same method - better to use explicit named references with let e.g. `let { inner -> inner.smth }` instead of apply or similar that changes `this.`

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: We could deprecate it, I suppose, wdyt?

Chaoba (approved):

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -13,51 +13,60 @@ import java.util.concurrent.CopyOnWriteArrayList
...
-  private val mapboxMap: MapboxMap,
+  private val nativeMapWeakRef: WeakReference<MapInterface>,
+  private val styleLoadedListener: Style.OnStyleLoaded,
   private val nativeObserver: NativeObserver,
   private val pixelRatio: Float
 ) : OnStyleLoadedListener, OnMapLoadErrorListener {
 
-  private val awaitingStyleLoadListeners = CopyOnWriteArrayList<Style.OnStyleLoaded>()
-  private var awaitingStyleErrorListener: OnMapLoadErrorListener? = null
+  @Volatile
```
kiryldz, line 22: our maps sdk is single-threaded, why volatile here and below?

yunikkk: Hm, is onStyleLoaded called from the same thread? 
Why was `CopyOnWriteArrayList` used previously?

yunikkk:
> @yunikkk could you verify if the PR fixes #321, #1048 with a snapshot build?

I've verified that 1tap originated [issue](https://github.com/mapbox/mapbox-maps-android/issues/321) is resolved, and the [compose](https://github.com/mapbox/mapbox-maps-android/issues/1048) actually is not (though it's reproduced only on the emulator).

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -232,29 +232,15 @@ class MapboxMap internal constructor(
...
-        onStyleLoaded.onStyleLoaded(style)
-      } else {
-        // style load is occurring now, add callback
-        styleObserver.addOnStyleLoadListener(onStyleLoaded)
-      }
-    } else {
-      // no style has loaded yet, add callback
-      styleObserver.addOnStyleLoadListener(onStyleLoaded)
-    }
+    style?.let(onStyleLoaded::onStyleLoaded)
```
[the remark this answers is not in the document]
yunikkk: You suggest 
```
    val style = this.style
    if (style != null) {
      onStyleLoaded.onStyleLoaded(style)
    } else {
      styleObserver.addGetStyleListener(onStyleLoaded)
    }
```
?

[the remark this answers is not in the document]
kiryldz: yes, just guess no need for explicit `val style = this.style`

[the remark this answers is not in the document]
yunikkk: > yes, just guess no need for explicit `val style = this.style`

Without it compiler won't allow to do a safe cast, either `style!!` or such separate val is needed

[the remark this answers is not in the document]
yunikkk: Discussed offline, keeping as is

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
kiryldz: IDK
however even if callbacks arrive from different threads (although I doubt it) - running that code from this collection will most likely lead to a problem

yunikkk: Removed, callback arrives on the main thread

kiryldz (approved):

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: Deprecated

PR #1123, opened 2022-02-07
Add plugins version

ank27:
### Summary of changes
- Added plugins version v1.1.3 (https://github.com/mapbox/android-sdk-versions-plugin)

## Pull request checklist:
    
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Added sdk versions plugin v1.1.3.</changelog>`.
 - [x] If this PR is a `v10.3` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.3` release branch.


yunikkk (approved):

yunikkk:
Btw ktlint check fails on CI

Chaoba (approved):

ank27:
> Btw ktlint check fails on CI

Yeah, that's benchmark-check. it will be fixed by internal ticket.

kiryldz:
@ank27 shouldn't we actually add generated asset files as part of this PR?

kiryldz:
> @ank27 shouldn't we actually add generated asset files as part of this PR?

As discussed offline seems answer is no. 
However we need to think of some concrete steps how / when to populate correctly `VERSION_NAME` inside `gradle.properties` so that version obtainable in runtime will be the correct one when we perform the release of our SDK.

I believe correct solution for now will be updating our release template with a step of not forgetting to specify correct version in `gradle.properties` when we're cutting a release but perhaps there's a better way cc @mapbox/maps-android

tobrun (approved):

sdk/build.gradle.kts
```diff
@@ -6,6 +6,7 @@ plugins {
   kotlin("android")
   id("com.jaredsburrows.license")
   id("org.jetbrains.dokka")
+  id("com.mapbox.android.sdk.versions")
```
pengdev, line 9: currently we only apply the version plugin to the sdk module, however we could discuss if we want to add it to all our plugins, as mentioned in https://github.com/mapbox/android-sdk-versions-plugin#about-android-sdk-versions-plugin
> All Mapbox android SDK modules that incorporate the Telemetry SDK and meant to be released as libraries(artifacts) should apply this plugin to facilitate the Telemetry SDK access all Mapbox library versions at run time.

pengdev (approved):
LGTM, could we test it out and validate the asset file is generated correctly?

sdk/build.gradle.kts
ank27: As discussed offline with @nagineni 
> we should be adding the dependency only to the core module.

We can discuss more if we want to add it to plugins as well, but we will iterate this in future releases. currently, core doesn't have a parsing logic in place and they are working on another solution as well.

ank27:
> LGTM, could we test it out and validate the asset file is generated correctly?

Yes, Asset file is generated correctly. 
@tarigo @nagineni can you confirm this is correct name that core wants?

asset_file

tarigo:
> > LGTM, could we test it out and validate the asset file is generated correctly?
> 
> Yes, Asset file is generated correctly. @tarigo @nagineni can you confirm this is correct name that core wants?
> 
> asset_file

No, unfortunately this is not correct. But it's a good point to consider in the new plugin: project.name is not enough, probably we have to use `ext.mapboxRegistrySdkName`  or POM_GROUP_ID+POM_ARTIFACT_ID to build SDK name.

pengdev:
> project.name is not enough, probably we have to use ext.mapboxRegistrySdkName or POM_GROUP_ID+POM_ARTIFACT_ID to build SDK name.

These information are all available in the sdk-registry plugin, probably we can combine the versions plugin to the sdk-registry plugin by adding a separate task?

tarigo:
> > project.name is not enough, probably we have to use ext.mapboxRegistrySdkName or POM_GROUP_ID+POM_ARTIFACT_ID to build SDK name.
> 
> These information are all available in the sdk-registry plugin, probably we can combine the versions plugin to the sdk-registry plugin by adding a separate task?

Good idea!

PR #1151, opened 2022-02-11
View annotations: refactor public surface to be regular class

kiryldz:
### Summary of changes

Change `ViewAnnotationManager` to be an open regular class instead of an interface to allow adding methods to public API not leading to a breaking change. 
In any case `ViewAnnotationManager` is not designed for public inheritance and we state it now in a more clear way.  We hold public `ViewAnnotationManager` in the correct `com.mapbox.maps.viewannotation` package while internal `ViewAnnotationManagerImpl` is located in `com.mapbox.maps` to access some package-private classes.

In the future with some changes landing in v10.5 downstream impl class will be moved to `com.mapbox.maps.viewannotation` and renamed to `ViewAnnotationManager`.

Adding skip changelog label as this should not affect users anyhow.

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -16,7 +16,7 @@ import kotlin.collections.HashMap
 
 internal class ViewAnnotationManagerImpl(
   private val mapView: MapView
-) : ViewAnnotationManager, ViewAnnotationPositionsUpdateListener {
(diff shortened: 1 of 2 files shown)


kiryldz:
> I'm not really sure that I see the problem this change solves. Does user has any reasons to implement this interface in the first place? It seems it's not intended at all. Then we introduce ugly `open fun` that throws exceptions (that is not safe either, if we imagine that user might create / override this class themselves)...

I tend to agree, as discussed offline, I'll try to go with another way even now before waiting for v10.5 and fix package name in gl-native.
Closing for now.

PR #1161, opened 2022-02-15
Ignore generated BuildConfig from tracked APIs.

pengdev:
### Summary of changes

This PR ignore generated BuildConfig from tracked APIs and fixed the style extension's package name in the manifest.

## Pull request checklist:
    
### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -79,7 +79,23 @@ apiValidation {
    * Classes (fully qualified) that are excluded from public API dumps even if they
    * contain public API.
    */
-//  ignoredClasses.add("com.company.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.androidauto.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.style.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.localization.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.module.telemetry.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.camera.animation.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.annotation.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.attribution.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.compass.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.gestures.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.lifecycle.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.locationcomponent.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.logo.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.overlay.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.scalebar.BuildConfig")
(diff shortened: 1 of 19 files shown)


build.gradle.kts
```diff
@@ -79,7 +79,23 @@ apiValidation {
    * Classes (fully qualified) that are excluded from public API dumps even if they
    * contain public API.
    */
-//  ignoredClasses.add("com.company.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.androidauto.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.style.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.localization.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.module.telemetry.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.plugin.camera.animation.BuildConfig")
```
pengdev, line 86: noticed the package name in the plugin-animation's AndroidManifest file does not match the package of the source code `com.mapbox.maps.plugin.animation`, but I assume it should be ok.

extension-style/src/main/AndroidManifest.xml
```diff
@@ -1 +1 @@
-<manifest package="com.mapbox.maps.plugin.style" />
+<manifest package="com.mapbox.maps.extension.style" />
```
pengdev, line 1: Fixed the package name for the style extension, this shouldn't introduce breaking change, as described in  https://developer.android.com/guide/topics/manifest/manifest-element#package

> While building your app into the an application package (APK), the build system uses the package attribute for two things:
> 
> * It applies this name as the namespace for your app's generated R.java class (used to access your [app resources](https://developer.android.com/guide/topics/resources/overview)).
> For example, if package is set to "com.example.myapp", the R class is created at com.example.myapp.R.
> 
> * It uses this name to resolve any relative class names that are declared in the manifest file.
> For example, if package is set to "com.example.myapp", an activity declared as <activity android:name=".MainActivity"> is resolved to be com.example.myapp.MainActivity.

And there's no resources defined for the style extension or relative classes declared in the manifest file.

build.gradle.kts
Chaoba: We should fix this in v11

build.gradle.kts
```diff
@@ -79,7 +79,23 @@ apiValidation {
    * Classes (fully qualified) that are excluded from public API dumps even if they
    * contain public API.
    */
-//  ignoredClasses.add("com.company.BuildConfig")
+  ignoredClasses.add("com.mapbox.maps.extension.androidauto.BuildConfig")
```
Chaoba, line 82: How about use `ignoredClasses.addAll`?

ank27: yes, we can use `addAll` method here.

build.gradle.kts
```diff
@@ -79,7 +79,27 @@ apiValidation {
    * Classes (fully qualified) that are excluded from public API dumps even if they
    * contain public API.
    */
-//  ignoredClasses.add("com.company.BuildConfig")
+  ignoredClasses.addAll(
```
alexander-kulikovskii, line 82: just a question
can use smth like `com.mapbox.maps.*.BuildConfig`

pengdev: sadly no, it only takes fully qualified class names, e.g. the full path of the class without wildcards.

Chaoba (approved):

build.gradle.kts
pengdev: Yes, would be good to have an internal doc on how to create new plugins/modules to the system, such a guide/readme does not exist yet, will ticket it out to follow up.

PR #1174, opened 2022-02-22
Add map function to apply style transition during style load

kiryldz:
### Summary of changes

Introduce new function `MapboxMap.setStyleTransition` and deprecate `Style.setStyleTransition`. 

Main idea is that we need to apply transition options in the correct moment of time _during_ style is loading while our current API allows to actually operate on `Style` object when style is already fully loaded.

Please take a look at the visuals, what we do there is:
 - load first streets style with following options `TransitionOptions.Builder().duration(0).build()`
 - wait for 2 seconds and load dark style with `TransitionOptions.Builder().duration(5000).delay(3000).build()`

### User impact (optional)

Following pseudo-code:
```
loadStyleUri(Style.MAPBOX_STREETS) { firstStyle ->
  firstStyle.styleTransition = TransitionOptions.Builder()
     .duration(0)
     .build()
}
```
should be replaced with:
```

loadStyleUri(
  styleUri = Style.MAPBOX_STREETS,
  styleTransitionOptions = TransitionOptions.Builder()
     .duration(0)
     .build()
)
```

## Pull request checklist:

 | Before | After |
 | - | - |
 | <video src="https://user-images.githubusercontent.com/15800566/155156030-3177fcde-8d0f-4473-9c21-aaf91a801bd1.mp4" width = 250/> | <video src="https://user-images.githubusercontent.com/15800566/155156272-675603b1-17dd-44d6-8155-0952dfd91b16.mp4" width = 250/> |
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog>Add overloaded MapboxMap.loadStyleUri, MapboxMap.loadStyleJson, MapboxMap.loadStyle with `TransitionOptions` parameter allowing to apply transition style being loaded.</changelog>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
-* Add overloaded `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` with `TransitionOptions` parameter allowing to apply transition style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
+* Add optional `TransitionOptions` parameter to `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` to apply transition to style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
 
 ## Bug fixes 🐞
 * Fix skipping / crashing user events scheduled on a render thread with `MapView#queueEvent`. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
-* Add `MapboxMap.setStyleTransition` function and deprecate `Style.setStyleTransition` fixing style transition options not being applied. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
+* Add `MapboxMap.setStyleTransition` / `MapboxMap.getStyleTransition` function and deprecate `Style.setStyleTransition` / `Style.getStyleTransition` fixing style transition options not being applied. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
 
 ## Bug fixes 🐞
 * Fix skipping / crashing user events scheduled on a render thread with `MapView#queueEvent`. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
+* Add `MapboxMap.setStyleTransition` function and deprecate `Style.setStyleTransition` fixing style transition options not being applied. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
 
 ## Bug fixes 🐞
 * Fix skipping / crashing user events scheduled on a render thread with `MapView#queueEvent`. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
-* Add `MapboxMap.setStyleTransition` / `MapboxMap.getStyleTransition` function and deprecate `Style.setStyleTransition` / `Style.getStyleTransition` fixing style transition options not being applied. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
+* Add overloaded `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` with `TransitionOptions` parameter allowing to apply transition style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
 
 ## Bug fixes 🐞
 * Fix skipping / crashing user events scheduled on a render thread with `MapView#queueEvent`. ([#1068](https://github.com/mapbox/mapbox-maps-android/pull/1068))
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
@@ -35,35 +35,34 @@ class CircleLayerClusteringActivity : AppCompatActivity() {
 
     // Disable any type of fading transition when icons collide on the map. This enhances the visual
     // look of the data clustering together and breaking apart.
-    mapboxMap.setStyleTransition(
-      TransitionOptions.Builder()
+    mapboxMap.loadStyleUri(
+      styleUri = Style.LIGHT,
+      styleTransitionOptions = TransitionOptions.Builder()
         .duration(0)
         .delay(0)
         .enablePlacementTransitions(false)
-        .build()
-    )
-    mapboxMap.loadStyleUri(
-      Style.LIGHT
-    ) {
-      mapboxMap.flyTo(
-        CameraOptions.Builder()
-          .center(Point.fromLngLat(-79.045, 12.099))
-          .zoom(3.0)
-          .build()
-      )
+        .build(),
+      onStyleLoaded = {
+        mapboxMap.flyTo(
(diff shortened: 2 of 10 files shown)


ank27:
Since we apply `styleTransition` to map now, does the same transition will be applied when switching between styles?

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -76,12 +83,23 @@ internal class StyleObserver(
     loadStyleErrorListener?.onMapLoadError(eventData)
   }
 
+  override fun onStyleDataLoaded(eventData: StyleDataLoadedEventData) {
+    // transition options must be applied after style but before sprite and sources to take effect
```
pengdev, line 87: any tip on how it works? does it mean the transitions set after the style is fully loaded does not work any more?

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -91,6 +91,10 @@ class Style internal constructor(
    *
    * @param transitionOptions Map style transition options.
    */
+  @Deprecated(
+    "Transition options will not be applied properly, please use another function.",
+    replaceWith = ReplaceWith("mapboxMap.setStyleTransition")
+  )
   override fun setStyleTransition(transitionOptions: TransitionOptions) =
```
yunikkk, line 98: What about getStyleTransition though? Should it be moved / used via MapboxMap as well?

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -91,6 +91,10 @@ class Style internal constructor(
    *
    * @param transitionOptions Map style transition options.
    */
+  @Deprecated(
+    "Transition options will not be applied properly, please use another function.",
+    replaceWith = ReplaceWith("mapboxMap.setStyleTransition")
```
yunikkk, line 96: Does it replaces an expression correctly though?

yunikkk: `mapboxMap.setStyleTransition(transitionOptions)` should work better

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
kiryldz: how it works now? it does not work now, that's the point of the fix 😄 
technically it could work even now e.g. we do set duration to 0 now (default duration = 300 ms) - style gets loaded pretty fast, let's say in 50 ms - and after it's loaded setting duration to 0 while it's already loading as part of 300 ms - this _seems_ to work now, from what I see.

but in general it's not working and that's why I did attach visuals - if we set duration > 300 ms - it will not take effect before the fix. does it make sense @pengdev?

sdk/src/main/java/com/mapbox/maps/Style.kt
kiryldz: it's a bit tricky to implement it...
what should be returned if user did call `MapboxMap.setStyleTransition` but did not yet load a style? on the one hand we could return cached `styleTransitionOptions` but on the other hand those options are not yet applied in core. that's why I decided to leave getter for actually loaded `Style` object.

sdk/src/main/java/com/mapbox/maps/Style.kt
kiryldz: yeah, my miss, thanks for catching, will take another look so that it gets replaced correctly

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -35,9 +40,11 @@ internal class StyleObserver(
    * NOTE : listener is invoked only once after successful style load.
    */
   fun setLoadStyleListener(
+    transitionOptions: TransitionOptions?,
```
Chaoba, line 43: Maybe create another `setTransitionOptions` function for this? It's a little wired to set `TransitionOptions` in this function.

kiryldz: why? setting transition options is tightly connected with loading a style and setting listeners imho

sdk/src/main/java/com/mapbox/maps/Style.kt
pengdev: Imo, the get/set API should appear in pairs, we can keep the `getStyleTransition` in Style object, but maybe add a `getStyleTransition` to the `MapboxMap` object and return the cached one? (similar to what `MapboxMap.setStyleTransition` does, which is set the transition to the cache which will be applied later)

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
pengdev: I mean does the `setStyleTransition` change the default transition for all the layer properties when the property is changed after the style is fully loaded?
I think the use case you suggested here is that the transition will be applied while loading a new style for the first time.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -191,6 +214,16 @@ class MapboxMap internal constructor(
...
 
+  /**
+   * Set style transition options that will be applied to all style loads happening after.
+   *
+   * In order to obtain current transition options please use [Style.getStyleTransition]
+   * when style will be loaded.
+   */
+  fun setStyleTransition(transitionOptions: TransitionOptions) {
+    styleTransitionOptions = transitionOptions
+  }
```
pengdev, line 225: what if the style is loaded and then the `setStyleTransition` is applied? The setStyleTransition will never go through to gl-native until the next style is loaded..

Continuing my previous question in https://github.com/mapbox/mapbox-maps-android/pull/1174/files#r812699285

> I mean does the setStyleTransition change the default transition for all the layer properties when the property is changed after the style is fully loaded?

If setting the `StyleTransition` changes the default transition for all the transitional style properties, we should apply the `setStyleTransition` to gl-native immediately if the Style is loaded.

Otherwise this function will introduce bug when `MapboxMap.setStyleTransition` is called after the style is loaded, and some of the transitionable style properties is changed, while the default transition is not applied.

kiryldz: thanks, @pengdev, good point

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1488,6 +1577,13 @@ class MapboxMap internal constructor(
       Map.clearData(resourceOptions, callback)
     }
 
+    @VisibleForTesting(otherwise = PRIVATE)
+    internal val DEFAULT_STYLE_TRANSITION = TransitionOptions.Builder()
```
kiryldz, line 1581: I'll create an issue in gl-native later so that this default object could be returned from gl-native _before_ we actually load a style (for now it returns empty object).

kiryldz:
> > Since we apply `styleTransition` to map now, does the same transition will be applied when switching between styles?
> 
> Previously it was not applied at all but yes, I did even state in docs that user could set it once - and it will be applied to all style loads happening afterwards.

@ank27 correction after deeper look - we do actually reset transition options for the next style, it's reflected in code now, PTAL

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -191,6 +218,67 @@ class MapboxMap internal constructor(
...
+   *
+   * Style transition options will be applied to current style immediately if it is loaded and
+   * to the upcoming style as well if [forUpcomingStyle] is True.
+   * If no style is loaded [transitionOptions] will be applied to an upcoming style not taking
+   * [forUpcomingStyle] flag in consideration.
+   */
+  @JvmOverloads
+  fun setStyleTransition(
+    transitionOptions: TransitionOptions,
+    forUpcomingStyle: Boolean = true
```
pengdev, line 234: shall we keep the `setStyleTransition` for both `MapboxMap` and `Style`? Then we can remove the flag, and only update the cached transition for `MapboxMap`. The reason is the settings in the `MapboxMap` should be independent of the `Style`, adding this flag could be confusing.

pengdev: If I'm the user, I expect the `setStyleTransition` in MapboxMap object to take effect immediately(pass through to gl-native) and also be cached for future transitions. Also I would suggest rename this method to 
```
fun setTransition(transitionOptions: TransitionOptions)
```

kiryldz: refactored again

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
kiryldz: not relevant anymore

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
+* Add overloaded `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` with `TransitionOptions` parameter allowing to apply transition style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
```
yunikkk, line 36:
```suggestion
* Add `TransitionOptions` parameter to `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` to apply transition to style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
```

app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
```diff
@@ -33,36 +33,36 @@ class CircleLayerClusteringActivity : AppCompatActivity() {
     setContentView(mapView)
     val mapboxMap = mapView.getMapboxMap()
 
+    // Disable any type of fading transition when icons collide on the map. This enhances the visual
+    // look of the data clustering together and breaking apart.
```
yunikkk, line 37: nit : move comment above the styleTransitionOptions

yunikkk (approved):
1 nit

ank27 (approved):

app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
```diff
@@ -34,35 +34,35 @@ class CircleLayerClusteringActivity : AppCompatActivity() {
...
       // Disable any type of fading transition when icons collide on the map. This enhances the visual
       // look of the data clustering together and breaking apart.
-      it.styleTransition = TransitionOptions.Builder()
+      styleTransitionOptions = TransitionOptions.Builder()
         .duration(0)
         .delay(0)
         .enablePlacementTransitions(false)
-        .build()
+        .build(),
+      onStyleLoaded = {
```
pengdev, line 45: nit: can it be an ending block without the `onStyleLoaded =`?

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ mapView.location2.puckBearingEnabled = false
 * Enable tile packs for DEM terrain tiles, it includes both Offline API and `TileStoreUsageMode::ReadAndUpdate` resource option. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Render tiles with partial content while the glyph dependencies are loading. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 * Canonicalize URLs and enable Offline API usage for the 3dtiles/v1 tiles. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
+* Add `TransitionOptions` parameter to `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` to apply transition to style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
```
pengdev, line 36:
```suggestion
* Add optional `TransitionOptions` parameter to `MapboxMap.loadStyleUri`, `MapboxMap.loadStyleJson`, `MapboxMap.loadStyle` to apply transition to style being loaded. ([#1174](https://github.com/mapbox/mapbox-maps-android/pull/1174))
```

pengdev (approved):
LGTM % nits

app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
kiryldz: sadly it can't as in that case lambda `{}` is ambiguous as we have 2 callbacks so that's the only way.

PR #1175, opened 2022-02-22
[Deps] Bump gl native 10.4.0 beta.2

ank27:
### Summary of changes
Bump gl-native to 10.4.0-beta.2
Revert common.so changes.

## Pull request checklist:
    
 - [ ] Run `./gradlew apiDump` to update generated api files, if there's public API changes, otherwise the `verify-kotlin-binary-compatibility` CI will fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,6 @@ commands:
                 --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
                 --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
                 --device model=harpia,version=23,locale=en,orientation=portrait \
-                --device model=AOP_sprout,version=28,locale=en,orientation=portrait \
                 --timeout 45m --use-orchestrator --no-record-video --num-flaky-test-attempts 3 \
                 --num-uniform-shards 4
               fi
@@ -310,7 +309,6 @@ commands:
             if [[ -n "${GCLOUD_SERVICE_ACCOUNT_JSON}" ]]; then
               gcloud firebase test android run --type robo \
                 --app app/build/outputs/apk/release/app-release.apk \
-                --device model=AOP_sprout,version=28,locale=en,orientation=portrait \
                 --device model=harpia,version=23,locale=en,orientation=portrait \
                 --device model=Pixel2,version=30,locale=en,orientation=portrait \
                 --device model=G8142,version=25,locale=en,orientation=portrait \
--- a/android-auto-app/build.gradle.kts
+++ b/android-auto-app/build.gradle.kts
@@ -35,7 +35,6 @@ android {
     if (buildFromSource.toBoolean()) {
       jniLibs.pickFirsts.add("**/libc++_shared.so")
     }
-    jniLibs.pickFirsts.add("**/libmapbox-common.so")
   }
 }
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -58,7 +58,6 @@ android {
       if (buildFromSource.toBoolean()) {
         jniLibs.pickFirsts.add("**/libc++_shared.so")
       }
-      jniLibs.pickFirsts.add("**/libmapbox-common.so")
     }
   }
 
--- a/extension-style-app/build.gradle.kts
+++ b/extension-style-app/build.gradle.kts
@@ -35,7 +35,6 @@ android {
     if (buildFromSource.toBoolean()) {
       jniLibs.pickFirsts.add("**/libc++_shared.so")
     }
-    jniLibs.pickFirsts.add("**/libmapbox-common.so")
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ mapView.location2.puckBearingEnabled = false
 * Fix shimmering artifact when pitched raster tiles with compressed textures are rendered. ([#1160](https://github.com/mapbox/mapbox-maps-android/pull/1160))
 
 ## Dependencies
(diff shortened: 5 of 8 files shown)


app/build.gradle.kts
```diff
@@ -58,7 +58,6 @@ android {
       if (buildFromSource.toBoolean()) {
         jniLibs.pickFirsts.add("**/libc++_shared.so")
       }
-      jniLibs.pickFirsts.add("**/libmapbox-common.so")
```
Chaoba, line 61: We still need it when build from source

extension-style-app/build.gradle.kts
```diff
@@ -35,7 +35,6 @@ android {
     if (buildFromSource.toBoolean()) {
       jniLibs.pickFirsts.add("**/libc++_shared.so")
     }
-    jniLibs.pickFirsts.add("**/libmapbox-common.so")
```
Chaoba, line 38: The same here

android-auto-app/build.gradle.kts
```diff
@@ -35,7 +35,6 @@ android {
     if (buildFromSource.toBoolean()) {
       jniLibs.pickFirsts.add("**/libc++_shared.so")
     }
-    jniLibs.pickFirsts.add("**/libmapbox-common.so")
```
Chaoba, line 38: The same here

kiryldz: @Chaoba with previous gradle plugin it was more of an AGP bug that `so` from previous build was not cleared properly leading to duplicate error. Cleaning `build` directories fixed that. If that issue is fixed with new gradle plugin - we indeed do not need it here I guess.
@ank27 any more comments here?

android-auto-app/build.gradle.kts
```diff
@@ -35,7 +35,6 @@ android {
     if (buildFromSource.toBoolean()) {
       jniLibs.pickFirsts.add("**/libc++_shared.so")
```
kiryldz, line 36: @ank27 do we need that by the way? I was under an impression that gl-native does not contain libc++_shared and only common does and in that case we don't need pick first

android-auto-app/build.gradle.kts
ank27: Yes correct, and since gl-native exclude `common.so`  as well, we don't need this here.

android-auto-app/build.gradle.kts
ank27: Technically we don't need this as well. But I guess we can keep it the same as before. 
Also, I don't know if any other project builds `libc++_shared.so` file, why it was added in the first place?

pengdev: This is used for building from source, where I think gl-native indeed will bundle the `libc++_shared.so` when built from source, we could verify it by building the project from source.

android-auto-app/build.gradle.kts
pengdev: the common.so could be duplicated when building from source, and it was 

```
if (buildFromSource.toBoolean()) {
    packagingOptions {
      pickFirst("**/libc++_shared.so")
      pickFirst("**/libmapbox-common.so")
    }
  }
}
```

before the [v10.4.0-beta.1 bump](https://github.com/mapbox/mapbox-maps-android/pull/1160/files#diff-8cff73265af19c059547b76aca8882cbaa3209291406f52df1dafbbc78e80c46L59-L60), we should probably move `jniLibs.pickFirsts.add("**/libmapbox-common.so")` to the `if (buildFromSource.toBoolean())` closure.

kiryldz: @pengdev as I've described [here](https://github.com/mapbox/mapbox-maps-android/pull/1175#discussion_r812630460) those conditions were related to an issue in older AGP and could be fixed locally by cleaning previous build files. IMO we should remove it from end application gradle files to see if issue is resolved now and even if it's not - AFAIR we have a step in build from source section noting to call `make clean` before building from source. If we leave it - we may simply hide some error.

ank27: Earlier releases of gl-native did not exclude `common.so` explicitly from `packagingOptions`, i guess that's why pickFirst was added in the above config. If i understand correctly even building from source, the libs files are packaged as per `packaginOptions` in gradle. since now we exclude `common.so` from package with 10.4.0-beta.2, it shouldn't matter if we `pickFirst` in the config above. But we can keep it just be sure, wdyt? 🤔

yunikkk (approved):

android-auto-app/build.gradle.kts
pengdev: we can remove it and if we encounter issues while building from source, we add it back 🙂

Chaoba (approved):

kiryldz (approved):

android-auto-app/build.gradle.kts
ank27: > we can remove it and if we encounter issues while building from source, we add it back 🙂 Better to verify build from source first.

Attached internal ticket points to this pr which has `buildFromSource` passed. I think that would be fine to verify this behaviour?

pengdev (approved):

.circleci/config.yml
```diff
@@ -287,15 +287,15 @@ commands:
                 gcloud beta firebase test android run --type instrumentation \
                   --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
                   --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
-                  --device model=AOP_sprout,version=28,locale=en,orientation=portrait \
+#                  --device model=AOP_sprout,version=28,locale=en,orientation=portrait \
```
yunikkk, line 290: Submit a ticket to track?

ank27: done

PR #1200, opened 2022-03-03
Add extension function coordinatesForPixels

Chaoba:
### Summary of changes
This PR adds a new extension function `coordinatesForPixel` with rectangle parameter 

## Pull request checklist:
    
 - [ ] Run `make update-api` to update generated api files, if there's public API changes, otherwise the `verify-kotlin-binary-compatibility` CI will fail.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMapExt.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMapExt.kt
@@ -1,0 +1,19 @@
+@file:JvmName("MapboxMapUtils")
+
+package com.mapbox.maps
+
+import com.mapbox.geojson.Point
+import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
+
+/**
+ * Calculates geographical coordinates(i.e., longitude-latitude pair) that corresponds
+ * to screen box.
+ *
+ * @param screenBox A box on the screen in \link MapOptions#size platform pixels \endlink.
+ *
+ * @return Returns a batch of geographical coordinates corresponding to the screen coordinates
+ * on the screen.
+ */
+fun MapCameraManagerDelegate.coordinatesForPixels(screenBox: ScreenBox): List<Point> {
+  return this.coordinatesForPixels(listOf(screenBox.min, screenBox.max))
+}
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/com/mapbox/maps/MapboxMapExt.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Calculates geographical coordinates(i.e., longitude-latitude pair) that corresponds
+ * to screen box.
+ *
+ * @param screenBox A box on the screen in \link MapOptions#size platform pixels \endlink.
+ *
+ * @return Returns a batch of geographical coordinates corresponding to the screen coordinates
+ * on the screen.
+ */
+fun MapCameraManagerDelegate.coordinatesForPixels(screenBox: ScreenBox): List<Point> {
+  return this.coordinatesForPixels(listOf(screenBox.min, screenBox.max))
```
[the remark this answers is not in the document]
Chaoba: Yes, but I think we can use this method temporarily and update it when we have the generated method

alexander-kulikovskii, line 18: is it important to get real `min` and `max` values? Because right now `ScreenBox` stores `ScreenCoordinate` as first and second argument.

alexander-kulikovskii: if user set in another order. is it critical for internal logic or not?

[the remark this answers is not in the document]
pengdev: I guess it would be hard to deprecate this method later when the method is directly exposed in MapboxMap, due to the same signature in Kotlin, and it will be breaking for Java users.

Chaoba: Then that would be a different ScreenBox, right?

alexander-kulikovskii: I mean setting `this.coordinatesForPixels(listOf((1,2), (4,5)))` and `this.coordinatesForPixels(listOf((4,5), (1,2)))` will lead to different results.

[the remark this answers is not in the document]
kiryldz: agree with @pengdev - if we really need it (however I do not see any linked task meaning it's a customer request) - we should add it in gl-native and expose both on Android and iOS at the same time. Otherwise @pengdev 's argument are 💯 correct - adding non-critical method that we're about to deprecate is not wise.

Chaoba:
Let's resolve this based on the discussion and generate the api from gl-native side.

PR #1202, opened 2022-03-04
Reset nativeMap to null in onDestroy

Chaoba:
### Summary of changes
In `MapboxMap` and `Style`, the reference to `nativeMap` will be reset to null in onDestroy, so any method calls related to `nativeMap` will throw `MapboxMapMemoryLeakException`

`tearDown` method in BaseMapTest.kt is removed to avoid duplicate method call to `MapboxMap.onDestroy()` as it will be triggerd by lifecycle plugin.

Automatically unsubscribe all observers in MapboxMap when `MapboxMap.onDestroy()` is invoked.

## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. If tests were not written, please explain why.
 - [ ] Optimize code for java consumption (`@JvmOverloads`, `@file:JvmName`, etc).
 - [ ] Add example if relevant.
 - [ ] Document any changes to public APIs.
 - [ ] Run `make update-api` to update generated api files, if there's public API changes, otherwise the `verify-kotlin-binary-compatibility` CI will fail.
 - [x] Apply changelog label ('breaking change', 'bug :beetle:', 'build', 'docs', 'feature :green_apple:', 'performance :zap:', 'testing :100:') or use the label 'skip changelog'
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-android` changelog: `<changelog></changelog>`.
 - [x] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.[version]` release branch.

Fixes: < Link to related issues that will be fixed by this pull request, if they exist >

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-android/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1230, opened 2022-03-17
Refactor native Map and Style to log error and not throw exception

kiryldz:
### Summary of changes

Add `MapboxMap.isValid()` and `Style.isValid()` methods. Map and style become invalid when `MapView.onDestroy` is called: after that map and style objects are treated as not valid and accessing any method will cause printing error log and introduce a memory leak in user's code. 

Another thing worse to mention - for async geojson parsing we do check for style being valid when parsing is over and we're about to send geojson to core.

Additionally going thru all our example activities to fix potential leaks.

### User impact (optional)

That PR enhances public `StyleInterface` with a new method `isValid(): Boolean`. Technically that's a breaking change but this interface is not intended to be actually overwritten by user (and we do not provide any APIs to register such interfaces) so we treat this as an enhancement or bug fix.

That PR _may_ break existing tests in case mocking `StyleInterface` in some very specific cases, this should be easily fixed with:
```
every { style.isValid() } returns true
```

## Pull request checklist:
    

tobrun (commented):
could we validate if having these strings inline doesn't result in a bunch of allocations/de-allocations with each function invocation? Else we need to make them constants instead

yunikkk:
Should we also remove WeakRefs from [NativeObserver](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/NativeObserver.kt#L12) and [StyleObserver](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/StyleObserver.kt#L19) ?

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -764,14 +834,23 @@ class Style internal constructor(
...
    */
   override fun isStyleLoaded(): Boolean {
-    return getStyleManager().isStyleLoaded
+    checkNativeStyle("isStyleLoaded")
+    return styleManager.isStyleLoaded
+  }
+
+  private fun checkNativeStyle(methodName: String) {
+    if (!isValid) {
+      Logger.e(TAG, "Accessing Style.$methodName after MapView is already destroyed meaning you are leaking Style object.")
```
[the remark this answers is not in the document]
kiryldz: User does not operate (and most likely does not even know) what is a native map object. Leak could happen if `MapView` was already destroyed but user keeps on working with cached variable for `Style` or `MapboxMap` object and accessing methods. That's precisely what I tried to express.

kiryldz:
@tobrun I do use `String.format` as we did in v9 maps here assuming that code there is efficient, how does this sound to you? 😄

tobrun:
> @tobrun I do use String.format now as we did in v9 maps here assuming that code there is efficient, how does this sound to you? smile

Yeah I was recently looking at that code and I was yikes this doesn't look good

kiryldz:
> > @tobrun I do use String.format now as we did in v9 maps here assuming that code there is efficient, how does this sound to you? smile
> 
> Yeah I was recently looking at that code and I was yikes this doesn't look good

In any case it should not matter - hitting that `if` block and writing an error log should happen rarely and signalize that user is leaking style or map object. I guess we should not overthink here how efficiently string is built 😄

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -764,14 +834,23 @@ class Style internal constructor(
    * @return TRUE if and only if the style JSON contents, the style specified sprite and sources are all loaded, otherwise returns FALSE.
    */
   override fun isStyleLoaded(): Boolean {
-    return getStyleManager().isStyleLoaded
+    checkNativeStyle("isStyleLoaded")
+    return styleManager.isStyleLoaded
+  }
+
+  private fun checkNativeStyle(methodName: String) {
```
alexander-kulikovskii, line 841: what about getting `methodName` from stack trace?
only one concern that it can speed down a little bit code.

btw, maybe mark this function as `inline`?

sdk/src/main/java/com/mapbox/maps/Style.kt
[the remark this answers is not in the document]
yunikkk: I mean - Style holds a lot of native memory, would be nice to explain it in more details, so that log message clearly said that : 
- Style/MapboxMap/etc should NOT be used after MapView is destroyed 
- It'll leak a lot of native memory in addition to Java object itself

sdk/src/main/java/com/mapbox/maps/Style.kt
kiryldz: `inline` does not take effect if not using functional parameter types (that's what IDE told me when I tried to apply). Getting method from stacktrace may indeed be unexpectedly slow so guess we should not risk here

sdk/src/main/java/com/mapbox/maps/Style.kt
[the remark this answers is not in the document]
kiryldz: ok, I'll try to rephrase message text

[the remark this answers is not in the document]
kiryldz: done

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
```diff
@@ -48,13 +50,31 @@ abstract class Source(
...
    *
    * @param delegate The style delegate
    */
   override fun bindTo(delegate: StyleInterface) {
     this.delegate = delegate
+    // geojson source is the only one holding async logic -
+    // in order to avoid accessing native style object when MapView is destroyed we need to check
+    // if delegate is still valid.
+    // Sadly due to current architecture it's now easily achievable with reflection only
+    // as this module does not have dependency on main SDK module and Style object
```
pengdev, line 66: why don't we add isValid to the StyleInterface? I think it's an acceptable addition as it is a bug fix, and using reflection is not efficient or safe.. And StyleInterface isn't supposed to be implemented by user at all..

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
```diff
@@ -93,6 +113,14 @@ abstract class Source(
...
     delegate?.let { styleDelegate ->
       try {
+        if (this is GeoJsonSource) {
+          val isNativeStyleValid = styleObjectIsValidMethod?.invoke(styleDelegate)
+          // explicitly reset native reference and return
+          if (isNativeStyleValid == false) {
+            delegate = null
+            return@let
+          }
+        }
```
pengdev, line 123: Consider the update property can be called very frequently, I'd suggest avoid using reflection. Introducing a new method in style interface is less harmful imo.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -47,31 +47,28 @@ class MapboxMap :
   MapCameraManagerDelegate,
   MapStyleStateDelegate {
 
-  private var nativeMap: MapInterface?
+  private val nativeMap: MapInterface
```
Chaoba, line 50: If users hold the reference to MapboxMap, it will also keep nativeMap and maybe have memory leak issue.

kiryldz: that's totally correct - that's the whole purpose of this PR. If users hold the ref to `MapboxMap` in their codebase and try to access any method after `MapView.onDestroy()` is triggered - we would print an error log for that. And yes, that's a leak, but not a leak from our end, it's a clear leak from user's end, we just clearly state that but not throw any exceptions as this could be a breaking behaviour change. Accessing native object after map view and renderer are destroyed should not bring in a crash, if it does in some cases - that should be fixed in gl-native side

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: my thinking was that it's still a public interface and adding new property / method there is still a breaking change. If we're OK with actually extending `StyleInterface` with a new property - I'm 2 hands for it.

pengdev: I think it's fine in this case, as user should not implement the StyleInterface besides writing some tests.. Would like thoughts from the Android team as well @mapbox/maps-android

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: as discussed with @yunikkk a bit and looking into e.g. [this](https://www.optaplanner.org/blog/2018/01/09/JavaReflectionButMuchFaster.html#:~:text=Adding%20setAccessible(true)%20call%20makes,takes%205.5%20nanoseconds%20per%20call.&text=Reflection%20is%20104%25%20slower%20than,takes%20longer%20to%20warm%20up) it seems that when method is found - calling it using reflection is pretty fast but in general I tend to agree, reflection is evil. I'll try extending an interface as proposed by you in https://github.com/mapbox/mapbox-maps-android/pull/1230#discussion_r830917178

kiryldz: reflection now removed

CHANGELOG.md
```diff
@@ -5,9 +5,7 @@ Mapbox welcomes participation and contributions from everyone.
...
-* Add exception type `MapboxMapMemoryLeakException` which will be thrown when users try to invoke functions related with `nativeMap` in `MapboxMap` and `Style` after onDestroy.
-  Users check the status of MapboxMap and Style by new method MapboxMap.isValid() and Style.isValid() to avoid such exception.
-  Automatically unsubscribe all observers in MapboxMap when `MapboxMap.onDestroy()` is invoked. ([1193](https://github.com/mapbox/mapbox-maps-android/pull/1193)) ([1202](https://github.com/mapbox/mapbox-maps-android/pull/1202))
+* Add `MapboxMap.isValid()` and `Style.isValid()` methods. Map and style become invalid when `MapView.onDestroy` is called: after that map and style objects are treated as not valid and accessing any method will cause printing error log and introduce a memory leak in user's code. Additionally unsubscribe all map observers when `MapboxMap.onDestroy()` is invoked. ([1193](https://github.com/mapbox/mapbox-maps-android/pull/1193)) ([1202](https://github.com/mapbox/mapbox-maps-android/pull/1202) ([1230](https://github.com/mapbox/mapbox-maps-android/pull/1230)))
```
yunikkk, line 8: Add `MapboxMap.isValid()` and `Style.isValid()` methods. `MapboxMap` and `Style` become invalid after `MapView.onDestroy`, calling any method will print error reminder. 
Additionally unsubscribe all map observers when `MapboxMap.onDestroy()` is invoked. ([1193](https://github.com/mapbox/mapbox-maps-android/pull/1193)) ([1202](https://github.com/mapbox/mapbox-maps-android/pull/1202) ([1230](https://github.com/mapbox/mapbox-maps-android/pull/1230)))
```
?

File: sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt [hunk 7]
```diff
@@ -15,6 +15,12 @@ interface StyleInterface : StyleManagerInterface {
    */
   val pixelRatio: Float
 
+  /**
```
yunikkk, line 18: We're 100% sure that interface is not implemented with any customer, right?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: Let's add documentation for that.
Add a note at to the top-level of the file, also to the `MapView.getMapboxMap` highlighting that MapboxMap keeps a lot of memory with it and should NOT be cached outside of the activity.

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -20,24 +21,22 @@ import java.nio.ByteBuffer
...
-    return styleManager ?: throw MapboxMapMemoryLeakException()
+  internal fun onMapViewDestroyed() {
+    isMapValid = false
   }
 
   /**
-   * Whether the Style instance is valid. Style will be invalid after MapView is destroyed or a new style is loaded and
-   * accessing Style will throw [MapboxMapMemoryLeakException].
+   * Whether the Style instance is valid.
+   * Style will be invalid after MapView is destroyed or a new style is loaded.
```
yunikkk, line 36: Is it correct about the `or a new style is loaded.` ?

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
kiryldz: yes, it's our abstractions. While for plugin we _potentially_ have a way to overwrite our default plugins - we do not have that for extensions and this interface is used with style extension only.
@pengdev @tobrun to confirm my understanding is correct here

sdk/src/main/java/com/mapbox/maps/Style.kt
kiryldz: it's interesting but as per current impl it's indeed not true, I'll rework it

kiryldz: created follow-up task and fixed comment here to reflect true state of things

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
pengdev: StyleInterface is implemented in the SDK module itself, we allow user to overwrite the plugin implementations, but the SDK module itself shouldn't and couldn't be implemented by the end users.

The cases I can think of user use the StyleInterface is that they could mock it in their tests.

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
```diff
@@ -15,6 +15,18 @@ interface StyleInterface : StyleManagerInterface {
...
+   * Whether the Style instance is valid.
+   *
+   * Style becomes invalid after MapView.onDestroy() is invoked,
+   * calling any method then could result in undefined behaviour and will print an error log.
+   *
+   * Keeping the reference to an invalid Style instance introduces significant native memory leak.
```
yunikkk, line 24: Pls also add similar comments to the MapView.getMapboxMap / MapView.getStyle

yunikkk (approved):

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
kiryldz: @pengdev sounds a bit strange to me still as we're operating with `Style` in our public API. Mostly it's one place:
```
/**
   * Callback to be invoked when a style has finished loading.
   */
  fun interface OnStyleLoaded {
    /**
     * Invoked when a style has finished loading.
     *
     * @param style the style that has finished loading
     */
    fun onStyleLoaded(style: Style)
  }
``` 
so for an end user guess it makes sense to mock `Style` then... In any case I'll drop instructions in PR description for that case.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: guess we decided to go with that solution

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
kiryldz: 👍

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -61,20 +60,19 @@ internal class StyleObserver(
...
    */
   override fun onStyleLoaded(eventData: StyleLoadedEventData) {
-    nativeMapWeakRef.get()?.let {
-      val style = Style(it, pixelRatio)
-      // Cache all the weak reference for new styles
-      styleWeakReferenceList.add(WeakReference(style))
-      styleLoadedListener.onStyleLoaded(style)
+    val style = Style(nativeMap, pixelRatio)
+    // cache loaded style as users may have saved strong reference in their code -
+    // so we want to mark that style as not valid when MapView will be destroyed
```
Chaoba, line 65:
```suggestion
    // so we want to mark that style as not valid when MapView is destroyed
```

Chaoba (approved):
Left a question, others LGTM. Thank you for taking this.

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -27,7 +26,7 @@ internal class StyleObserver(
   private var loadStyleTransitionOptions: TransitionOptions? = null
 
   private val getStyleListeners = CopyOnWriteArraySet<Style.OnStyleLoaded>()
-  private val styleWeakReferenceList = CopyOnWriteArraySet<WeakReference<Style>>()
+  private val loadedStyleList = CopyOnWriteArraySet<Style>()
```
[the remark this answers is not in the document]
kiryldz: I see your point but taking into account that we do have a single place in code to destruct everything - working with strong refs is better imho. In any case we will revisit that logic in tailwork.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
```diff
@@ -93,6 +93,14 @@ abstract class Source(
   private fun updateProperty(property: PropertyValue<*>, throwRuntimeException: Boolean = true) {
     delegate?.let { styleDelegate ->
       try {
+        // checking for validness makes sense only for geojson as it uses async parsing
+        if (this is GeoJsonSource) {
+          // explicitly reset native reference and return if map is not valid anymore
+          if (!styleDelegate.isValid()) {
+            delegate = null
+            return@let
+          }
```
pengdev, line 102: Would it make sense to add some error logs here?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/SourceExt.kt
```diff
@@ -17,11 +17,11 @@ import com.mapbox.maps.extension.style.utils.silentUnwrap
 fun StyleManagerInterface.getSource(sourceId: String): Source? {
```
pengdev, line 17: Not in this release, but would consider to directly extend `StyleInterface` in the next major release.

sdk-base/src/main/java/com/mapbox/maps/extension/style/StyleInterface.kt
pengdev: 👍 didn't remember we return Style async instead of StyleInterface

pengdev (commented):
LGTM with 1 comment, really like the documentation updates to the `MapboxMap` and` Style`, great work!

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: it's non-error situation - map destroyed while we were parsing so see no need in logs here

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/SourceExt.kt
kiryldz: 👍

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/Source.kt
kiryldz: that code is not correct when using persistent layers, rollbacked in https://github.com/mapbox/mapbox-maps-android/pull/1324

PR #1255, opened 2022-04-04
Introduce map projection as part of the style-spec

kiryldz:
## Pull request checklist:
    

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1669,6 +1688,14 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase {
     )
   }
 
+  /**
+   * Called when new style is loaded.
+   */
+  override fun onStyleChanged(styleDelegate: StyleInterface) {
+    Logger.e("KIRYLDD", "new style: $styleDelegate")
```
alexander-kulikovskii, line 1695:
```suggestion
    Logger.e(TAG, "new style: $styleDelegate")
```

alexander-kulikovskii: or maybe remove it. it's not an `error` level

alexander-kulikovskii (approved):

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
```diff
@@ -533,6 +533,46 @@ enum class Anchor(override val value: String) : LayerProperty {
    */
   VIEWPORT("viewport"),
 }
+// NAME: The name of the projection to be used for rendering the map.
+/**
+ * The name of the projection to be used for rendering the map.
+ *
+ * @param value String value of this property
+ */
+enum class Name(override val value: String) : LayerProperty {
```
alexander-kulikovskii, line 542:
```suggestion
enum class ProjectionName(override val value: String) : LayerProperty {
```

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: thanks, already removed 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
kiryldz: this is generated code based on the style.. but guess it makes sense, I'll try to update generator code.

kiryldz: sadly that will overcomplicate generator code... IMO we could leave this as is, we do not actually change property names for generated code in general from what I see. In general this enum will not be used standalone, it's used only in Style DSL + methods like `StyleInterface.getProjectionName` / `StyleInterface.setProjectionName`.

@mapbox/maps-android @mapbox/maps-ios - what are your thoughts here?

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
```diff
@@ -533,6 +533,46 @@ enum class Anchor(override val value: String) : LayerProperty {
...
+  NATURALEARTH("naturalEarth"),
+  /**
+   * A Winkel Tripel projection.
+   */
+  WINKELTRIPEL("winkelTripel"),
+  /**
+   * A globe projection.
+   */
+  GLOBE("globe"),
+}
```
macdrevx, line 575: @kiryldz does gl native support all of these? I thought only globe and mercator were supported so far.

kiryldz: I did add the test covering this, please see actual instrument tests in `StyleProjectionTest` file.

extension-style/api/metalava.txt
```diff
@@ -2,16 +2,18 @@
 package com.mapbox.maps.extension.style {
 
   public final class StyleExtensionImpl implements com.mapbox.maps.extension.style.StyleContract.StyleExtension {
-    method public java.util.List<com.mapbox.maps.extension.style.StyleContract.StyleImageExtension> getImages();
-    method public java.util.List<kotlin.Pair<com.mapbox.maps.extension.style.layers.Layer,com.mapbox.maps.LayerPosition>> getLayers();
+    method public error.NonExistentClass! getImages();
+    method public error.NonExistentClass! getLayers();
```
ank27, line 6: Are we adding `error.NonExistentClass` thing in metalava 🙈 . I guess its needed to pass the checks, correct ?

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
ank27: `ProjectionName` does sound better than just `Name`. but if it will overcomplicate the generator code. we could leave it.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
ank27: can we use `_` in the names?. for eg. `LAMBERT_CONFORMAL_CONIC` instead of `LAMBERTCONFORMALCONIC`.

extension-style/api/metalava.txt
kiryldz: I simply did run the script and got following output. Already reverted that, we need to track separately why metalava did break

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
kiryldz: @ank27 it's not only about naming here in the enum. We should re-organize generator code to understand this as well when generating methods `name()`in `Projection` file. Ideally we want something like:
```
Projection.name(name: ProjectionName)
```
but it's a bit tricky to achieve and we don't do this anywhere else.

@pengdev - as the most experienced person in generated code - what are your thoughts here?

sdk-base/api/metalava.txt
```diff
@@ -589,12 +595,12 @@ package com.mapbox.maps.plugin {
...
-    ctor public LocationPuck3D(String modelUri, java.util.List<java.lang.Float> position = listOf(0.0, 0.0), float modelOpacity = 1f, java.util.List<java.lang.Float> modelScale = listOf(1.0, 1.0, 1.0), String? modelScaleExpression = null);
-    ctor public LocationPuck3D(String modelUri, java.util.List<java.lang.Float> position = listOf(0.0, 0.0), float modelOpacity = 1f, java.util.List<java.lang.Float> modelScale = listOf(1.0, 1.0, 1.0));
-    ctor public LocationPuck3D(String modelUri, java.util.List<java.lang.Float> position = listOf(0.0, 0.0), float modelOpacity = 1f);
-    ctor public LocationPuck3D(String modelUri, java.util.List<java.lang.Float> position = listOf(0.0, 0.0));
+    ctor public LocationPuck3D(String modelUri, java.util.List<java.lang.Float> position = <anonymous class>(0.0, 0.0), float modelOpacity = 1f, java.util.List<java.lang.Float> modelScale = <anonymous class>(1.0, 1.0, 1.0), String? modelScaleExpression = null, java.util.List<java.lang.Float> modelTranslation = <anonymous class>(0.0, 0.0, 0.0), java.util.List<java.lang.Float> modelRotation = <anonymous class>(0.0, 0.0, 90.0));
```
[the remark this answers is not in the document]
ank27: maybe we can leave metalava changes as it is, so ci will pass for other / future prs too? wdyt?

ank27:
LGTM with few nits 💎

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
kiryldz: @ank27 fixed!

macdrevx: I think putting these constants in the public API will suggest to developers that they can actually be used and will result in avoidable support burden.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,166 @@
...
+     */
+    get() {
+      getPropertyValue<Expression>("name")?.let {
+        return it
+      }
+      name?.let {
+        return Expression.literal(it.value)
+      }
+      return null
+    }
```
macdrevx, line 70: In what situation would you need to use this?

macdrevx: I'm looking at the style spec and [docs](https://docs.mapbox.com/mapbox-gl-js/style-spec/projection/) and it's not clear to me that expressions are even supported for Projection.name

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
pengdev: @kiryldz we can overwrite the style spec json in the style-parser.js, I think it makes sense to rename `name` to `projection_name`. We had similar tweaks to rename `GeojsonSource` to `GeoJsonSource` in the style-parser.js, you can use it as a reference.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
pengdev: we can remove these unsupported enums from the style-parser.js for now, and add them back when they become available to gl-native. The only concern is that adding something to enum later would be breaking change.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
pengdev: let's double check with gl-native team.

kiryldz: verified that our basic expressions are supported, e.g setting.:
```
concat {
  literal("glo")
  literal("b")
  literal("e")
}
```
works fine

kiryldz: @macdrevx in general we do support setting property as expression everywhere in generated style code, if it does not apply for iOS - I guess it's fine not to expose it

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/properties/generated/Property.kt
kiryldz: @pengdev I did remove not supported enums. Good point about breaking change, let me ask if gl-native is planning to support other projections in nearest future.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1683,35 +1683,6 @@ class MapboxMap :
     isMapValid = false
   }
 
-  /**
-   * Set map projection for Mapbox map.
-   *
-   * @param mapProjection either [MapProjection.Globe] or [MapProjection.Mercator] projection that will be applied to Mapbox map.
-   */
-  @MapboxExperimental
-  override fun setMapProjection(mapProjection: MapProjection) {
```
pengdev, line 1692: wonder if we should Deprecate first and then remove, it would likely break user's code? but as it is experimental, I guess it's fine to just remove.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,166 @@
...
+   */
+  fun name(name: ProjectionName = ProjectionName.MERCATOR): Projection
+
+  /**
+   * Set the Name property
+   *
+   * @param name value of name as Expression
+   */
+  fun name(name: Expression): Projection
+}
```
pengdev, line 157: Not in this PR, we might want to get rid of the DslReceiver interfaces in v11 and use Builder pattern instead to avoid breaking change.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/ProjectionExt.kt
```diff
@@ -0,0 +1,66 @@
...
+fun StyleInterface.getProjectionName(): ProjectionName {
+  getStyleProjectionProperty("name").apply {
+    return if (kind == StylePropertyValueKind.UNDEFINED) {
+      ProjectionName.MERCATOR
+    } else {
+      val projectionAsString = value.unwrapToAny() as String
+      return ProjectionName.valueOf(projectionAsString.uppercase())
+    }
+  }
+}
```
pengdev, line 25: We could probably just have getProjection/setProjection methods? the API should already be easy to use with these 2 extensions. Otherwise, we need to introduce additional project properties extension functions as well, which seems a bit overkill.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/ProjectionExt.kt
```diff
@@ -0,0 +1,66 @@
...
+  projection.bindTo(this)
+  return projection
+}
+
+/**
+ * Extension function to set the Projection by name provided by the Style Extension to the Style.
+ *
+ * @param projectionName The projection [ProjectionName] to be set.
+ */
+fun StyleInterface.setProjectionName(projectionName: ProjectionName) {
```
pengdev, line 51: propose to remove this extension

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/ProjectionExt.kt
pengdev: Still wonder the necessity to expose `getProjectionName` and `setProjectionName` extension functions. Have we done similar extension in for example `terrain` and `light`?  I think it's a bit inconsistent with other APIs, here we are providing the property below the `Projection` abstraction.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/ProjectionExt.kt
pengdev: I think as we have projection DSL, it could be simplified as 
```
style.setProjection(
  projection {
    name(ProjectionName.GLOBE)
  } 
)
```

I think it's a more aligned API interface with terrain and light, and it's also future-proof.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,104 @@
...
+      return try {
+        it.getStyleProjectionProperty(propertyName).unwrap()
+      } catch (e: RuntimeException) {
+        Log.e(TAG, "Get projection property failed: ${e.message}")
+        Log.e(TAG, it.getStyleProjectionProperty(propertyName).toString())
+        null
+      }
+    }
+    throw MapboxStyleException("Get property $propertyName failed: projection is not added to style yet.")
+  }
```
macdrevx, line 73: Unused?

sdk-base/src/main/java/com/mapbox/maps/plugin/MapProjection.kt
```diff
@@ -1,65 +0,0 @@
...
-     * If MapboxMap is using [MapProjection.Globe] and current map zoom level is >= [TRANSITION_ZOOM_LEVEL] -
-     * map will use [MapProjection.Mercator] and [MapProjectionDelegate.getMapProjection] will return [MapProjection.Mercator].
-     *
-     * If MapboxMap is using [MapProjection.Globe] and current map zoom level is < [TRANSITION_ZOOM_LEVEL] -
-     * map will use [MapProjection.Globe] and [MapProjectionDelegate.getMapProjection] will return [MapProjection.Globe].
-     *
-     * If MapboxMap is using [MapProjection.Mercator] - map will use [MapProjection.Mercator] for any zoom level and
-     * [MapProjectionDelegate.getMapProjection] will return [MapProjection.Mercator].
-     */
-    const val TRANSITION_ZOOM_LEVEL = 5.0
```
macdrevx, line 63: Should we keep this constant?

macdrevx, line 63: Should we keep this constant?

macdrevx: I'm leaning toward no (or if we do keep it, exposing it in the GL Native interface) since it seems like something that could be subject to change.

macdrevx: I'm leaning toward no (or if we do keep it, exposing it in the GL Native interface) since it seems like something that could be subject to change.

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/ProjectionExt.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+fun StyleInterface.getProjection(): Projection {
+  getStyleProjectionProperty("name").apply {
+    return Projection(
+      if (kind == StylePropertyValueKind.UNDEFINED) {
+        ProjectionName.MERCATOR
+      } else {
+        val projectionAsString = value.unwrapToAny() as String
+        ProjectionName.valueOf(projectionAsString.uppercase())
+      }
```
macdrevx, line 24: Would it make more sense to check for `kind == string` and have the else be the branch that returns mercator as a fallback? Not sure if GL Native would ever return a property value with a different kind, but verifying string before we cast to string seems more robust.

sdk-base/src/main/java/com/mapbox/maps/plugin/MapProjection.kt
kiryldz: previously we did need that constant because map in globe projection started returning mercator when zoom > TRANSITION_ZOOM_LEVEL. now transition is not time-based but zoom-based and getter always returns globe so there is no practical need for that constant (and even previously it was doubtful that we need it)

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
kiryldz: indeed not used as several other things in this class as well, will remove them

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/**
+ * This Interface contains all the functions that will be exposed to Koltin DSL.
+ *
+ * Separated the DSL receiver class to this interface to avoid IDE code suggestion for
+ * property getters.
+ */
+@ProjectionDsl
+interface ProjectionDslReceiver
```
pengdev, line 90: seems this is not needed? One question, how can we get the Projection class from the gl-native? as `getPropertyValue` is never used?

kiryldz: yep, will drop it

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,83 @@
...
+    expected?.error?.let {
+      throw MapboxStyleException("Set projection property failed: $it")
+    }
+  }
+
+  /**
+   * Static variables and methods.
+   */
+  private companion object {
+    const val TAG = "Mbgl-Projection"
```
pengdev, line 72: nit: not used

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+/**
+ * DSL function for [Projection].
+ */
+fun projection(name: ProjectionName): Projection {
+  return Projection(name)
+}
```
pengdev, line 81: nit: as we don't have properties now, this DSL could be removed

pengdev (approved):
LGTM with nits

extension-style/src/main/java/com/mapbox/maps/extension/style/projection/generated/Projection.kt
kiryldz: this DSL function is used in Style DSL to be consistent with other layers / sources we add e.g.
```
+projection(ProjectionName.GLOBE)
```

PR #1316, opened 2022-04-29
[sdk] add thread checking

tobrun:
Proposal to incorporate thread checking.

### Summary of changes

This PR adds main thread checking functionality to the native map. If the map is accessed through a worker thread, it will throw a `WorkerThreadException`. We have seen a couple of occurrences were developers get sporadic crashes because they were trying to manipulate the style object or try to query the map from a worked thread. This PR prevents them from making these mistakes and by accident releasing this publicly. 

The PR does try to attempt limiting the impact of thread checking to only do it for debug builds. When running in release, we won't perform this thread check. To make this work, I did had to use relfection to lookup the BuildConfig of the integrating application.  

This PR also allows to disable thread checking fully by inclusion of an Android Manifest entry:
```
        <meta-data
            android:name="com.mapbox.maps.ThreadChecker"
            android:value="false" />
```

## Pull request checklist:
    

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
```diff
@@ -7,55 +7,69 @@ import com.mapbox.common.Cancelable
...
+import com.mapbox.maps.exception.WorkerThreadException
 import java.util.*
 
 internal class NativeMapImpl(private val map: MapInterface) :
   MapInterface, StyleManagerInterface, ObservableInterface {
 
+  private val threadChecker = ThreadChecker()
+
   override fun setSize(size: Size) {
+    checkThread()
```
kiryldz, line 19: we already have `checkNativeMap` for `MapboxMap` which is actually the public API surface - perhaps add `checkThread()` there instead? delta will be very small then

tobrun: :+1:

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
```diff
@@ -7,55 +7,69 @@ import com.mapbox.common.Cancelable
 import com.mapbox.geojson.Feature
 import com.mapbox.geojson.Geometry
 import com.mapbox.geojson.Point
+import com.mapbox.maps.exception.WorkerThreadException
 import java.util.*
 
 internal class NativeMapImpl(private val map: MapInterface) :
   MapInterface, StyleManagerInterface, ObservableInterface {
 
+  private val threadChecker = ThreadChecker()
```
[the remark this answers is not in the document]
tobrun: :100: possible but just wanted to point out one limitation of KSP. You can't adjust any existing classes, you can only generate new code, not inline inject it. So we would need some kind of factory/impl setup to make it work. 
This all said, I do think we could do this code generated in our underlying bindings generator.

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
```diff
@@ -7,55 +7,69 @@ import com.mapbox.common.Cancelable
...
     map.createRenderer()
   }
 
   override fun destroyRenderer() {
+    checkThread()
     map.destroyRenderer()
   }
 
   override fun render() {
+    checkThread()
```
yunikkk, line 39: Seems that at least this `render` is not called from the main thread

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,24 @@
+package com.mapbox.maps
+
+import android.os.Looper
+import com.mapbox.common.MapboxSDKCommon
+
+class ThreadChecker {
```
alexander-kulikovskii, line 6:
```suggestion
internal class ThreadChecker {
```

yunikkk:
Suggestions from slack : 
```
kiryldz
however still think having API flag to control it may be useful even for debug builds

Dmitry Yunitski
What if we log warning for all the builds, release and debug
And if the API flag is set (like strict mode in Android) - throw independently of the build type
```

yunikkk:
Btw similar is implemented in google's lifecycle https://androidx.tech/artifacts/lifecycle/lifecycle-runtime/2.3.0-source/androidx/lifecycle/LifecycleRegistry.java.html, `enforceMainThreadIfNeeded` method

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,27 @@
...
+    debug = buildConfigClass.getField("DEBUG").getBoolean(null)
+  }
+
+  fun isMainThread(): Boolean {
+    val startTime = System.nanoTime()
+    if (debug && Thread.currentThread() != Looper.getMainLooper().thread) {
+      return false
+    }
+    val diff = System.nanoTime() - startTime
+    Log.e("TVN", "Main thread checking took: $diff nano")
```
kiryldz, line 24: some debug logs that we need not forget to remove if we'll shipit

sdk/src/main/java/com/mapbox/maps/NativeMapImpl.kt
tobrun: added now in mapboxmap and style

tobrun:
> Btw similar is implemented in google's lifecycle https://androidx.tech/artifacts/lifecycle/lifecycle-runtime/2.3.0-source/androidx/lifecycle/LifecycleRegistry.java.html, enforceMainThreadIfNeeded method

ArchTaskExecutor.isMainThread can only be called from within the same library group prefix.

alexander-kulikovskii (approved):

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import android.os.Looper
+import android.util.Log
+import com.mapbox.common.MapboxSDKCommon
+
+internal object ThreadChecker {
+
+  // Captures if the consumer application is running in debug
+  private var debug = false
+  private const val TAG = "mbgl-ThreadChecker"
```
kiryldz, line 11: with new logging no need to add `mbgl-`, just `ThreadChecker` will be fine and will result in:
```
I/Mapbox: [maps-android\Mbgl-Renderer]: message
```

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,31 @@
...
+  private const val DEBUG_FIELD = "DEBUG"
+  private const val LOOKUP_FAILED = "Unable to lookup build config of application."
+
+  init {
+    try {
+      val packageName = MapboxSDKCommon.getContext().applicationInfo.packageName
+      val buildConfigClass = Class.forName("$packageName.BuildConfig")
+      debug = buildConfigClass.getField(DEBUG_FIELD).getBoolean(null)
+    } catch (exception: Exception) {
+      Log.v(TAG, LOOKUP_FAILED, exception)
```
kiryldz, line 21:
```suggestion
      logI(TAG, LOOKUP_FAILED, exception)
```

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,37 @@
...
+      val packageName = MapboxSDKCommon.getContext().applicationInfo.packageName
+      val buildConfigClass = Class.forName("$packageName.BuildConfig")
+      debug = buildConfigClass.getField(DEBUG_FIELD).getBoolean(null)
+    } catch (exception: Exception) {
+      logI(TAG, "$LOOKUP_FAILED ${exception.localizedMessage}")
+    }
+  }
+
+  fun throwIfNotMainThread() {
+    if (isMainThread()) {
```
alexander-kulikovskii, line 26:
```suggestion
    if (!isMainThread()) {
```
or
```suggestion
    if (isMainThread().not()) {
```

kiryldz (commented):
it seems that tests are failing and also would be great to rebase on latest main

sdk/src/main/java/com/mapbox/maps/ThreadChecker.kt
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  fun throwIfNotMainThread() {
+    if (!isMainThread()) {
+      throw WorkerThreadException()
+    }
+  }
+
+  private fun isMainThread(): Boolean {
+    if (debug && Thread.currentThread() != Looper.getMainLooper().thread) {
```
yunikkk, line 32: Nit : imo `isMainThread` should not check `debug` flag, from the name itself it's natural to just check the thread. WDYT about moving `debug` flag check to `throwIfNotMainThread` instead?

yunikkk: In addition this one will be simpler `private fun isMainThread() = Thread.currentThread() == Looper.getMainLooper().thread)`

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1767,6 +1767,7 @@ class MapboxMap :
   }
 
   private fun checkNativeMap(methodName: String) {
+    ThreadChecker.throwIfNotMainThread()
```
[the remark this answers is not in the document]
tobrun: `queryRenderFeatures` not sure if I have the right context but historically all methods from MapboxMap should be invoked from the main thread. If we move to something else, that would happen under the hood.

sdk/src/main/java/com/mapbox/maps/exception/WorkerThreadException.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import java.lang.RuntimeException
+
+/**
+ * Exception thrown when a worker thread calls in a main thread only class.
+ */
+class WorkerThreadException : RuntimeException(
+  """
+  The exception that is thrown when an application attempts to 
+  perform a map operation on a worked thread.
```
yunikkk, line 11:
```suggestion
  perform a map operation on a worker thread.
```

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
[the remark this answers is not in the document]
yunikkk: E.g. https://github.com/mapbox/mapbox-maps-android/commit/b91fa7ff0084e813a479395e7b5aaa05983efa1e#diff-42b07126c22a902e2dd8b68cda6f045eebbf1dce0a43847bbd3daefefcc12f63R67

sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt
```diff
@@ -901,6 +954,6 @@ class CameraAnimationsPluginTest : BaseAnimationMapTest() {
   companion object {
     private const val TAG = "Mbgl-CameraManager"
     private const val EPS = 0.000001
-    private const val LATCH_MAX_TIME = 5_000L
+    private const val LATCH_MAX_TIME = 6_000L
```
yunikkk, line 957: Why 6?)

tobrun: sorry, test for not getting the test to pass, is now removed

app/src/androidTest/java/com/mapbox/maps/testapp/observable/ObservableExtensionTest.kt
```diff
@@ -142,11 +149,15 @@ class ObservableExtensionTest : BaseMapTest() {
     if (!latch.await(20000, TimeUnit.MILLISECONDS)) {
       throw TimeoutException()
     }
-    mapboxMap.removeOnMapLoadErrorListener(listener)
+
+    rule.scenario.onActivity { activity ->
+      activity.runOnUiThread {
+        mapboxMap.removeOnMapLoadErrorListener(listener)
```
kiryldz, line 155: just curious - don't we see any errors / warning in logcat that we're trying to access outdated map object when running this test?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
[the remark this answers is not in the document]
kiryldz: @tobrun QRF is a method that is OK to be run from render thread (as render thread has looper and thus scheduler). Moreover it's the only way now to perform synchronous QRF which is needed e.g. to understand whether we need to consume annotation click or not, please refer to @yunikkk's link above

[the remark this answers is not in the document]
tobrun: sounds good, updated all query methods to not check main thread

Chaoba (approved):

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -1766,7 +1766,10 @@ class MapboxMap :
     return nativeMap.setViewAnnotationPositionsUpdateListener(listener)
   }
 
-  private fun checkNativeMap(methodName: String) {
+  private fun checkNativeMap(methodName: String, checkMainThread: Boolean = true) {
```
yunikkk, line 1769: Suppose the flag `checkMainThread` should be false for QRF?

kiryldz: from what I see - it is

yunikkk:
@tobrun should we somehow advertise this check to our users so that they could validate the usage of maps? IMHO changelog entry isn't really enough to discover for many

tobrun:
```
@tobrun should we somehow advertise this check to our users so that they could validate the usage of maps? IMHO changelog entry isn't really enough to discover for many
```
Could you add more color on how we would achieve that?
I rewrote the changelog entry to have a bit more color.
To me, crashing with a clear exception should be a good indication on what is wrong or what has changed :)

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Mapbox welcomes participation and contributions from everyone.
 * Remove `MapboxExperimental` annotation from viewport plugin and promote viewport plugin as stable API. ([1425](https://github.com/mapbox/mapbox-maps-android/pull/1425))
 * Introduce `addRendererSetupErrorListener`/`removeRendererSetupErrorListener` methods for `MapView` and `MapSurface` to listen to renderer setup errors and give opportunity to control some edge cases. ([1427](https://github.com/mapbox/mapbox-maps-android/pull/1427))
 * Introduce transition properties for atmosphere and terrain. ([1451](https://github.com/mapbox/mapbox-maps-android/pull/1451))
+* Enable main thread checking on the map/style object when running applications in debug mode. It's required to call these object functions on the main thread, otherwise you will hit edge case crashes.
```
pengdev, line 16:
```suggestion
* Enable main thread checking on the map/style object when running applications in debug build. It's required to call these object functions on the main thread, otherwise you will hit edge case crashes. 
```

pengdev (approved):

yunikkk:
> Could you add more color on how we would achieve that?
> I rewrote the changelog entry to have a bit more color.
> To me, crashing with a clear exception should be a good indication on what is wrong or what has changed :)
Honestly I don't see clear way, just seems the changelog message is a bit blury, not really clear that debug applications will crash if maps is called from the bg thread

tobrun:
> Honestly I don't see clear way, just seems the changelog message is a bit blurry, not really clear that debug applications will crash if maps is called from the bg thread

Updated the changelog to include that is will crash

PR #1330, opened 2022-05-05
[lifecycle] Introduce ViewLifecycleOwner to have granular control over MapView's lifecycle. - MAPSAND-76

pengdev:
### Summary of changes

This PR ports [ViewLifecycleRegistry](https://github.com/mapbox/mapbox-navigation-android/blob/main/libnavui-util/src/main/java/com/mapbox/navigation/ui/utils/internal/lifecycle/ViewLifecycleRegistry.kt) logic from Nav SDK to the Maps SDK as `ViewLifecycleOwner`. 

Previously when the MapView is added to the activity or fragment, the lifecycle plugin will automatically get the hosting lifecycle from the Activity or fragment, thus receive lifecycle events to start, stop or destroy the MapView. However the issue could raise when the mapView is removed/detached from the containing activity or fragment programmatically, and in these cases, the lifecycle plugin still monitors the lifecycle of the container activity/fragment and the MapboxRenderer still tries to render the map. In the mentioned cases, the mapview's lifecycle should be detached from the container's lifecycle and stop rendering.

This PR solves the above issue by introducing the `ViewLifecycleOwner` that merges the hosting lifecycle (of activity or fragment) with the view's surface lifecycle.

## Pull request checklist:
    

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -10,58 +10,57 @@ import android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE
...
    */
   override fun registerLifecycleObserver(mapView: View, observer: MapboxLifecycleObserver) {
-    val lifecycleOwner = ViewTreeLifecycleOwner.get(mapView)
-    if (lifecycleOwner == null) {
-      logE(
-        TAG,
+    viewLifecycleRegistry = ViewLifecycleRegistry(
+      view = mapView,
+      localLifecycleOwner = this,
+      hostingLifecycleOwner = requireNotNull(ViewTreeLifecycleOwner.get(mapView)) {
```
LukasPaczos, line 35: This seems unnecessarily late. Aren't there cases where `getLifecycle` could be called before the view is attached to window? Couldn't we pass the `mapView` to the constructor of the plugin and setup the registry right-away? Also we don't need the view tree to find the lifecycle owner, we just need the context - https://github.com/mapbox/mapbox-navigation-android/blob/889e9f383a3777aef700df8b720e5120da6b7c1f/libnavui-dropin/src/main/java/com/mapbox/navigation/dropin/internal/extensions/ContextEx.kt#L21.

pengdev: > This seems unnecessarily late. Aren't there cases where getLifecycle could be called before the view is attached to window? Couldn't we pass the mapView to the constructor of the plugin and setup the registry right-away? 

@LukasPaczos this is due to our plugin setup, where we don't have direct access to MapView inside the lifecycle plugin. And currently we do only call the `getLifecycle` and `registerLifecycleObserver` [when the MapView is attached to window](https://github.com/mapbox/mapbox-maps-android/blob/d5c1fdf195133f6e6719bc6beb280945c9b8f307/sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt#L175), during our plugin initialisation phase.

> Also we don't need the view tree to find the lifecycle owner, we just need the context

I wonder what's the benefit to loop through the context recursively comparing to using in stock API for it? As [this article](https://medium.com/@takahirom/about-the-newly-added-viewtreelifecycleowner-4588e4013ca2) describes, Android has hooked the `ViewTreeLifecycleOwner` setter to `ComponentActivity` and `Fragment` and handles everything internally?

LukasPaczos: > we don't have direct access to MapView inside the lifecycle plugin

Looking though the code, you could pass the mapView when the instance is created in `MapController#initializePlugins`, couldn't you?

> I wonder what's the benefit to loop through the context recursively comparing to using in stock API for it?

`ViewTreeLifecycleOwner` will only return the lifecycle owner of a view that is attached to a view hierarchy, won't it? That's what I see from the source code. So if we wanted to initialize the registry on creation of the plugin as suggested above, the view wouldn't yet be attached to the window and `ViewTreeLifecycleOwner.get` would return `null`. That's my understanding but it would be good to double-check.

pengdev: > I didn't notice the implementation is public (should it be?), in this case it might not make sense to introduce the constructor parameter that interferes with the registration method. 

The implementation is public so that we can initialise it in the sdk without using reflection.

> The problem is that MapboxLifecyclePluginImpl#getLifecycle will throw an uninitialized exception when called before the view is attached to the window. That said, we are not actually using the lifecycle scope created by this owner, we just rely on lifecycle callbacks to pass them down to MapView, so why don't we inject a dummy LifecycleOwner into the registry instead of MapboxLifecyclePluginImpl implementing the interface?

Sounds good, I now made the localLifecycleOwner a dummy private property instead letting MapboxLifecyclePluginImpl implementing the interface.
> 
> In the ideal situation, the MapView should be the LifecycleOwner, but since the lifecycle plugin is optional, I don't think we can do that. For v11 we could evaluate removing the lifecycle plugin and just always handling the lifecycle out-of-the-box 🙃

Yes, I'm definitely for it in v11.

pengdev: > > with this change, I can see the ON_DESTROY is not fired when the fragment is destroyed
> 
> Interesting, is the returned owner tied to the fragment or to the activity? In any case, the problem shouldn't exist if we incorporate above, but this could pose a similar problem for Nav SDK if that's the case.

@LukasPaczos I took a deeper look and the returned owner is tied to the activity it seems, and `Context.recursiveUnwrap()` returns the current context immediately. For more context, the MapFragment extends [androidx.fragment.app.Fragment](https://github.com/mapbox/mapbox-maps-android/blob/2e3e0123f6be7ff39d72d8b59c75d1760900ec0b/app/src/main/java/com/mapbox/maps/testapp/examples/fragment/MapFragment.kt#L12), it is neither a `Activity` or `ContextWrapper`, I think the `Context.recursiveUnwrap()` fall short in this use case, and probably should be revisited in the navigation SDK.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -10,58 +10,65 @@ import android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE
...
  * Concrete implementation of MapboxLifecyclePlugin.
  */
-class MapboxLifecyclePluginImpl : MapboxLifecyclePlugin {
+class MapboxLifecyclePluginImpl constructor(mapView: View?) : MapboxLifecyclePlugin,
+  LifecycleOwner {
+  private lateinit var viewLifecycleRegistry: ViewLifecycleRegistry
+
+  init {
+    mapView?.let {
+      viewLifecycleRegistry = ViewLifecycleRegistry(
```
yunikkk, line 28: If no subscriptions is made here - what is the point to init viewLifecycleRegistry?

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -10,58 +10,65 @@ import android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE
...
 import com.mapbox.maps.logW
 import com.mapbox.maps.plugin.Plugin.Companion.MAPBOX_LIFECYCLE_PLUGIN_ID
 import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
 
 /**
  * Concrete implementation of MapboxLifecyclePlugin.
  */
-class MapboxLifecyclePluginImpl : MapboxLifecyclePlugin {
+class MapboxLifecyclePluginImpl constructor(mapView: View?) : MapboxLifecyclePlugin,
+  LifecycleOwner {
```
yunikkk, line 23: A bit confusing that MapboxLifecyclePluginImpl is LifecycleOwner, returning instance of ViewLifecycleRegistry for the ViewLifecycleRegistry. Could ViewLifecycleRegistry implement LifecycleOwner itself instead? Code should become simpler then

yunikkk: E.g. 
```
internal class ViewLifecycleOwner(
  view: View,
  hostingLifecycleOwner: LifecycleOwner,
) : LifecycleOwner {

  private val lifecycleRegistry = LifecycleRegistry(this)

  override fun getLifecycle() = lifecycleRegistry
```

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleRegistry.kt
```diff
@@ -0,0 +1,57 @@
...
+    override fun onViewDetachedFromWindow(p0: View?) {
+      isAttached = false
+      if (hostingLifecycleOwner.lifecycle.currentState.isAtLeast(State.STARTED)) {
+        currentState = State.CREATED
+      }
+    }
+  }
+
+  init {
+    hostingLifecycleOwner.lifecycle.addObserver(hostingLifecycleObserver)
```
yunikkk, line 54: Should we detach them at some point?

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
pengdev: `ViewLifecycleRegistry` has to implement the `LifecycleRegistry`, in order to manipulate the lifecycle state, and merge the lifecycle with the view's detach/attach state.

pengdev: However pushed a commit to use a dummy `locaLifecycleOwner` instead of making `MapboxLifecyclePluginImpl` a `LifecycleOwner`

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
pengdev: yeah, this is a commit that prototypes Lukas's idea of initialising the viewLifecycleRegistry in constructor, however given the above conversation, we'd probably not initialising it in the constructor.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
yunikkk: > `ViewLifecycleRegistry` has to implement the `LifecycleRegistry`, in order to manipulate the lifecycle state, and merge the lifecycle with the view's detach/attach state.

Shouldn't be able to update the state `private val lifecycleRegistry = LifecycleRegistry(this)` should be used, so instead of `state = ` it becomes `lifecycleRegistry.state = `

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleRegistry.kt
yunikkk: Hmm, aren't they different objects actually?

yunikkk: Also, the same q applies to the `addOnAttachStateChangeListener` below

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
yunikkk: So it becomes : 
```
internal class ViewLifecycleOwner(
  view: View,
  hostingLifecycleOwner: LifecycleOwner,
) : LifecycleOwner {

  private val viewLifecycleRegistry = LifecycleRegistry(this)
  override fun getLifecycle() = viewLifecycleRegistry

  private var isAttached = view.isAttachedToWindow

  private val hostingLifecycleObserver = LifecycleEventObserver { _, event ->
    val isAtLeastCreated = currentState.isAtLeast(State.CREATED)
    if (isAttached || (isAtLeastCreated && event == Event.ON_DESTROY)) {
      viewLifecycleRegistry.handleLifecycleEvent(event)
    }
  }

  private val attachStateChangeListener = object : View.OnAttachStateChangeListener {
    override fun onViewAttachedToWindow(p0: View?) {
      viewLifecycleRegistry.currentState = hostingLifecycleOwner.lifecycle.currentState
      isAttached = true
    }

    override fun onViewDetachedFromWindow(p0: View?) {
      isAttached = false
      if (hostingLifecycleOwner.lifecycle.currentState.isAtLeast(State.STARTED)) {
        lifecycleRegistry.currentState = State.CREATED
      }
    }
  }

  init {
    hostingLifecycleOwner.lifecycle.addObserver(hostingLifecycleObserver)
    view.addOnAttachStateChangeListener(attachStateChangeListener)
  }
}
```

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleRegistry.kt
pengdev: oh, I misread the listener, yes, we should remove the observer/listener inside the ViewLifecycleRegistry

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
pengdev: pushed a commit with suggested changes.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
pengdev: As the above mentioned limitations, rolled back to `ViewTreeLifecycleOwner` instead of `context.toLifecyle`, and only construct the `ViewLifecycleRegistry` inside the `registerLifecycleObserver`, which is executed when the view is attached to window.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/MapboxLifecyclePluginImpl.kt
```diff
@@ -10,58 +10,53 @@ import android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE
...
   override fun registerLifecycleObserver(mapView: View, observer: MapboxLifecycleObserver) {
-    val lifecycleOwner = ViewTreeLifecycleOwner.get(mapView)
-    if (lifecycleOwner == null) {
-      logE(
-        TAG,
-        """Can't get lifecycleOwner for mapview,
-          please make sure the host Activity is AppCompatActivity and the version of appcompat is 1.3.0+.
-          If the host Activity is not AppCompatActivity,
-          you need manually invoke the corresponding lifecycle methods in onStart/onStop/onDestroy/onLowMemory methods of the host Activity"""
+    if (!this::viewLifecycleRegistry.isInitialized) {
```
yunikkk, line 32: Suppose `lateinit` is not needed and `val viewLifecycleRegistry: ViewLifecycleRegistry`  can be declared here now?

Chaoba (commented):
Need update the unit tests.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ContexExt.kt
```diff
@@ -0,0 +1,25 @@
...
+    logD("testtest", "recursiveUnwrap: $this")
+    this.baseContext.recursiveUnwrap()
+  } else {
+    this
+  }
+
+internal fun Context.toLifecycleOwner(): LifecycleOwner {
+  logD("testtest", "toLifecycleOwner: $this")
+  val lifecycleOwner = this.recursiveUnwrap() as? LifecycleOwner
+  checkNotNull(lifecycleOwner) {
```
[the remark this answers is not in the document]
pengdev: actually this method has flaws as discussed above, I'm now using ViewTreeLifecycleOwner instead, will remove this extension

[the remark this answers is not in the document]
pengdev: actually this method has flaws as discussed above, I'm now using ViewTreeLifecycleOwner instead, will remove this extension

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,95 @@
...
+        "Please ensure that the hosting activity/fragment is a valid LifecycleOwner"
+      )
+    viewLifecycleRegistry.currentState = hostingLifecycleOwner.lifecycle.currentState
+    hostingLifecycleOwner.lifecycle.addObserver(hostingLifecycleObserver)
+    this@ViewLifecycleOwner.hostingLifecycleOwner = hostingLifecycleOwner
+    isAttached = true
+  }
+
+  private fun doOnDetached() {
+    if (!isAttached) return
```
kiryldz, line 82: do we need some sort of warning here before return?

kiryldz (commented):
@pengdev side question - what particularly improves with better lifecycle management? can we describe it either in changelog or in PR description? otherwise at least for me it's not that clear what we are improving

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.Flow
+
+/**
+ * A [LifecycleOwner] that merges the hosting lifecycle owner's [Lifecycle] (of an Activity or Fragment)
+ * with the [View]'s drawing surface lifecycle.
+ *
+ * This lifecycle is useful for continuously running tasks that obtain the data and draw it in the [View],
+ * for example by collecting [Flow]s and drawing results in the nested views or on the map.
+ *
+ * In general, this lifecycle reflects the hosting lifecycle owner's [Lifecycle] whenever the `View` is attached
```
yunikkk, line 22: @pengdev this is what I've referred to, main thing fixed for Nav was the coroutines , e.g. with [repeatOnLifecycle](https://developer.android.com/topic/libraries/architecture/coroutines#restart).

I'd probably just highlighted that we keep MapView's lifecycle as CREATED when it is detached, this seem the most important thing for us.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 * Optimise the frequency to update location layer's visibility. ([1399](https://github.com/mapbox/mapbox-maps-android/pull/1399))
 * Fix the delay in the first appearance of the location puck. ([1403](https://github.com/mapbox/mapbox-maps-android/pull/1403))
+* Introduce internal `ViewLifecycleOwner` to have granular control over MapView's lifecycle. ([1330](https://github.com/mapbox/mapbox-maps-android/pull/1330))
```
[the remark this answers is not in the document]
pengdev: It's a bug, as we are not handling the lifecycle of the mapview properly. Also updated the description to showcase the issue.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
pengdev: will update the inline docs as well.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
pengdev: I don't think we need to warn anything, these are valid use cases, and we just need to handle them properly.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,95 @@
...
+
+  init {
+    view.addOnAttachStateChangeListener(attachStateChangeListener)
+    if (view.isAttachedToWindow) {
+      doOnAttached(view)
+    }
+  }
+
+  private fun doOnAttached(view: View) {
+    if (isAttached) return
```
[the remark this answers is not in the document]
pengdev: I don't think we need to warn anything, these are valid use cases, and we just need to handle them properly.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A [LifecycleOwner] that merges the hosting lifecycle owner's [Lifecycle] (of an Activity or Fragment)
+ * with the [View]'s drawing surface lifecycle.
+ *
+ * This lifecycle is useful for continuously running tasks that obtain the data and draw it in the [View],
+ * for example by collecting [Flow]s and drawing results in the nested views or on the map.
+ *
+ * In general, this lifecycle reflects the hosting lifecycle owner's [Lifecycle] whenever the `View` is attached
+ * but whenever the `View` is detached and we want to stop the coroutines tied to the lifecycle's scope,
+ * this lifecycle will transition to [Lifecycle.State.CREATED] (and invoke [Lifecycle.Event.ON_STOP])
```
yunikkk, line 24: Looked in the [onStop](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/MapController.kt#L144) method, couldn't we have an issue if view was detached, then onStyleLoaded event arrives and plugin dont receive the updated one since they've unsubscribed? 


```
    this.onStyleDataLoadedListener = OnStyleDataLoadedListener { eventData ->
      if (eventData.type == StyleDataType.STYLE) {
        mapboxMap.getStyle { style ->
          pluginRegistry.onStyleChanged(style)
        }
      }
    }
```

pengdev:
> @pengdev side question - what particularly improves with better lifecycle management? can we describe it either in changelog or in PR description? otherwise at least for me it's not that clear what we are improving

Updated the PR description to be more descriptive and the fixed use cases.

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
pengdev: 🤔 in this case, if the mapview is detached, do we still want to get the callbacks? I think all the plugin is aware of the onStart/onStop lifecycle events and should be prepared for re-initialisation during `onStart`?

pengdev: @yunikkk we don't know if the style loaded event is received before the MapView's onStop, so I'd rather prefer to initialise the plugin in onStart and do proper clean up in onStop. And whenever the style is needed to initialise the plugin, we should get the style using `MapboxMap.getStyle { style -> }`, wdyt?

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
pengdev: Updated, please take a look.

yunikkk: Seems fine to me now

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
pengdev: I was mainly concerned that triggering `onStyleChanged` manually on every `onStart` might result in adding the same layer/source to the same style repeatedly, and result in crash. As the documentation of the `onStyleChanged` suggest that it will only be called when a new Style is loaded, users may under the impression they need to redo all the style related changes.

Just checked code base and the plugins that extends `MapStyleObserverPlugin`, indeed it seems we are only updating the style reference but not triggering any other operations, so I guess it's safe to trigger `onStyleChanged` with our current code base, although it might still a bit confusing for the future plugin developers.

pengdev: Having said that, I think it's still a valid issue that we need to think more about, @yunikkk could you cut a new ticket for it and we can address it in a separate PR?

yunikkk (approved):

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
yunikkk: Added the ticket

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,91 @@
...
+      viewLifecycleRegistry.handleLifecycleEvent(event)
+    }
+  }
+
+  private val attachStateChangeListener = object : View.OnAttachStateChangeListener {
+    override fun onViewAttachedToWindow(p0: View) {
+      doOnAttached(p0)
+    }
+
+    override fun onViewDetachedFromWindow(p0: View) {
```
alexander-kulikovskii, line 50:
```suggestion
    override fun onViewDetachedFromWindow(view: View) {
```

CHANGELOG.md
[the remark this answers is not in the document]
kiryldz: just a small one: if it's a bug (and I agree it's a bug) - perhaps rephrase:
```suggestion
* Fix lifecycle edge cases not being handled properly by introducing internal `ViewLifecycleOwner` to have granular control over MapView's lifecycle. ([1330](https://github.com/mapbox/mapbox-maps-android/pull/1330))
```

kiryldz (approved):
LGTM with a nit

plugin-lifecycle/src/main/java/com/mapbox/maps/plugin/lifecycle/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  private fun doOnAttached(view: View) {
+    if (isAttached) return
+
+    hostingLifecycleOwner?.lifecycle?.removeObserver(hostingLifecycleObserver)
+
+    val hostingLifecycleOwner = ViewTreeLifecycleOwner.get(view)
+      ?: throw IllegalStateException(
```
alexander-kulikovskii, line 68: should we add one more test for such case?

pengdev: Added a test case

PR #1353, opened 2022-05-13
Add OnStylePreloaded parameter to loadStyle APIs.

pengdev:
### Summary of changes

Introduce earlyStyleCallback parameter to loadStyle APIs.

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Use a single shared buffer across all globe tiles to increase globe rendering performance. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * Re-introduce partial tile loading feature that decreases map load times. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * The `TilesetDescriptorOptions.pixelRatio` parameter is now passed to the TileStore and considered for the raster tile pack loading. This enables loading of a raster tilepacks for retina displays. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
+* Introduce earlyStyleCallback parameter to loadStyle APIs. ([#1353](https://github.com/mapbox/mapbox-maps-android/pull/1353))
 
 ## Bug fixes 🐞
 * Enable two finger pan gesture. ([#1280](https://github.com/mapbox/mapbox-maps-android/pull/1280))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Use a single shared buffer across all globe tiles to increase globe rendering performance. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * Re-introduce partial tile loading feature that decreases map load times. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * The `TilesetDescriptorOptions.pixelRatio` parameter is now passed to the TileStore and considered for the raster tile pack loading. This enables loading of a raster tilepacks for retina displays. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
-* Introduce earlyStyleCallback parameter to loadStyle APIs. ([#1353](https://github.com/mapbox/mapbox-maps-android/pull/1353))
+* Add optional `OnStylePreLoaded` parameter to loadStyle APIs. The `OnStylePreLoaded` is invoked when a style is parsed, so that the style layer properties could be read and modified, style layers and sources could be added or removed before rendering is started. ([#1353](https://github.com/mapbox/mapbox-maps-android/pull/1353))
 
 ## Bug fixes 🐞
 * Enable two finger pan gesture. ([#1280](https://github.com/mapbox/mapbox-maps-android/pull/1280))
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
@@ -35,13 +35,15 @@ class CircleLayerClusteringActivity : AppCompatActivity() {
 
     mapboxMap.loadStyleUri(
       styleUri = Style.LIGHT,
-      // Disable any type of fading transition when icons collide on the map. This enhances the visual
-      // look of the data clustering together and breaking apart.
-      styleTransitionOptions = TransitionOptions.Builder()
-        .duration(0)
-        .delay(0)
-        .enablePlacementTransitions(false)
-        .build(),
+      onStylePreLoaded = {
+        // Disable any type of fading transition when icons collide on the map. This enhances the visual
+        // look of the data clustering together and breaking apart.
+        it.styleTransition = TransitionOptions.Builder()
+          .duration(0)
+          .delay(0)
+          .enablePlacementTransitions(false)
+          .build()
+      },
       onStyleLoaded = {
         mapboxMap.flyTo(
           CameraOptions.Builder()
--- a/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
@@ -139,6 +139,54 @@ class MapboxMap :
    *
    * @param styleUri The style URI
    * @param styleTransitionOptions style transition options applied when loading the style
+   * @param earlyStyleCallback whether to invoke [onStyleLoaded] callback earlier on the `StyleDataLoaded#Style`]
+   *    event, default to false. Enable earlyStyleCallback helps with getting the Style Object faster,
+   *    so that the runtime styling can be applied immediately after the style is parsed.
+   * @param onStyleLoaded The OnStyleLoaded callback
+   * @param onMapLoadErrorListener The OnMapLoadErrorListener callback
+   */
+  fun loadStyleUri(
+    styleUri: String,
+    styleTransitionOptions: TransitionOptions? = null,
+    earlyStyleCallback: Boolean = false,
+    onStyleLoaded: Style.OnStyleLoaded? = null,
+    onMapLoadErrorListener: OnMapLoadErrorListener? = null
+  ) {
+    checkNativeMap("loadStyleUri")
+    initializeStyleLoad(earlyStyleCallback, onStyleLoaded, onMapLoadErrorListener, styleTransitionOptions)
+    if (styleUri.isEmpty()) {
+      nativeMap.styleJSON = EMPTY_STYLE_JSON
+    } else {
+      nativeMap.styleURI = styleUri
+    }
+  }
+
+  /**
+   * Will load a new map style asynchronous from the specified URI.
+   *
+   * URI can take the following forms:
+   *
+   * - **Constants**: load one of the bundled styles in [Style].
+   *
+   * - **`mapbox://styles/<user>/<style>`**:
+   * loads the style from a [Mapbox account](https://www.mapbox.com/account/).
+   * *user* is your username. *style* is the ID of your custom
+   * style created in [Mapbox Studio](https://www.mapbox.com/studio).
+   *
+   * - **`http://...` or `https://...`**:
+   * loads the style over the Internet from any web server.
+   *
+   * - **`asset://...`**:
+   * loads the style from the APK *assets* directory.
+   * This is used to load a style bundled with your app.
+   *
+   * - **`file://...`**:
+   * loads the style from a file path. This is used to load a style from disk.
+   *
+   * Will load an empty json `{}` if the styleUri is empty.
+   *
+   * @param styleUri The style URI
+   * @param styleTransitionOptions style transition options applied when loading the style
    * @param onStyleLoaded The OnStyleLoaded callback
    * @param onMapLoadErrorListener The OnMapLoadErrorListener callback
    */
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Use a single shared buffer across all globe tiles to increase globe rendering performance. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * Re-introduce partial tile loading feature that decreases map load times. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
 * The `TilesetDescriptorOptions.pixelRatio` parameter is now passed to the TileStore and considered for the raster tile pack loading. This enables loading of a raster tilepacks for retina displays. ([#1351](https://github.com/mapbox/mapbox-maps-android/pull/1351))
+* Introduce earlyStyleCallback parameter to loadStyle APIs. ([#1353](https://github.com/mapbox/mapbox-maps-android/pull/1353))
```
kiryldz, line 14: I wonder how we should clearly state to users what _could_ be run in earlyStyleCallback and what could not. For now early style callback will fit for projection and 3D modeling but will not work with runtime styling fully

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -139,17 +139,21 @@ class MapboxMap :
...
+   * @param earlyStyleCallback whether to invoke [onStyleLoaded] callback earlier on the `StyleDataLoaded#Style`]
+   *    event, default to false. Enable earlyStyleCallback helps with getting the Style Object faster,
+   *    so that the runtime styling can be applied immediately after the style is parsed.
    * @param onStyleLoaded The OnStyleLoaded callback
    * @param onMapLoadErrorListener The OnMapLoadErrorListener callback
    */
   fun loadStyleUri(
     styleUri: String,
     styleTransitionOptions: TransitionOptions? = null,
+    earlyStyleCallback: Boolean = false,
```
kiryldz, line 151: not sure `earlyStyleCallback` is good and descriptive naming but can't think of anything better for now..

CHANGELOG.md
pengdev: yeah, I've added detailed documentation to this flag below:
```
@param earlyStyleCallback whether to invoke [onStyleLoaded] callback earlier on the `StyleDataLoaded#Style`]
   *    event, default to false. Enable earlyStyleCallback helps with getting the Style Object faster,
   *    so that the runtime styling can be applied immediately after the style is parsed.
```
However, probably we should also state the limitations where this flag shouldn't be used, e.g. when the runtime styling is using the sources from the style url.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
pengdev: Any suggestion is welcome, that's the most relevant name I came up with so far..

kiryldz:
@pengdev another thought that came to me mind is that if we introduce this as a flag (like it' implemented right now) - user's existing code in `OnStyleLoaded` callback may be broken if I remember correctly from what you were saying some time ago as we're triggering that callback too early in that case.

Instead of extending bunch of `loadStyle` methods - WDYT about actually introducing new API: `MapboxMap.addOnStylePreLoaded` listener and clearly document it that this callback is triggered right in time _before_ `loadStyle` (and similar methods) is called and should be used to apply some style properties to work as expected (e.g. projection, fog, terrain, light).

Chaoba (commented):
Add some unit tests?

pengdev:
> @pengdev another thought that came to me mind is that if we introduce this as a flag (like it' implemented right now) - user's existing code in OnStyleLoaded callback may be broken if I remember correctly from what you were saying some time ago as we're triggering that callback too early in that case.

By default, user's code won't be broken, since the flag is default to false, so the original behaviour is still the default behaviour. And if user need to use early StyleLoaded callback, both adding the feature as a flag or a new `MapboxMap.addOnStylePreLoaded` will have issues if the user's runtime styling depends on the Sources in the style url, since the sources are not available yet. In this situation, user could have to split their runtime styling into two-parts and hook the Source related runtime styling inside the the StyleFullyLoaded event.

> 
> Instead of extending bunch of loadStyle methods - WDYT about actually introducing new API: MapboxMap.addOnStylePreLoaded listener and clearly document it that this callback is triggered right in time before loadStyle (and similar methods) is called and should be used to apply some style properties to work as expected (e.g. projection, fog, terrain, light).

Sounds interesting, `MapboxMap.addOnStylePreLoaded` listener is basically the same `MapboxMap.addOnStyleDataLoadedListener` with event data being style type, the issue is when should we create the Style instance, so in this case we need to create Style instance as early as possible and return the Style object inside the `addOnStylePreLoaded` callback I assume.

I do see the benefit that separating this callback to another API makes the concepts easier to understand, however if we are just talking about `clearly document it that this callback is triggered right in time before loadStyle (and similar methods) is called`, I guess similar documentation to the flag proposed in this PR also works, wdyt?

Another consideration is that with `MapboxMap.addOnStylePreLoaded`  approach, if the user want to use the early style loaded callback, instead using the closure of the `loadStyle` API, they have to to add this listener immediately after loadStyle APIs, and it doesn't work with Style DSLs, as the runtime styling part is done inside the original onStyleLoaded callback.

I think both approaches could work, but adding the parameter directly inside the loadStyle APIs seems more natural to use(using closure versus a separate API call following the loadStyle APIs) and also supports Style extension better. wdyt cc @kiryldz @mapbox/maps-android

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -92,6 +101,11 @@ internal class StyleObserver(
         loadStyleTransitionOptions = null
       }
     }
+    if (loadStyleEarlyStyleCallback) {
+      if (eventData.type == StyleDataType.STYLE) {
+        invokeStyleListeners()
+      }
+    }
```
alexander-kulikovskii, line 108:
```suggestion
    if (loadStyleEarlyStyleCallback && eventData.type == StyleDataType.STYLE) {
        invokeStyleListeners()
      }
    }
```

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
ank27: maybe `enableQuickStyleLoad` or `quickStyleLoad` 🤔

pengdev:
Discussed offline with @kiryldz  indeed it would cause less confusion if we expose a separate callback for `OnStylePreloaded`, as user can leave their current logic inside the OnStyleLoaded callback, but have the option to do some actions with the Style at an earlier point of time. But instead of having a separate `MapboxMap.addOnStylePreLoaded`, I made it a part of the `loadStyle` APIs. So when loading a new Style, user can choose to do some extra action in the OnStylePreloaded callback. Also updated the documentation to indicate when exactly these callbacks will be invoked, and what should be considered when using the `OnStylePreloaded` listener. @kiryldz please take a look at the changes and see if it matches your expectation?

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
pengdev: the `loadStyleEarlyStyleCallback` is removed so this change is not valid any more.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
pengdev: Please check https://github.com/mapbox/mapbox-maps-android/pull/1353#issuecomment-1127501925 , we have removed this boolean parameter and used a separate `OnStylePreloaded` listener instead. Please provide feedback on the new listener's name.

app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
```diff
@@ -35,13 +35,15 @@ class CircleLayerClusteringActivity : AppCompatActivity() {
...
-        .build(),
+      onStylePreLoaded = {
+        // Disable any type of fading transition when icons collide on the map. This enhances the visual
+        // look of the data clustering together and breaking apart.
+        it.styleTransition = TransitionOptions.Builder()
+          .duration(0)
+          .delay(0)
+          .enablePlacementTransitions(false)
+          .build()
+      },
```
kiryldz, line 46: why did we move setting style transition to this block? it work fine when specifying `styleTransitionOptions` parameter directly, that's why we did introduce that parameter

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -139,17 +139,19 @@ class MapboxMap :
    *
    * @param styleUri The style URI
    * @param styleTransitionOptions style transition options applied when loading the style
+   * @param onStylePreLoaded The OnStylePreloaded callback
```
yunikkk, line 142: Imho might be dangerous. Users can (and some will definitely, I suppose) accidentally use `onStylePreLoaded` and treat style as loaded, using APIs not allowed yet

pengdev: There's no APIs that's not allowed, it's just some use cases, like modifying a source property that has not been loaded will cause some issues.

app/src/main/java/com/mapbox/maps/testapp/examples/CircleLayerClusteringActivity.kt
pengdev: Both the original  `MapboxMap.styleTransitionOptions` and the newly introduced `OnStylePreloaded` will work in this case, because under the hood, they have exactly exactly the same behaviour. Modified here to demonstrate this specific use case.

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
pengdev: And these dangerous use cases is clearly documented in the `onStylePreLoaded` listener, so users should use this API at their own risk.

yunikkk: I mean - it's easy to mistakenly use `map.loadStyle(onStylePreloaded = {})` instead of `map.loadStyle(onStyleLoaded = {})`, calls look similarly enough too confuse them writing manually or with autocomplete

yunikkk: > There's no APIs that's not allowed, it's just some use cases, like modifying a source property that has not been loaded will cause some issues.

IMHO it's close to `not allowed`, if the api call can cause issues, even if this doesn't mean crash =)

pengdev: > I mean - it's easy to mistakenly use map.loadStyle(onStylePreloaded = {}) instead of map.loadStyle(onStyleLoaded = {}), calls look similarly enough too confuse them writing manually or with autocomplete

Do you have suggestion on the naming so that it will become less confusing?

> IMHO it's close to not allowed, if the api call can cause issues, even if this doesn't mean crash =)

The misuse will most likely cause errors like `Source with Source id is not found in the style` I assume 😄

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -60,18 +63,29 @@ internal class StyleObserver(
...
+      }
+      getStyleListeners.clear()
     }
-    getStyleListeners.clear()
+  }
+
+  private fun invokeStylePreloadedListeners() {
+    val style = Style(nativeMap, pixelRatio)
+    loadedStyle?.markInvalid()
+    loadedStyle = style
```
kiryldz, line 86: shouldn't that logically be located in `invokeStyleLoadedListeners()`? although that events (style preloaded and style loaded) are close enough - guess it makes sense to mark previous style invalid and store new style in a var in `invokeStyleLoadedListeners()`

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -92,6 +106,9 @@ internal class StyleObserver(
         loadStyleTransitionOptions = null
       }
     }
+    if (eventData.type == StyleDataType.STYLE) {
```
kiryldz, line 109: combine into one `if` with line 103 for better readability

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -1037,14 +1037,30 @@ class Style internal constructor(
...
      */
     fun onStyleLoaded(style: Style)
   }
+
+  /**
+   * Callback to be invoked when a style is parsed, style layer properties could be read and modified, style layers and sources could be
+   * added or removed before rendering is started.
+   *
+   * Note that the style sources has not been loaded at this point, be careful not to use runtime styling to modify the style source properties.
+   */
```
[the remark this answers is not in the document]
pengdev: Would like 👀 from @alexshalamov on the documentation here as well.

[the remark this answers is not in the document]
alexshalamov: _style layer properties could be read and modified, style layers and sources could be added or removed before rendering is started_

In 99% of cases, yes, that will work. Should we explain edge cases? Like, source cannot be removed if it is used by layers? Also, if images are added at this point, they will be removed when a style sprite sheet is loaded.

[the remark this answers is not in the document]
pengdev: > Should we explain edge cases? Like, source cannot be removed if it is used by layers? Also, if images are added at this point, they will be removed when a style sprite sheet is loaded.

@alexshalamov Yes, I do think we should try to cover as much corner cases as possible in this callback, and also mentioned the typical use cases we have currently(e.g. set projection, terrain, style transition etc.)

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
pengdev: This is the place we save the reference of the newly created Style object inside the preload phase, and we are returning the same object inside the `OnStyleLoaded` callback. 

Imo we should mark the previous style invalid at this point of time, as the new Style object is already created, previous style url is already overwritten.

pengdev:
Closing this PR for now in favour of improve our Style DSL and hook the runtime styling inside the DSL to the appropriate and earliest MapEvents.

PR #1355, opened 2022-05-16
Get the correct last stable release name for revapi

Chaoba:
### Summary of changes
The last stable version number that gets from github is not correct if a patch(10.4.4) released is released after a stable release(10.5.0).
Instead of getting it from github, we now read `LAST_STABLE_VERSION_NAME` from gradle.properties which will be updated for each minor update.

## Pull request checklist:
    
### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,5 @@
 VERSION_NAME=10.5.0-SNAPSHOT
+LAST_STABLE_VERSION_NAME=10.5.0
 
 # Project-wide Gradle settings.
 # IDE (e.g. Android Studio) users:
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,4 @@
 VERSION_NAME=10.5.0-SNAPSHOT
-LAST_STABLE_VERSION_NAME=10.5.0
 
 # Project-wide Gradle settings.
 # IDE (e.g. Android Studio) users:
--- a/scripts/java-api-check-all.sh
+++ b/scripts/java-api-check-all.sh
@@ -6,6 +6,13 @@ set -Eeuo pipefail
 #
 # Note: if run locally, execute `mbx env` and `./gradlew assembleRelease` first.
 #
+if ! command -v gh &> /dev/null
+then
+    echo "gh (github cli tool) is not found, install it first"
+    exit
+fi
+gh auth login --with-token < ./gh_token.txt
+
 CURRENT_DIR=$(dirname "$0")
 MAJOR_CHANGE_FILE=${CURRENT_DIR}/../api_compat_report/major.txt
 [ -e $MAJOR_CHANGE_FILE ] && rm $MAJOR_CHANGE_FILE
@@ -25,8 +32,28 @@ if ! [ -x "$(command -v revapi.sh)" ]; then
     popd > /dev/null
 fi
 
-RELEATE_TAG=${1-""}
-echo "Release tag: $RELEATE_TAG"
+RELEASE_TAG=${1-""}
+
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+TARGET_BRANCH=$(gh pr view "$CURRENT_BRANCH" --json baseRefName --jq '.baseRefName')
+echo "CURRENT_BRANCH: $CURRENT_BRANCH"
+echo "TARGET_BRANCH: $TARGET_BRANCH"
+
+if ! [ "$TARGET_BRANCH" = "main" ]; then # Release branches
+  LATEST_RELEASE_TAG_IN_THIS_BRANCH=$(git tag --merged "$CURRENT_BRANCH" --sort=-creatordate | grep "android-v"| head -n 1)
+  if [ "$LATEST_RELEASE_TAG_IN_THIS_BRANCH" = "*beta*" ] || [ "$LATEST_RELEASE_TAG_IN_THIS_BRANCH" = "*rc*" ]; then
+    # Use the latest stable minor release tag during rc and beta release
+    LAST_STABLE_RELEASE_TAG=$(git tag --list 'android-v*.0' --sort=-creatordate | head -n 1)
+  else
+    # Use the latest stable minor release tag in this branch
+    LAST_STABLE_RELEASE_TAG=$(git tag --merged "$CURRENT_BRANCH" --sort=-creatordate | grep "android-v.*0$"| head -n 1)
+  fi
+else
+  # Use the latest stable minor release tag for main branch
+  LAST_STABLE_RELEASE_TAG=$(git tag --list 'android-v*.0' --sort=-creatordate | head -n 1)
+fi
+echo "LAST_STABLE_RELEASE_TAG: $LAST_STABLE_RELEASE_TAG"
+echo "Release tag: $RELEASE_TAG"
 MODULES=$(sed -e '/include(/,/)/!d' ./settings.gradle.kts | tr -d '\n' | sed 's/.*(\([^]]*\)).*/\1/g;s/,\n/\n/g' | tr , \\n | sed 's/.*":\([^]]*\)".*/\1/g')
 for i in $MODULES; do
   if [ -f "./$i/build/outputs/aar/$i-release.aar" ]; then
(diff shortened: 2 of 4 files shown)


scripts/java-api-check.sh
```diff
@@ -50,10 +44,7 @@ PREVIOUS_RELEASE=${TMPDIR}/previous/sdk-release.aar
 CURRENT_RELEASE_DIR=$(dirname "${CURRENT_RELEASE}")
 PREVIOUS_RELEASE_DIR=$(dirname "${PREVIOUS_RELEASE}")
 
-gh auth login --with-token < ./gh_token.txt
-LAST_VERSION_TAG=$(gh release list -L 1) #android-v10.3.0
-LAST_VERSION_TAG_ARRAY=($LAST_VERSION_TAG)
-LAST_VERSION=${LAST_VERSION_TAG_ARRAY[0]:9}
+LAST_VERSION=$(grep -o '[0-9]*\.[0-9]*\.[0-9]*' ./gradle.properties) #10.3.0
```
alexander-kulikovskii, line 47: will be nice to check only `VERSION_NAME=PATTERN`. because we can store smth else in this file.

ank27:
I think we should create a new variable in gradle.properties with something like `LAST_STABLE_VERSION`, so it doesn't conflict with `VERSION_NAME` that we currently use for versioning. Also, we should update the release template too.

Chaoba:
> I think we should create a new variable in gradle.properties with something like `LAST_STABLE_VERSION`, so it doesn't conflict with `VERSION_NAME` that we currently use for versioning. Also, we should update the release template too.

@ank27 We update this property for each stable release except the patch releases(10.3.0/10.4.0...) and that is exactly what we need for api check, so no need for an additional variable from my perspective.

scripts/java-api-check.sh
Chaoba: Updated.

yunikkk:
Cant we just look for the latest release ending with `.0` ( == latest minor release) to get `10.5.0` and not `10.4.4` from github instead of parsing properties?

Chaoba:
> Cant we just look for the latest release ending with `.0` ( == latest minor release) to get `10.5.0` and not `10.4.4` from github instead of parsing properties?

The version number will be not correct for release branches, for example for branch 10.4, it will choose 10.5.0 as the base to compare. Read from gradle.properties is the correct one for every branch.

scripts/java-api-check.sh
```diff
@@ -50,10 +44,9 @@ PREVIOUS_RELEASE=${TMPDIR}/previous/sdk-release.aar
 CURRENT_RELEASE_DIR=$(dirname "${CURRENT_RELEASE}")
 PREVIOUS_RELEASE_DIR=$(dirname "${PREVIOUS_RELEASE}")
 
-gh auth login --with-token < ./gh_token.txt
-LAST_VERSION_TAG=$(gh release list -L 1) #android-v10.3.0
-LAST_VERSION_TAG_ARRAY=($LAST_VERSION_TAG)
-LAST_VERSION=${LAST_VERSION_TAG_ARRAY[0]:9}
+LAST_VERSION_TAG=$(grep -o 'LAST_STABLE_VERSION_NAME=[0-9]*\.[0-9]*\.[0-9]*' ./gradle.properties) #10.5.0
+LAST_VERSION_ARRAY=(${LAST_VERSION_TAG//=/ })
+LAST_VERSION=${LAST_VERSION_ARRAY[1]}
```
alexander-kulikovskii, line 49:
```suggestion
LAST_STABLE_VERSION_TAG=$(grep -o 'LAST_STABLE_VERSION_NAME=[0-9]*\.[0-9]*\.[0-9]*' ./gradle.properties) #10.5.0
LAST_STABLE_VERSION_ARRAY=(${LAST_STABLE_VERSION_TAG//=/ })
LAST_STABLE_VERSION=${LAST_STABLE_VERSION_ARRAY[1]}
```

gradle.properties
```diff
@@ -1,4 +1,5 @@
 VERSION_NAME=10.5.0-SNAPSHOT
+LAST_STABLE_VERSION_NAME=10.5.0
```
kiryldz, line 2: link internal PR with instructions to update that value during the release? or could we even automate setting this property during the release idk

scripts/java-api-check.sh
```diff
@@ -50,10 +44,9 @@ PREVIOUS_RELEASE=${TMPDIR}/previous/sdk-release.aar
 CURRENT_RELEASE_DIR=$(dirname "${CURRENT_RELEASE}")
 PREVIOUS_RELEASE_DIR=$(dirname "${PREVIOUS_RELEASE}")
 
-gh auth login --with-token < ./gh_token.txt
-LAST_VERSION_TAG=$(gh release list -L 1) #android-v10.3.0
-LAST_VERSION_TAG_ARRAY=($LAST_VERSION_TAG)
-LAST_VERSION=${LAST_VERSION_TAG_ARRAY[0]:9}
+LAST_VERSION_TAG=$(grep -o 'LAST_STABLE_VERSION_NAME=[0-9]*\.[0-9]*\.[0-9]*' ./gradle.properties) #10.5.0
```
yunikkk, line 47: Thinking about using `gh list releases` instead of adding new property that needs to be managed manually? Can we just fix the way `gh` finds latest release? E.g. detect what branch is checked, then filter `gh list releases` differently?

scripts/java-api-check-all.sh
```diff
@@ -25,8 +32,28 @@ if ! [ -x "$(command -v revapi.sh)" ]; then
...
-echo "Release tag: $RELEATE_TAG"
+RELEASE_TAG=${1-""}
+
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+TARGET_BRANCH=$(gh pr view "$CURRENT_BRANCH" --json baseRefName --jq '.baseRefName')
+echo "CURRENT_BRANCH: $CURRENT_BRANCH"
+echo "TARGET_BRANCH: $TARGET_BRANCH"
+
+if ! [ "$TARGET_BRANCH" = "main" ]; then # Release branches
+  LATEST_RELEASE_TAG_IN_THIS_BRANCH=$(git tag --merged "$CURRENT_BRANCH" --sort=-creatordate | grep "android-v"| head -n 1)
```
yunikkk, line 43: Is the name of the target branch here already sufficient? E.g. `v10.4`

Chaoba: No, only with target branch we don't know whether it's a beta/rc or stable release.

yunikkk: Ah, so PRs to 10.4.0-beta / 10.4.0-rc we will compare with 10.3, and when 10.4.0 is released and we're working on 10.4.1 - we'll compare with 10.4.0, is that right?

Chaoba: That's right.

ank27: > Ah, so PRs to 10.4.0-beta / 10.4.0-rc we will compare with 10.3, and when 10.4.0 is released and we're working on 10.4.1 - we'll compare with 10.4.0, is that right?

I think it should compare with `10.3` even when we're working on 10.4.1. for example, anyone upgrading from `10.3` to latest stable `10.4.1`, their java code shouldn't break 🤔, any thoughts?

Chaoba: The reason why use the last stable release is to exclude any changes(maybe not breaking changes) that could fail the ci job. If 10.3 to 10.4 has the breaking change, it should be found before release 10.4, for  10.4.1 we just need to make sure no breaking change for this patch release.

yunikkk: @ank27 suppose in this case we'll have breaking change in 10.3-10.4, right? And if both updates 10.3-10.4 / 10.4-10.4.1 has no breaking changes then it means 10.3-10.4.1 is also safe.
At the same time if there is something breaking between 10.3-10.4 and we decide we allow such change comparing 10.4.1 with 10.3 will force us to review those breaking changes from the past once again

ank27: correct 👍

yunikkk (approved):

scripts/java-api-check-all.sh
```diff
@@ -25,8 +32,28 @@ if ! [ -x "$(command -v revapi.sh)" ]; then
     popd > /dev/null
 fi
 
-RELEATE_TAG=${1-""}
-echo "Release tag: $RELEATE_TAG"
+RELEASE_TAG=${1-""}
```
ank27, line 35: we could move this code to extract last_version_tag to another script so this can be reused. but probably later.

ank27 (approved):

PR #1364, opened 2022-05-18
Fix crash due to invalid focal point when panning the map

kiryldz:
### Summary of changes

Fix crash due to invalid focal point when panning the map by skipping camera movement logic in case focal point returned from `MoveGestureDetector` is invalid meaning that X or Y are non-finite values.

## Pull request checklist:
    
### DIFF
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
@@ -19,6 +19,7 @@ import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.ScreenCoordinate
 import com.mapbox.maps.StylePropertyValueKind
 import com.mapbox.maps.extension.style.StyleInterface
+import com.mapbox.maps.logE
 import com.mapbox.maps.plugin.InvalidPluginConfigurationException
 import com.mapbox.maps.plugin.MapStyleObserverPlugin
 import com.mapbox.maps.plugin.Plugin.Companion.MAPBOX_CAMERA_PLUGIN_ID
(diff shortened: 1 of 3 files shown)


plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -423,6 +423,25 @@ class GesturesPluginTest {
...
+  @Test
+  fun verifyHandleMoveExceptionFreeForInvalidFocalPoint() {
+    mockkStatic("com.mapbox.maps.MapboxLogger")
+    every { logE(any(), any()) } just Runs
+    val moveGestureDetector = mockk<MoveGestureDetector>()
+    every { moveGestureDetector.pointersCount } returns 1
+    every {
+      moveGestureDetector.focalPoint
+    } returns PointF(Float.NaN, Float.NaN)
+    var handled = presenter.handleMove(moveGestureDetector, 50.0f, 50.0f)
```
alexander-kulikovskii, line 435: nit: seems to be as several parameterized tests.
and should you verify `logE(...)` was called?

alexander-kulikovskii (approved):

yunikkk (approved):

ank27 (approved):

tobrun (approved):

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1387,6 +1388,12 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
       val fromX = focalPoint.x.toDouble()
       val fromY = focalPoint.y.toDouble()
 
+      // Skip invalid focal points with non-finite values
+      if (!focalPoint.x.isFinite() || !focalPoint.y.isFinite()) {
```
pengdev, line 1392: looks good for a temporary fix, but we should also investigate the root cause, seems the focal point is coming from the MultiFingerGesture class in MapboxGestures library, would be good to fix it there, as the focal point is also used in several other places in the gestures plugin.

pengdev (approved):

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
kiryldz: imo a bit too much for param tests here but added verify for calling logger, thanks!

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
kiryldz: we don't have reproducible use-case sadly so let's fix precisely an issue for now and indeed take a deeper look later if needed

PR #1357, opened 2022-05-16
[provider] rework map providers

tobrun:
This PR rewrites our map provider implementation.

PR #1366, opened 2022-05-18
[provider] rework map providers

tobrun:
Tailwork of #1357, a distinction needs to be made between the NativeMapImpl which as a wrapper around the native map object vs the actual native map object. This went unnoticed since they both comply to the MapInterface which is used as its return type.


pengdev (approved):

heangpov16 (commented):
Leave comment

PR #1377, opened 2022-05-20
Exposes a public function for the layer style json

enyciaa:
### Summary of changes
Exposes a public api on a Layer of `getLayerJson` which returns the style json as a string

### User impact (optional)
In our project we write style DSL on android, extract the layer JSON and share the layer JSON between multiple platforms, which means we only need to create the layer styles once. This is quite useful as some of our styling is quite complex, so sharing one json file reduces bugs between platforms.

Currently there's no public api for grabbing a Layer style JSON. So to extract a layer style JSON we breakpoint inside `Layer` and run the line of code in this PR.

And then create layers using the JSON like below.
```
this.addStyleLayer(
    Value.fromJson(myJson).value!!,
    LayerPosition(null, null, null) 
)
```

Adding a public api would make this process easier, and possibly let us automate JSON extraction

## Pull request checklist:
    
### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
@@ -119,6 +119,10 @@ abstract class Layer : StyleContract.StyleLayerExtension {
 
   // Layer Properties
 
+  fun getLayerJson(): String {
+    return getCachedLayerProperties().toJson()
+  }
+
   internal fun getCachedLayerProperties(): Value {
     val properties = HashMap<String, Value>()
     layerProperties.values.forEach {


enyciaa:
If you're happy with this addition in this format, I'll plug through the PR checklist to get it ready to merge

pengdev:
Hi @enyciaa thanks for the contribution, we do have a public [getStyleLayerProperties](https://docs.mapbox.com/android/maps/api/10.5.0/mapbox-maps-android/com.mapbox.maps/-style/get-style-layer-properties.html) API to get the style layer properties from the Style object. It can be acquired after the layer is added to the style, is this API helpful in your use case?

enyciaa:
That would certainly work! The issue I see with that method is it's a super obscure way to grab a layers json without having an in depth knowledge of the API. For an sdk user to know to use that they'd also have to know:
- To look in an overall style to access inner properties of a nested layer
- That layer properties = layer json
- That `Value` is a Mapbox sdk parsed version of the json
- That it's possible to convert a `Value` to a json

Adding a direct method to the Layer class would certainly simplify from a sdk consumer perspective

pengdev:
Thanks for the reasoning, these are good points, these APIs are kind of more hidden from user's perspective and requires in depth knowledge of the SDK, we are keep iterating on the API documentation and guides to make these APIs more discoverable.

We didn't expose these properties to the user directly under Layer because the properties we stored in the Layer class are flattened and doesn't reflected structure compared with the gl-js and style specification, where the layer properties are grouped into [layout properties and paint properties](https://docs.mapbox.com/help/glossary/layout-paint-property/).

The [getStyleLayerProperties](https://docs.mapbox.com/android/maps/api/10.5.0/mapbox-maps-android/com.mapbox.maps/-style/get-style-layer-properties.html) API on the other hand, will return a JSON that's fully compatible with gl-js and style specification. It's a more accurate representation of the layer JSON, and is the suggested API to fetch such kind of information, and we do provide an extension to the returned Value class to convert it to a JSON string `Value.toJson()`.

Issue #1371, opened 2022-05-19
MapboxMap.loadStyle fix.


yunikkk:
Followup for the #1362. 

Order of StyleDataType.SPRITES and StyleDataType.SOURCES is not defined, moreover both events are optional and does not arrive if style does not contain corresponding data.
To make sure every style extension component is loaded we need to check if SPRITES/SOURCES events arrived in the onStyleLoaded callback and load if not.

PR #1362, opened 2022-05-17
Rework MapboxMap.loadStyle

yunikkk:
Another approach to fix #1353 - apply styleExtension properties as soon as the corresponding style data event arrives.

With this change order of styleExtension properties applied is as follows :
1. fog, light, terrain, projection - after StyleDataType.STYLE
2. images - after StyleDataType.SPRITE
3. sources and layers - after StyleDataType.SOURCES

Note that order of StyleDataType.SPRITE and StyleDataType.SOURCES events is _NOT_ guaranteed and in fact often arrives differently. However this does not seem to be a problem (discussed with @alexshalamov, but please confirm). 

Tested briefly - everything seems working, however since change can be very impactful will require more thorough manual / unit tests.

## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. If tests were not written, please explain why.
 - [ ] Optimize code for java consumption (`@JvmOverloads`, `@file:JvmName`, etc).
 - [ ] Add example if relevant.
 - [ ] Document any changes to public APIs.
 - [ ] Run `make update-api` to update generated api files, if there's public API changes, otherwise the `verify-api-*` CI steps might fail.
 - [ ] Update [CHANGELOG.md](../CHANGELOG.md) or use the label 'skip changelog', otherwise `check changelog` CI step will fail.
 - [ ] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.[version]` release branch.

Fixes: #1127

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-android/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1378, opened 2022-05-24
Load sprites and sources style extensions earlier during onStyleLoad.

yunikkk:
Fire sprite/sources loaded MapboxMap callbacks on onStyleLoad if no style data events arrived. 

## Pull request checklist:
    
Fixes: #1371 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Features ✨ and improvements 🏁
 * Update SDK name in attribution action sheet. ([1375](https://github.com/mapbox/mapbox-maps-android/pull/1375))
 * Introduce experimental ModelLayer API to render 3D models on the map. ([#1369](https://github.com/mapbox/mapbox-maps-android/pull/1369))
+* Further optimize `MapboxMap.loadStyle()` to apply styling properties earlier. [#1378](https://github.com/mapbox/mapbox-maps-android/pull/1378)
 
 # 10.6.0-beta.1 May 19, 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Features ✨ and improvements 🏁
 * Optimise the bearing update frequency for the location puck animator. ([1398](https://github.com/mapbox/mapbox-maps-android/pull/1398))
 * Use `orientation` model source property to update the 3D puck's bearing, as it is more efficient than updating the `model-rotation` layer property. ([1407](https://github.com/mapbox/mapbox-maps-android/pull/1407))
+* Optimize `MapboxMap.loadStyle()` to apply images and models earlier. [#1378](https://github.com/mapbox/mapbox-maps-android/pull/1378)
 
 ## Bug fixes 🐞
 * Optimise the frequency to update location layer's visibility. ([1399](https://github.com/mapbox/mapbox-maps-android/pull/1399))
@@ -45,7 +46,6 @@ Mapbox welcomes participation and contributions from everyone.
 ## Features ✨ and improvements 🏁
 * Update SDK name in attribution action sheet. ([1375](https://github.com/mapbox/mapbox-maps-android/pull/1375))
 * Introduce experimental ModelLayer API to render 3D models on the map. ([#1369](https://github.com/mapbox/mapbox-maps-android/pull/1369))
-* Further optimize `MapboxMap.loadStyle()` to apply styling properties earlier. [#1378](https://github.com/mapbox/mapbox-maps-android/pull/1378)
 
 # 10.6.0-beta.1 May 19, 2022
 
--- a/sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
@@ -9,8 +9,6 @@ import junit.framework.TestCase.*
 import org.junit.*
 import org.junit.runner.RunWith
 import java.util.concurrent.CountDownLatch
-import java.util.concurrent.TimeUnit
-import java.util.concurrent.TimeoutException
 
 @RunWith(AndroidJUnit4::class)
 @LargeTest
@@ -76,6 +74,7 @@ class StyleLoadTest {
 
     mapboxMap.onDestroy()
 
+    assertFalse(styles[0].isValid())
     assertFalse(styles[1].isValid())
   }
 
@@ -109,8 +108,11 @@ class StyleLoadTest {
           Style.MAPBOX_STREETS
         ) { style ->
           assertTrue("Style should be fully loaded", style.isStyleLoaded)
-          mapboxMap.loadStyleUri(Style.SATELLITE) {
-            assertTrue("Style should be fully loaded again", style.isStyleLoaded)
+          mapboxMap.loadStyleUri(Style.SATELLITE) { style2 ->
+            assertTrue("Style should be still fully loaded although its flaw in our current implementation", style.isStyleLoaded)
+            assertTrue("Style should be fully loaded again", style2.isStyleLoaded)
+            assertTrue("New style should be valid", style2.isValid())
+            assertFalse("Old style should not be valid", style.isValid())
             countDownLatch.countDown()
           }
           assertFalse("Map shouldn't be fully loaded", style.isStyleLoaded)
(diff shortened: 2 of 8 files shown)


sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -69,6 +69,19 @@ internal class StyleObserver(
    */
   override fun onStyleLoaded(eventData: StyleLoadedEventData) {
     val style = loadedStyle ?: throw MapboxMapException("Style is not initialized on onStyleLoaded callback!")
+
+    // sprites listener is not null - it means style does not contain them thus StyleDataType.SPRITE
+    // was not fired, explicitly call it to bind all the necessary style extensions upstream
+    if (styleDataSpritesLoadedListener != null) {
+      styleDataSpritesLoadedListener?.onStyleLoaded(style)
+      styleDataSpritesLoadedListener = null
```
kiryldz, line 77: we should explicitly set to null before `styleDataSpritesLoadedListener?.onStyleLoaded(style)` otherwise we might have an issue if `styleDataSpritesLoadedListener?.onStyleLoaded(style)` will trigger any update in the same callchain

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -69,6 +69,19 @@ internal class StyleObserver(
...
+    }
+    // sources listener is not null - it means style does not contain them thus StyleDataType.SOURCES
+    // was not fired, explicitly call it to bind all the necessary style extensions upstream
+    if (styleDataSourcesLoadedListener != null) {
+      styleDataSourcesLoadedListener?.onStyleLoaded(style)
+      styleDataSourcesLoadedListener = null
```
kiryldz, line 83: same as https://github.com/mapbox/mapbox-maps-android/pull/1378/files#r880446875

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -98,16 +111,22 @@ internal class StyleObserver(
...
-        styleDataSpritesLoadedListener?.onStyleLoaded(
-          loadedStyle ?: throw MapboxMapException("Style is not initialized yet although SPRITES event has arrived!")
-        )
-        styleDataSpritesLoadedListener = null
+        styleDataSpritesLoadedListener?.let {
+          styleDataSpritesLoadedListener = null
+          it.onStyleLoaded(
+            loadedStyle ?: throw MapboxMapException("Style is not initialized yet although SPRITES event has arrived!")
+          )
+        }
```
kiryldz, line 119: firstly, we could try make `styleDataSpritesLoadedListener` non-nullable and introduce meaningful bool flag;
secondly, this code seems to duplicate code from https://github.com/mapbox/mapbox-maps-android/pull/1378/files#r880446875 - I think it makes sense to introduce function and re-use it

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -98,16 +111,22 @@ internal class StyleObserver(
...
-        )
-        styleDataSourcesLoadedListener = null
+        styleDataSourcesLoadedListener?.let {
+          // reset listener first, firing onStyleLoaded
+          // may produce another StyleDataType.SOURCES event if sources are added
+          styleDataSourcesLoadedListener = null
+          it.onStyleLoaded(
+            loadedStyle ?: throw MapboxMapException("Style is not initialized yet although SOURCES event has arrived!")
+          )
+        }
```
kiryldz, line 129: same as https://github.com/mapbox/mapbox-maps-android/pull/1378/files#r880451859

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
yunikkk: Talked offline - will extract logic with callback call to the function instead

kiryldz (commented):
Rebase needed and also making sure that models introduced are put in the correct callback. We also need new tests for that. cc @pengdev

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -98,20 +101,38 @@ internal class StyleObserver(
...
 
+  private fun onStyleSourcesReady() {
+    // sources listener is not null - it means style does not contain them thus StyleDataType.SOURCES
+    // was not fired, explicitly call it to bind all the necessary style extensions upstream
+    styleDataSourcesLoadedListener?.let {
+      // reset listener first, firing onStyleLoaded
+      // may produce another StyleDataType.SOURCES event if sources are added
+      styleDataSourcesLoadedListener = null
+      it.onStyleLoaded(
+        loadedStyle ?: throw MapboxMapException("Style is not initialized yet although SOURCES event has arrived!")
```
Chaoba, line 120: Can you verify whether this works as expected if switch two styles very quickly?

yunikkk: I've tested it in the app loading 10+ styles, but let me try to do the instrumentation test

kiryldz (approved):
LGTM but in this PR (or as a follow-up item) I think it may be decent idea to add couple of instrument tests to `sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt`.

pengdev:
Would be good to have @alexshalamov 's approval here

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -299,19 +284,21 @@ class MapboxMap :
...
-        // TODO https://github.com/mapbox/mapbox-maps-android/issues/1371
-//        styleExtension.sources.forEach {
-//          it.bindTo(style)
-//        }
-//        styleExtension.layers.forEach { (layer, layerPosition) ->
-//          layer.bindTo(style, layerPosition)
-//        }
+        styleExtension.sources.forEach {
+          it.bindTo(style)
+        }
```
alexshalamov, line 289: what would happen if source already exist? Would we skip adding it and continue with adding layers? Or we just throw exception?

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
```diff
@@ -299,19 +284,21 @@ class MapboxMap :
...
+          layer.bindTo(style, layerPosition)
+        }
       },
       styleDataSpritesLoadedListener = { style ->
-        // TODO https://github.com/mapbox/mapbox-maps-android/issues/1371
-//        styleExtension.images.forEach {
-//          it.bindTo(style)
-//        }
+        styleExtension.images.forEach {
+          it.bindTo(style)
```
alexshalamov, line 296: same question, could addition of an image affect addition of a models?

alexshalamov (approved):

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: It will throw an exception, thus we do only call this once after SOURCES event, or, if it's missing - after full onStyleLoaded callback

sdk/src/main/java/com/mapbox/maps/MapboxMap.kt
yunikkk: Hmm, not sure - how could image affect models here? Do you mean loading images several times may throw an exception?

alexshalamov: I'm not sure what is behind ` it.bindTo(style)`, but probably exception would be thrown if there is an error, then models would not load, similarly to https://github.com/mapbox/mapbox-maps-android/pull/1378#discussion_r881411322

yunikkk: Actually it would crash the app instead since the exception is thrown on the main thread. Models would not load but that's would not  be a problem anymore =)

yunikkk: If an error is returned - exception will be thrown here https://github.com/mapbox/mapbox-maps-android/blob/main/extension-style/src/main/java/com/mapbox/maps/extension/style/image/ImageExtensionImpl.kt#L36 and the app will crash

ank27 (approved):

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
```diff
@@ -35,94 +36,114 @@ class StyleLoadTest {
...
+      it.runOnUiThread {
+        mapboxMap.loadStyleUri(Style.MAPBOX_STREETS)
+        mapboxMap.loadStyle(style("") {})
+        mapboxMap.loadStyleUri(Style.SATELLITE)
+        mapboxMap.loadStyleUri(Style.MAPBOX_STREETS)
+        mapboxMap.loadStyle(style("") {})
+        mapboxMap.loadStyleUri(Style.MAPBOX_STREETS)
+        mapboxMap.loadStyleUri(Style.SATELLITE)
+        mapboxMap.loadStyleUri(Style.MAPBOX_STREETS) { style ->
+          assertTrue("Style should be fully loaded", style.isStyleLoaded)
```
kiryldz, line 137: also could check that all previous styles are not valid and the only valid one is the last one

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
```diff
@@ -69,10 +69,15 @@ internal class StyleObserver(
...
   override fun onStyleLoaded(eventData: StyleLoadedEventData) {
     val style = loadedStyle ?: throw MapboxMapException("Style is not initialized on onStyleLoaded callback!")
+
+    onStyleSpritesReady()
+    onStyleSourcesReady()
     styleLoadedListener.onStyleLoaded(style)
 
-    userStyleLoadedListener?.onStyleLoaded(style)
-    userStyleLoadedListener = null
+    userStyleLoadedListener?.let {
```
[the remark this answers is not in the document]
kiryldz: isn't this logic exactly the same as for `onStyleSpritesReady` - I'd move this to dedicated function

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
yunikkk: We won't receive any styles in the previous cases.
Added counter to check that.

sdk/src/main/java/com/mapbox/maps/StyleObserver.kt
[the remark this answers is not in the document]
yunikkk: Not really, those contains comments and are named accordingly. Talked to leave as is

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
```diff
@@ -265,6 +265,15 @@ internal class NativeObserver(
     }
   }
 
+  // This method is needed to prevent receiving styleLoaded events with the old style
```
yunikkk, line 268: @alexshalamov ptal, added resubscription to the STYLE_LOADED event that is called on every style change

@mapbox/maps-android is the comment clear at all?) Dirty hack but fixes the possible corner-case issues with loading multiple styles consecutively

alexshalamov: @yunikkk I was thinking whether we could push this logic down to the core renderer. Maybe for some of the events, gl-native could add `cancel and re-subscribe` code when new style is loaded, so that both iOS and Android platform would benefit from the fix.

Meanwhile, workaround on platform side should be sufficient for corner-cases.

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
```diff
@@ -265,6 +265,15 @@ internal class NativeObserver(
...
+  // when new style is loaded. For example loading empty style and another style immediately after
+  // that will notify STYLE_LOADED event for the first style after second style has already started
+  // loading
+  fun resubscribeStyleLoadListeners() {
+    unsubscribeUnusedEvent(MapEvents.STYLE_LOADED)
```
alexshalamov, line 273: I would also add `MapEvents.STYLE_DATA_LOADED`

alexshalamov (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Features ✨ and improvements 🏁
 * Update SDK name in attribution action sheet. ([1375](https://github.com/mapbox/mapbox-maps-android/pull/1375))
 * Introduce experimental ModelLayer API to render 3D models on the map. ([#1369](https://github.com/mapbox/mapbox-maps-android/pull/1369))
+* Further optimize `MapboxMap.loadStyle()` to apply styling properties earlier. [#1378](https://github.com/mapbox/mapbox-maps-android/pull/1378)
```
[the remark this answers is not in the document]
yunikkk: Ok, I've wanted to highlight that it's different from the previous optimization at #1362

sdk/src/main/java/com/mapbox/maps/NativeObserver.kt
yunikkk: Added

PR #1416, opened 2022-06-07
Fix issue when tasks are skipped if surface is recreated - MAPSAND-177.

yunikkk:
### Summary of changes

Previously we were ignoring tasks scheduled from gl-native if there was no EGL / render was not prepared fully, as well as clearing the queues with tasks when new surface arrives. 
Hence, various async tasks scheduled from gl-native could be just silently removed from the queue when new surface arrives (e.g. if activity was paused and then resumed). 
For example such behaviour caused 3d puck (and other models loaded async, potentially) not being loaded. 

With this change we will reschedule all the tasks indefinitely (with 50 millis delay), until with render thread is not fully destroyed (together with gl-native's renderer).

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
@@ -5,14 +5,15 @@ package com.mapbox.maps.renderer
  */
 internal enum class EventType {
   /**
-   * Those events are scheduled not by the Maps SDK (most likely by the user using MapView#queueEvent)
+   * Those events are scheduled by the Maps SDK or by user.
    * Such events do not get cleared when Android provides new surface or texture.
-   * In case render thread is not fully prepared - such messages will keep being rescheduled until
+   * If render thread is not fully prepared - such messages will keep being rescheduled until
    * thread is prepared or killed.
    */
-  OTHER,
+  DEFAULT,
 
   /**
+   * Special event type to handle when renderer is being destroyed
    * TODO remove when gl-native fix will land
    */
   DESTROY_RENDERER,
--- a/sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
@@ -5,15 +5,10 @@ package com.mapbox.maps.renderer
  */
 internal enum class EventType {
   /**
-   * Those events are scheduled by our SDK.
-   * Such events get cleared when Android provides new surface or texture.
-   */
-  SDK,
-
-  /**
-   * Those events are scheduled by not our SDK (most likely by the user using MapView#queueEvent)
+   * Those events are scheduled not by the Maps SDK (most likely by the user using MapView#queueEvent)
    * Such events do not get cleared when Android provides new surface or texture.
-   * In case render thread is not fully prepared - such messages will keep being rescheduled until thread is prepared or killed.
+   * In case render thread is not fully prepared - such messages will keep being rescheduled until
+   * thread is prepared or killed.
    */
   OTHER,
 
--- a/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
@@ -307,6 +307,8 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     renderDestroyCallChain = true
     mapboxRenderer.destroyRenderer()
     renderDestroyCallChain = false
+    renderEventQueue.clear()
+    nonRenderEventQueue.clear()
     renderCreated = false
     releaseEgl()
     surface?.release()
@@ -325,6 +327,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       // at least on Android 8 devices we create surface before Activity#onStart
       // so we need to proceed to EGL creation in any case to avoid deadlock
       if (!creatingSurface) {
+        logW(TAG, "Skip render frame - NOT creating surface but renderNotSupported ($renderNotSupported) || paused ($paused)")
         return
       }
     }
@@ -332,6 +335,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     if (creatingSurface || !renderThreadPrepared) {
       val renderThreadPreparedOk = setUpRenderThread(creatingSurface)
       if (!renderThreadPreparedOk) {
+        logW(TAG, "Skip render frame - NOT render thread prepared but creatingSurface ($creatingSurface) || !renderThreadPrepared (${!renderThreadPrepared})")
         return
       }
     }
@@ -355,6 +359,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
 
   @UiThread
   fun onSurfaceDestroyed() {
+    logW(TAG, "RenderThread : surface destroyed")
     lock.withLock {
       // in some situations `destroy` is called earlier than onSurfaceDestroyed - in that case no need to clean up
       if (renderHandlerThread.started) {
@@ -386,6 +391,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   internal fun processAndroidSurface(surface: Surface, width: Int, height: Int) {
     if (this.surface != surface) {
       if (this.surface != null) {
+        logW(TAG, "Process android surface while current is not null, will release EGL")
         releaseEgl()
         this.surface?.release()
       }
@@ -394,10 +400,6 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     this.width = width
     this.height = height
     widgetRenderer.onSurfaceChanged(width = width, height = height)
-    renderEventQueue.removeAll { it.eventType == EventType.SDK }
-    nonRenderEventQueue.removeAll { it.eventType == EventType.SDK }
-    // we do not want to clear render events scheduled by user
-    renderHandlerThread.clearMessageQueue(clearAll = false)
     prepareRenderFrame(creatingSurface = true)
   }
 
@@ -435,24 +437,9 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       renderEvent.runnable?.let {
         renderEventQueue.add(renderEvent)
       }
-      // in case of native Mapbox events we schedule render event only when we're fully ready for render
-      if (renderEvent.eventType == EventType.SDK) {
-        if (renderThreadPrepared) {
-          postPrepareRenderFrame()
-        }
-      } else {
-        postPrepareRenderFrame()
-      }
+      postPrepareRenderFrame()
     } else {
-      // no sense in scheduling non-render SDK tasks if thread is not fully ready for render
-      // as render thread SDK tasks queue will be cleared when new surface will arrive
-      if (renderEvent.eventType == EventType.SDK) {
-        if (renderThreadPrepared) {
-          postNonRenderEvent(renderEvent)
-        }
-      } else {
-        postNonRenderEvent(renderEvent)
-      }
+      postNonRenderEvent(renderEvent)
     }
   }
 
@@ -501,7 +488,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
             if (renderCreated) {
               releaseAll()
             }
-            renderHandlerThread.clearMessageQueue(clearAll = true)
+            renderHandlerThread.clearMessageQueue()
             destroyCondition.signal()
           }
         }
--- a/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
@@ -328,7 +328,11 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       // at least on Android 8 devices we create surface before Activity#onStart
       // so we need to proceed to EGL creation in any case to avoid deadlock
       if (!creatingSurface) {
-        logI(TAG, "Skip render frame - NOT creating surface but renderNotSupported ($renderNotSupported) || paused ($paused)")
+        logI(
+          TAG,
+          "Skip render frame - NOT creating surface although " +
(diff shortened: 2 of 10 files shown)


sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -435,24 +437,9 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
       }
-      // in case of native Mapbox events we schedule render event only when we're fully ready for render
-      if (renderEvent.eventType == EventType.SDK) {
-        if (renderThreadPrepared) {
-          postPrepareRenderFrame()
-        }
-      } else {
-        postPrepareRenderFrame()
-      }
+      postPrepareRenderFrame()
```
yunikkk, line 440: Here we previously were skipping render events if renderThreadPrepared was false

sdk/src/main/java/com/mapbox/maps/renderer/EventType.kt
```diff
@@ -5,15 +5,10 @@ package com.mapbox.maps.renderer
...
-
-  /**
-   * Those events are scheduled by not our SDK (most likely by the user using MapView#queueEvent)
+   * Those events are scheduled not by the Maps SDK (most likely by the user using MapView#queueEvent)
    * Such events do not get cleared when Android provides new surface or texture.
-   * In case render thread is not fully prepared - such messages will keep being rescheduled until thread is prepared or killed.
+   * In case render thread is not fully prepared - such messages will keep being rescheduled until
+   * thread is prepared or killed.
    */
   OTHER,
```
kiryldz, line 13: rename to `DEFAULT`?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -355,6 +359,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
 
   @UiThread
   fun onSurfaceDestroyed() {
+    logW(TAG, "RenderThread : surface destroyed")
```
kiryldz, line 362: let's make it `info` instead, I don't think it's a warning

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -325,13 +327,15 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
       if (!creatingSurface) {
+        logW(TAG, "Skip render frame - NOT creating surface but renderNotSupported ($renderNotSupported) || paused ($paused)")
         return
       }
     }
     // check for creatingSurface flag to make sure we don't hit deadlock
     if (creatingSurface || !renderThreadPrepared) {
       val renderThreadPreparedOk = setUpRenderThread(creatingSurface)
       if (!renderThreadPreparedOk) {
+        logW(TAG, "Skip render frame - NOT render thread prepared but creatingSurface ($creatingSurface) || !renderThreadPrepared (${!renderThreadPrepared})")
```
kiryldz, line 338:
```suggestion
        logW(TAG, "Skip render frame - render thread NOT prepared although creatingSurface ($creatingSurface) || !renderThreadPrepared (${!renderThreadPrepared})")
```

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -325,13 +327,15 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
       // at least on Android 8 devices we create surface before Activity#onStart
       // so we need to proceed to EGL creation in any case to avoid deadlock
       if (!creatingSurface) {
+        logW(TAG, "Skip render frame - NOT creating surface but renderNotSupported ($renderNotSupported) || paused ($paused)")
```
kiryldz, line 330:
```suggestion
        logW(TAG, "Skip render frame - NOT creating surface although renderNotSupported ($renderNotSupported) || paused ($paused)")
```

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -232,6 +234,6 @@ internal abstract class MapboxRenderer : MapClient {
   companion object {
     private const val TAG = "Mbgl-Renderer"
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal val renderEventSdk = RenderEvent(null, true, EventType.SDK)
+    internal val repaintRenderEvent = RenderEvent(null, true, EventType.OTHER)
```
pengdev, line 237: the name is more meaningful to me now 👍

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -48,6 +48,7 @@ internal abstract class MapboxRenderer : MapClient {
 
   @UiThread
   fun onDestroy() {
+    logW(TAG, "onDestroy")
```
kiryldz, line 51: remove or introduce better log message with `info` level

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -113,6 +114,7 @@ internal abstract class MapboxRenderer : MapClient {
 
   @WorkerThread
   fun destroyRenderer() {
+    logW(TAG, "Destroy renderer")
```
kiryldz, line 117: level `info` will be better

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
pengdev, line 362: nit: We are adding too many logs imo, and these shouldn't be warning messages.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -307,6 +307,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     renderDestroyCallChain = true
     mapboxRenderer.destroyRenderer()
     renderDestroyCallChain = false
+    nonRenderEventQueue.clear()
```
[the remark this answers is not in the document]
kiryldz: Still concerned that `renderHandlerThread.clearMessageQueue()` is not executed here. That actually means that if we're using texture mode in current implementation and we do have some __non `EventType.DESTROY_RENDERER` SDK events__ in handler thread message queue - they will be actually executed with new surface (when `renderThreadPrepared` will become `true`) and new native renderer instance leading potentially to hard catching errors.

IMO we need to revert flag and call `renderHandlerThread.clearMessageQueue(clearSdkEventsOnly = true)` - in that case we will leave `DESTROY_RENDERER` in the queue and they will be executed ASAP (as they don't need `renderThreadPrepared == true`) but truly render events will be cleared. However ideally we should not clear user events then meaning we still need to differentiate SDK and non-SDK events.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: personally I agree on leaving logs related to surface / activity lifecycle here as those events are pretty rare but at the same time will give us pretty much info in case of some rendering issues.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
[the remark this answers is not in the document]
yunikkk: Added clearing messages in the texture view destroy usecase as well. However I dont think we need to keep user's render events in that case, since: 
- it's an edge case for the old Androids, so will rarely happen
- in theory events may be triggered by map loaded events, so will depend on the renderer lifecycle, thus executing the events is probably not expected. Lets sync offline on that

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: @pengdev I've switched all new logs to INFO, and @kiryldz is right about the intentions - debug will be much easier while logs only happen on map creation/destruction or pauses, which is pretty occasional

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: I assume it was just incorrect. This was the actual reason why non-render events (e.g. 3d puck model load event) that may arrive after surface destroy (hence renderThreadPrepared == false) were just skipped.

yunikkk: @kiryldz there were no other reasons, right?

yunikkk: UPD - this was about non-render events, of course. 
For the render events it might be performance optimization indeed, to avoid making `post`, that doesn't seem that impactful anyway. 
@kiryldz could you bring more context here?

kiryldz: seems fine to avoid making redundant post, runnable is saved in the queue anyway before

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
```diff
@@ -46,12 +48,11 @@ internal class RenderHandlerThread {
...
   }
 
-  fun clearMessageQueue(clearAll: Boolean = true) {
-    if (clearAll) {
-      handler?.removeCallbacksAndMessages(null)
-    } else {
-      handler?.removeCallbacksAndMessages(EventType.SDK)
-    }
+  /**
+   * Clears all messages except of [EventType.DESTROY_RENDERER] messages.
```
kiryldz, line 52: perhaps add few more words here clarifying why we never actually need `DESTROY_RENDERER` be cleared?

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
```diff
@@ -37,7 +37,9 @@ internal class RenderHandlerThread {
   }
 
   fun stop() {
-    handlerThread.quit()
+    // quit and wait until all the messages will be processed - DESTROY_RENDERER messages
+    // are still there
+    handlerThread.quitSafely()
```
kiryldz, line 42: As discussed - may affect performance as we'll block main Android UI thread. Assuming destroy render tasks are not related to GPU resources and that we have additional mutex control to block UI thread until EGL is cleared - perhaps remove join at all then?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
yunikkk: > seems fine to avoid making redundant post, runnable is saved in the queue anyway before

Agree, postRender will anyway be triggered when rendering will continue

yunikkk: Updated

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
yunikkk: Well `quitSafely` itself will wait for all the tasks to be executed - `join` won't add anything in that case.

yunikkk: At the same time join seems indeed unneeded, lets remove it then.

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
yunikkk: Added

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRenderThreadTest.kt
```diff
@@ -679,4 +687,52 @@ class MapboxRenderThreadTest {
...
+
+    lateinit var runnable: Runnable
+    lateinit var runnable2: Runnable
+    waitZeroCounter(startCounter = 2) {
+      runnable = mockCountdownRunnable(this)
+      runnable2 = mockCountdownRunnable(this)
+      mapboxRenderThread.onSurfaceDestroyed()
+
+      mapboxRenderThread.queueRenderEvent(RenderEvent(runnable, false, EventType.DEFAULT))
+      mapboxRenderThread.queueRenderEvent(RenderEvent(runnable2, false, EventType.DEFAULT))
```
kiryldz, line 728: not verifying `runnable2`

kiryldz (approved):

sdk/src/test/java/com/mapbox/maps/renderer/RenderHandlerThreadTest.kt
```diff
@@ -70,7 +59,23 @@ class RenderHandlerThreadTest {
...
+  @Test
+  fun postThreadStopped() {
+    val actionOne = mockk<() -> Unit>(relaxed = true)
+    val actionTwo = mockk<() -> Unit>(relaxed = true)
+    Shadows.shadowOf(Looper.getMainLooper()).pause()
+    renderHandlerThread.apply {
+      start()
+      post { actionOne() }
+      stop()
+      post { actionTwo() }
```
alexander-kulikovskii, line 74: should we check also `postDelayed` with some delay that happens after `stop`?
and check thread queue.

yunikkk: Ok with the delay, not sure it's worth to check the queue directly though

pengdev (approved):
LGTM, Great work 💎

alexander-kulikovskii (approved):

PR #1427, opened 2022-06-14
Introduce API to control errors during renderer creation

kiryldz:
### Summary of changes

Introduce `addRendererSetupErrorListener`/`removeRendererSetupErrorListener` methods for `MapView` and `MapSurface` to listen to renderer setup errors and give opportunity to control some edge cases.

This could be useful to use this API and listen to `RendererError.OutOfMemory` error when initiating the `MapView`. Then, in case of no resources, it could be an option try cleaning up the memory for your app and re-create the `MapView`. Sample use-case was added as the test in `RendererSetupTest`.

## Pull request checklist:
    
### DIFF
--- a/sdk/src/androidTest/java/com/mapbox/maps/renderer/egl/RendererSetupTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/renderer/egl/RendererSetupTest.kt
@@ -5,6 +5,7 @@ import android.os.Handler
 import android.os.Looper
 import android.view.ViewGroup
 import androidx.core.os.postDelayed
+import androidx.test.annotation.UiThreadTest
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
@@ -35,6 +36,7 @@ class RendererSetupTest {
   }
 
   @Before
+  @UiThreadTest
   fun setUp() {
     countDownLatch = CountDownLatch(1)
     eventList.clear()
@@ -70,6 +72,41 @@ class RendererSetupTest {
   }
 
   @Test
+  fun severalErrorListenersTest() {
+    countDownLatch = CountDownLatch(2)
+    rule.scenario.onActivity { activity ->
+      EGLConfigChooser.STENCIL_SIZE = INVALID_STENCIL_SIZE
+      val mapView = MapView(activity)
+      activity.frameLayout.addView(mapView)
+      mapView.addRendererSetupErrorListener {
+        when (it) {
+          RendererError.NO_VALID_EGL_CONFIG_FOUND -> {
+            eventList.add(Event.RENDERER_ERROR)
+            countDownLatch.countDown()
+          }
+          RendererError.OUT_OF_MEMORY -> { /** no-op **/ }
+          else -> { /** no-op **/ }
+        }
+      }
+      mapView.addRendererSetupErrorListener {
+        when (it) {
+          RendererError.NO_VALID_EGL_CONFIG_FOUND -> {
+            eventList.add(Event.RENDERER_ERROR)
+            countDownLatch.countDown()
+          }
+          RendererError.OUT_OF_MEMORY -> { /** no-op **/ }
+          else -> { /** no-op **/ }
+        }
+      }
+    }
+    countDownLatch.await(DEFAULT_LATCH_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+    assertArrayEquals(
+      arrayOf(Event.RENDERER_ERROR, Event.RENDERER_ERROR),
+      eventList.toArray()
+    )
+  }
+
+  @Test
   fun recreateMapWithDelayOnErrorTest() {
     rule.scenario.onActivity {
       EGLConfigChooser.STENCIL_SIZE = INVALID_STENCIL_SIZE
@@ -87,18 +124,19 @@ class RendererSetupTest {
   private fun createMapView(
     activity: Activity,
     mapView: MapView,
-    withDelayMs: Long
+    withDelayMs: Long,
   ) {
     mapView.onStart()
     mapView.getMapboxMap().addOnMapLoadedListener {
       eventList.add(Event.MAP_LOAD_SUCCESS)
+      mapView.onStop()
+      mapView.onDestroy()
       countDownLatch.countDown()
     }
     mainHandler.postDelayed(withDelayMs) {
       mapView.addRendererSetupErrorListener {
         when (it) {
-          is RendererError.EglError -> { /** no-op **/ }
-          RendererError.NoValidEglConfigFound -> {
+          RendererError.NO_VALID_EGL_CONFIG_FOUND -> {
             eventList.add(Event.RENDERER_ERROR)
             mapView.onStop()
             mapView.onDestroy()
@@ -110,13 +148,15 @@ class RendererSetupTest {
             // create valid map view without delay to subscribe to new renderer errors asap
             createMapView(activity, validMapView, withDelayMs = 0)
           }
-          RendererError.OutOfMemory -> { /** no-op **/ }
+          RendererError.OUT_OF_MEMORY -> { /** no-op **/ }
+          else -> { /** no-op **/ }
         }
       }
     }
   }
 
   @After
+  @UiThreadTest
   fun cleanUp() {
     EGLConfigChooser.STENCIL_SIZE = VALID_STENCIL_SIZE
   }
--- a/sdk/src/androidTest/java/com/mapbox/maps/renderer/egl/RendererSetupTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/renderer/egl/RendererSetupTest.kt
@@ -1,0 +1,129 @@
+package com.mapbox.maps.renderer.egl
+
+import android.app.Activity
+import android.os.Handler
+import android.os.Looper
+import android.view.ViewGroup
+import androidx.core.os.postDelayed
+import androidx.test.ext.junit.rules.ActivityScenarioRule
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.LargeTest
+import com.mapbox.maps.EmptyActivity
+import com.mapbox.maps.MapView
+import com.mapbox.maps.renderer.RendererError
+import org.junit.After
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CopyOnWriteArrayList
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class RendererSetupTest {
+
+  private lateinit var countDownLatch: CountDownLatch
+  private val eventList = CopyOnWriteArrayList<Event>()
+  private val mainHandler = Handler(Looper.getMainLooper())
+
+  @get:Rule
+  val rule = ActivityScenarioRule(EmptyActivity::class.java)
+
+  enum class Event {
+    MAP_LOAD_SUCCESS,
+    RENDERER_ERROR,
+  }
+
+  @Test
+  fun regularCreateMapTest() {
+    countDownLatch = CountDownLatch(1)
+    eventList.clear()
+    rule.scenario.onActivity {
+      val mapView = MapView(it)
+      it.frameLayout.addView(mapView)
+      createMapView(it, mapView, withDelayMs = 0L)
+    }
+    countDownLatch.await(DEFAULT_LATCH_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+    Assert.assertArrayEquals(
+      arrayOf(Event.MAP_LOAD_SUCCESS),
+      eventList.toArray()
+    )
+  }
+
+  @Test
+  fun recreateMapImmediateOnErrorTest() {
+    countDownLatch = CountDownLatch(1)
+    eventList.clear()
+    rule.scenario.onActivity {
+      EGLConfigChooser.INVALID_CONFIG_FOR_TEST = true
+      val mapView = MapView(it)
+      it.frameLayout.addView(mapView)
+      createMapView(it, mapView, withDelayMs = 0L)
+    }
+    countDownLatch.await(DEFAULT_LATCH_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+    Assert.assertArrayEquals(
+      arrayOf(Event.RENDERER_ERROR, Event.MAP_LOAD_SUCCESS),
+      eventList.toArray()
+    )
+  }
+
+  @Test
+  fun recreateMapWithDelayOnErrorTest() {
+    countDownLatch = CountDownLatch(1)
+    eventList.clear()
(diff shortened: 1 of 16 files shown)


sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
```diff
@@ -0,0 +1,100 @@
+package com.mapbox.maps.renderer
+
+import com.mapbox.maps.MapView
+import javax.microedition.khronos.egl.EGL10
+import javax.microedition.khronos.egl.EGL11
+
+/**
+ * Typed enum wrapping up all the EGL error codes represented as ints.
+ */
+enum class RendererError {
```
pengdev, line 10: Does it cover all the error types? Maybe we should just propagate the error code as Integers to the user?

kiryldz: it does cover all EGL errors based on the [spec](https://www.khronos.org/files/egl-1-4-quick-reference-card.pdf). Even this is imho a bit of an overkill but decided that better more than less 😄

sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
```diff
@@ -0,0 +1,100 @@
...
+   */
+  EGL_CONTEXT_LOST,
+
+  /**
+   * Represents unknown error that we could not map to any existing EGL error code.
+   */
+  EGL_UNKNOWN
+}
+
+internal fun Int?.mapEglErrorCode(): RendererError {
```
yunikkk, line 82: I personally don't like extensions on basic type - could be imported accidentally in the wrong context. 
IMHO static method `fromInt` in the enum itself would equally clear

yunikkk: Aaaand actually if you'll add int as a member of the enum constant itself - this `when` will be unneeded, will be able to just iterate through values and pick the right one in a simple loop

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -176,16 +179,37 @@ internal class EGLCore(
...
+
+  internal fun removeRendererStateListener(listener: RendererSetupErrorListener) {
+    rendererSetupErrorListenerSet.remove(listener)
+  }
+
+  internal fun clearRendererStateListeners() {
+    rendererSetupErrorListenerSet.clear()
+  }
+
+  private fun notifyRendererSetupErrorListenersIfNeeded(functionName: String): Int? {
```
yunikkk, line 194: IMHO for `notify` it's taking too much responsibilities, I'd rename to e.g. `checkEglErrorAndNotify` or split it to several funs

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -176,16 +179,37 @@ internal class EGLCore(
...
+    val eglError = checkEglError(functionName)
```
alexander-kulikovskii, line 195: btw, if `checkEglError(..)` checks `eglGetError()` and it return `null` as success result, next line will check `null != EGL10.EGL_SUCCESS` and executes block below.

kiryldz: I didn't find a way to actually trigger `eglGetError()` return `null`. Even if does - with current logic we will trigger an error `EglError(null)` which is OK as null returned is def some kind of error.

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: fixed

sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
kiryldz: revisited whole class, PTAL again

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: testing did actually show functions when `null` is returned and this is not considered to be an error, will fix it. Thanks for pointing it out!

kiryldz: in general I did mess up with logic a bit, EGL error could not even be null.

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -176,16 +192,60 @@ internal class EGLCore(
     return EGL10.EGL_SUCCESS
   }
 
-  /**
-   * Checks for EGL errors.
-   */
-  private fun checkEglErrorNoException(msg: String): Boolean {
+  internal fun addRendererStateListener(listener: RendererSetupErrorListener) {
```
yunikkk, line 195: Isn't it supposed to be launched on main thread already?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -24,45 +33,56 @@ internal class EGLCore(
...
+   */
+  private val mainHandler = Handler(Looper.getMainLooper())
+  /**
+   * We require error accumulation as renderer is created when [MapView] is created and
+   * some errors could already be triggered on render thread
+   * before user will call [MapView.addRendererSetupErrorListener] -
+   * in that case accumulated errors will still be delivered to the user.
+   */
+  private val accumulatedRendererErrorList = CopyOnWriteArrayList<RendererError>()
+  private val rendererSetupErrorListenerSet = CopyOnWriteArraySet<RendererSetupErrorListener>()
```
yunikkk, line 47: nit : add annotations or comments to indicate which thread accesses both?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: adding listener - yes (however does not matter tbh), why are you asking?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: they are both touched from both threads

yunikkk: For `accumulatedRendererErrorList` there seem one single place with `isNotEmpty` check made from main is that right?

kiryldz: `accumulatedRendererErrorList` is also cleared from main thread and local copy is created from main thread as well

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
yunikkk: Since there's `mainHandler.post {` that is unneeded then

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -176,16 +192,60 @@ internal class EGLCore(
...
   }
 
-  /**
-   * Checks for EGL errors.
-   */
-  private fun checkEglErrorNoException(msg: String): Boolean {
+  internal fun addRendererStateListener(listener: RendererSetupErrorListener) {
+    rendererSetupErrorListenerSet.add(listener)
+    if (accumulatedRendererErrorList.isNotEmpty()) {
+      val errorListCopy = LinkedList(accumulatedRendererErrorList)
```
yunikkk, line 198: Can you just take iterator() and clear afterwards instead of copy?

kiryldz: did hide this under the lock, making a copy then is at least more clear for me to understand than using iterator for thread-safe collection.

kiryldz: actually revisited and did as you said

sdk/src/androidTest/java/com/mapbox/maps/renderer/egl/RendererSetupTest.kt
```diff
@@ -0,0 +1,129 @@
...
+  fun regularCreateMapTest() {
+    countDownLatch = CountDownLatch(1)
+    eventList.clear()
+    rule.scenario.onActivity {
+      val mapView = MapView(it)
+      it.frameLayout.addView(mapView)
+      createMapView(it, mapView, withDelayMs = 0L)
+    }
+    countDownLatch.await(DEFAULT_LATCH_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+    Assert.assertArrayEquals(
```
yunikkk, line 49: nit : import

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLConfigChooser.kt
```diff
@@ -372,5 +373,8 @@ internal class EGLConfigChooser constructor(
      * @see android.opengl.EGL14.EGL_OPENGL_ES2_BIT;
      */
     private const val EGL_OPENGL_ES2_BIT = 0x0004
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
yunikkk, line 377: Would you like to also expose the rest of attributes and tests for them?

yunikkk: Just curious

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -176,16 +194,67 @@ internal class EGLCore(
...
-  private fun checkEglErrorNoException(msg: String): Boolean {
+  @MainThread
+  internal fun addRendererStateListener(listener: RendererSetupErrorListener) {
+    rendererErrorLock.withLock {
+      rendererSetupErrorListenerSet.add(listener)
+      if (accumulatedRendererErrorList.isNotEmpty()) {
+        accumulatedRendererErrorList.forEach { error ->
+          listener.onError(error)
+        }
+        accumulatedRendererErrorList.clear()
```
pengdev, line 205: wonder if we should clear the error list, we could keep them and only clear on destroy the renderer. And document that all the `addRendererStateListener` will return all the egl errors.

Asking because how would user know which `addRendererStateListener` call is made "first" if they add the listener multiple places in their code base? And the listener itself doesn't report this "isFirst" flag back to the caller. Imo we should make this API consistent no matter if it's called the first time or multiple times after the initial subscription.

sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
```diff
@@ -0,0 +1,28 @@
...
+   * This could be also checked with static method [MapView.isRenderingSupported] prior to creating the [MapView].
+   */
+  object NoValidEglConfigFound : RendererError(0)
+  /**
+   * Represents out of memory caused by [EGL10.EGL_BAD_ALLOC].
+   *
+   * Could be thrown when creating context, surface or display and system is running low on resources.
+   * Best action will be try to release Android resources and re-create the [MapView].
+   */
+  object OutOfMemory : RendererError(EGL10.EGL_BAD_ALLOC)
```
pengdev, line 21: are these two cases the only special errors? would we add something later as well? to avoid breaking change, we could probably use 
```
class RendererError(val eglErrorCode: Int) {
  companion object {
    object NoValidEglConfigFound : RendererError(0)
    object OutOfMemory : RendererError(EGL10.EGL_BAD_ALLOC)
  }
}
```
so that we can add more objects without breaking changes later..

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: done!

sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
kiryldz: done!

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -194,40 +191,37 @@ internal class EGLCore(
...
     }
   }
 
-  @MainThread
+  @WorkerThread
   internal fun removeRendererStateListener(listener: RendererSetupErrorListener) {
-    rendererErrorLock.withLock {
-      rendererSetupErrorListenerSet.remove(listener)
-    }
+    rendererSetupErrorListenerSet.remove(listener)
```
yunikkk, line 209: Maybe just post to render thread here and keep semantics of adding listeners from MainThread since we notify on main also?

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
```diff
@@ -244,14 +238,11 @@ internal class EGLCore(
...
-    rendererErrorLock.withLock {
-      if (rendererSetupErrorListenerSet.isEmpty()) {
-        accumulatedRendererErrorList.add(error)
-      } else {
-        mainHandler.post {
-          rendererSetupErrorListenerSet.forEach {
-            it.onError(error)
-          }
+    accumulatedRendererErrorList.add(error)
+    if (rendererSetupErrorListenerSet.isNotEmpty()) {
```
yunikkk, line 242: You'll need to copy lists here as well right

sdk/src/main/java/com/mapbox/maps/renderer/egl/EGLCore.kt
kiryldz: we don't have render thread instance in egl core

sdk/src/main/java/com/mapbox/maps/renderer/RendererError.kt
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Typed enum wrapping up all the EGL error codes represented as ints.
+ * We do have SDK specific error types: [NO_VALID_EGL_CONFIG_FOUND] and [OUT_OF_MEMORY].
+ */
+class RendererError(val eglErrorCode: Int) {
+  /**
+   * Indicates whether some other object is "equal to" this one.
+   */
+  override fun equals(other: Any?): Boolean = other is RendererError &&
```
alexander-kulikovskii, line 14: will be nice to add [EqualsHashCodeTest](https://github.com/mapbox/mapbox-maps-android/blob/main/extension-style/src/test/java/com/mapbox/maps/extension/style/EqualsHashCodeTest.kt)

kiryldz: thanks for reminding, will add

yunikkk (approved):

PR #1440, opened 2022-06-17
Fix possibly slow move gesture.

yunikkk:
### Summary of changes

Fixes issue observed in some cases when part of the movement gesture is ignored (feels like move is slower). 

## Pull request checklist:
    

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -99,10 +99,17 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
...
-  // duration = 0 will still make animation end / cancel not immediately
-  // this may cause cancelling some easeTo animation without single camera update
+  // Devices with API <= 23 animating with duration = 0 will send initial value from Looper with
+  // some small delay as a result.
+  // Hence multiple immediate animations sent close to each other may cancel previous ones.
+  // This becomes visible with move gesture since animation steps are relative to each other.
+  // So we defer move distances that are not applied to apply later (when immediate ease will end).
+  // Does not apply to shove/scale/rotate gestures since they use value from detector that is
+  // relative to the animation start.
   private var immediateEaseInProcess = false
```
yunikkk, line 109: `onAnimationEnd` callback is fired from the Handler of the main thread, thus busy thread or some system lag may keep the `immediateEaseInProcess` flag set until the next onMove gesture arrives and is skipped because of the flag.

kiryldz (commented):
Great improvement, but several points I'd like to mention:
1. Could this be relevant to other (non-move) gestures? 
2. On weak devices this fix will actually bring in sluggish pan behaviour - but correct in terms of actual finger velocity comparing to current approach where we actually do not have sluggish behaviour but finger velocity is not treated correctly (map does not pan as far as it is supposed).
3. Tests are missing, just flagging.

yunikkk:
> Great improvement, but several points I'd like to mention:
> 
> 1. Could this be relevant to other (non-move) gestures?
> 2. On weak devices this fix will actually bring in sluggish pan behaviour - but correct in terms of actual finger velocity comparing to current approach where we actually do not have sluggish behaviour but finger velocity is not treated correctly (map does not pan as far as it is supposed).
> 3. Tests are missing, just flagging.

1. Answer was in the comment - however I've thought once more about it and it seems that only `shove` won't need this fix. Double-checking.
2. Doesn't feel sluggish, actually. E.g. instead of 2 pixels next frame will move for 4. Will record the video before/after to illustrate.
3. Yup

yunikkk:
@kiryldz added the rest of gestures, all of them were affected, except for the quick-zoom.

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1013,8 +1036,13 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
     deltaPixelsSinceLast: Float
   ): Boolean {
     // Get pitch value (scale and clamp)
```
kiryldz, line 1038: this comment becomes confusing now, should be moved to line 1043

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -675,6 +688,7 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
 
   internal fun handleScaleBegin(detector: StandardScaleGestureDetector): Boolean {
     quickZoom = detector.pointersCount == 1
+    deferredRotate = 0.0
```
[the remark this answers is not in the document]
kiryldz: applies for several similar code blocks below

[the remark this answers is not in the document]
yunikkk: Well it will conflict with the current `end` logic - since there a different animation is applied based on the velocity and it's not really clear how to mix those two. I'd just allowed the last value to be silently ignored, that's not really noticeable.

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1423,4 +1581,15 @@ class FlingGestureTest(
...
+internal fun verifyNo(
+  ordering: Ordering = Ordering.UNORDERED,
+  timeout: Long = 0,
+  verifyBlock: MockKVerificationScope.() -> Unit
+) = verify(
+  ordering = ordering,
+  exactly = 0,
+  timeout = timeout,
+  verifyBlock = verifyBlock,
+)
```
kiryldz, line 1595: I'd rather either move this to some separate common test module and replace across the whole codebase or revert here back to `verify(exactly = 0)` to make those tests consistent with all others

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -585,7 +585,7 @@ class GesturesPluginTest {
     verify { listener.onScaleBegin(any()) }
     presenter.removeOnScaleListener(listener)
     presenter.handleScaleBegin(scaleDetector)
-    verify(exactly = 1) { listener.onScaleBegin(any()) }
+    verify { listener.onScaleBegin(any()) }
```
kiryldz, line 588: potential regression as `verify` without parameters does not take number of verifies into account, please revert back here and everywhere else

yunikkk: Thanks, indeed, somehow seemed `verify` uses exactly = 1 by default. Reverted.

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
yunikkk: Yeah, can't use veiryfoNo util from the sdk tests unfortunately, tried to introduce here, but ok to use exactly = 0 until extenstion will be imported

kiryldz (commented):
changelog entry missing

kiryldz (approved):

pengdev (approved):
Great work! would be good to have some pixel match based instrumentation test.

yunikkk:
> Great work! would be good to have some pixel match based instrumentation test.

Ticketed that, nice idea

PR #1441, opened 2022-06-17
[test] enable release testapp to playstore MAPSAND-120

ank27:
### Summary of changes
Adding support to publish test application on internal channel.

## Pull request checklist:
    

gradle/play-publisher.gradle
```diff
@@ -0,0 +1,7 @@
+apply plugin: 'com.github.triplet.play'
+
+play {
+    track = findProperty("releaseTrack") as String ?: "internal"
```
alexander-kulikovskii, line 4: maybe throw exception?

gradle/script-git-version.gradle
```diff
@@ -0,0 +1,18 @@
...
+    }
+}
+
+import org.ajoberstar.grgit.Grgit
+
+ext {
+    println("projectDir = "+projectDir)
+    git = Grgit.open(currentDir: projectDir)
+    def tagSize = git.tag.list().size()
+    gitVersionCode = tagSize+1
```
alexander-kulikovskii, line 16: nit:
```suggestion
    gitVersionCode = tagSize + 1
```

gradle/script-git-version.gradle
```diff
@@ -0,0 +1,18 @@
...
+    }
+    dependencies {
+        classpath("org.ajoberstar.grgit:grgit-gradle:4.1.1")
+    }
+}
+
+import org.ajoberstar.grgit.Grgit
+
+ext {
+    println("projectDir = "+projectDir)
```
Chaoba, line 13:
```suggestion
    println("projectDir = " + projectDir)
```

Chaoba (approved):

gradle/play-publisher.gradle
```diff
@@ -0,0 +1,7 @@
...
+    serviceAccountCredentials = file("test-app-play-publisher.json")
```
yunikkk, line 5: Where is this file supposed to come from?

gradle/script-git-version.gradle
```diff
@@ -0,0 +1,18 @@
...
+    git = Grgit.open(currentDir: projectDir)
+    def tagSize = git.tag.list().size()
+    gitVersionCode = tagSize+1
+    gitVersionName = "0.1.$tagSize".toString()
```
yunikkk, line 17: So just number of tags + 1? Do we filter release tags somehow?

gradle/play-publisher.gradle
ank27: its in ci environment of internal branch.

gradle/script-git-version.gradle
yunikkk: Right, it will increase, but since all the tags are counted and not just release ones - I assume numbers will be pretty random, e.g. if number of tags is 1, 5, 6, 9 the version names would be `0.1.1`,`0.1.5`, `0.1.6`, `0.1.9`, not that informative IMO. Could we filter release tags somehow, since we should know the release tag naming pattern?

ank27: yeah correct. it could be pretty random. but the rave team doesn't care about version, they check release notes for testing. I will try if i can filter with naming pattern. otherwise, we will keep it like this for now. :)

gradle/play-publisher.gradle
ank27: I would keep it default internal. this app will not going for production and only for internal testing.

gradle/script-git-version.gradle
alexander-kulikovskii: as an option we can use date+[time] for this version
`0.1.20222906`

ank27: done using `git.tag.list().findAll { (it.name =~ "android-v*") **}.size()`

alexander-kulikovskii (approved):

app/build.gradle.kts
```diff
@@ -5,23 +5,51 @@ plugins {
...
+      }
+      keyAlias = if (project.hasProperty("APP_KEYSTORE_ALIAS")) {
+        project.property("APP_KEYSTORE_ALIAS") as String
+      } else {
+        System.getenv("APP_KEYSTORE_ALIAS")
+      }
+      keyPassword = if (project.hasProperty("APP_KEY_PASSWORD")) {
+        project.property("APP_KEY_PASSWORD") as String
+      } else {
+        System.getenv("APP_KEY_PASSWORD")
```
yunikkk, line 31: Throw exception if any of key* resolves to empty string?

PR #1459, opened 2022-06-27
Add check for heading updates for avoid map-idle blocking

kifio:
### Summary of changes
Preventing notifying `locationConsumers` if `userHeading` changed on less than 1 degrees
I faced with issue, when event `map-idle` was never emitted on some devices. After long investigation, my colleagues and me found some clue:
 1) According to documentation `map-idle` would send `when there are no ongoing transitions and the `map` has rendered all requested non-volatile tiles`.
 2) Issue is missing out on our devices, when heading updates are disabled.

Further research has led us to fact, that the reason is constantly updates of user heading, which comes from `LocationCompassEngine.onSensorChanged`. 

For fix this problem I suppose to filter `userHeading` updates, when they are small enough to be noticeable.

## Pull request checklist:
    
### DIFF
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
@@ -18,6 +18,7 @@ import com.mapbox.maps.logW
 import com.mapbox.maps.plugin.PuckBearingSource
 import java.lang.ref.WeakReference
 import java.util.concurrent.CopyOnWriteArraySet
+import kotlin.math.abs
 
 /**
  * Default Location Provider implementation that produces location updates according to the device's
@@ -42,6 +43,7 @@ class DefaultLocationProvider @VisibleForTesting(otherwise = PRIVATE) internal c
   private var handler: Handler? = null
   private lateinit var runnable: Runnable
   private var updateDelay = INIT_UPDATE_DELAY
+  private var lastKnownUserHeading: Float = 0F
 
   private val locationEngineCallback: LocationEngineCallback<LocationEngineResult> =
     CurrentLocationEngineCallback(this)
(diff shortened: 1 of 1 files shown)


pengdev:
Hi @kifio thank you for the contribution and the detailed investigation/reasoning, these are valid points. 
We have lately merged https://github.com/mapbox/mapbox-maps-android/pull/1398 , which introduced a threshold to trigger the bearing update, the change was introduced inside the location component plugin before we send bearing updates to puck animators, which would work for both default location provider and custom location provider that users create later. Do you think that https://github.com/mapbox/mapbox-maps-android/pull/1398 will fix your issue?

For more context, https://github.com/mapbox/mapbox-maps-android/pull/1398 will be released with v10.7.0-beta.1 later this week, in the mean time, you can try to validate the fix using snapshot following the instructions [here](https://github.com/mapbox/mapbox-maps-android/blob/main/DEVELOPING.md#working-with-snapshots) and use the latest hash on the main branch `10.6.0-7deaac8ee-SNAPSHOT`.

kifio:
Hello, @pengdev .
I tried to use `10.6.0-7deaac8ee-SNAPSHOT`, but still got the issue. 
Method `updateCurrentBearing` calls continuously, but check`abs(bearings.last() - lastBearing) < BEARING_UPDATE_THRESHOLD)` always returns `false`, cause value `lastBearing` doesn't change. 
Also, differences between values of `bearings.last()` often less than `BEARING_UPDATE_THRESHOLD`, but this looks like minor problem.

pengdev:
@kifio Thanks for trying it out.
> Method `updateCurrentBearing` calls continuously, but check`abs(bearings.last() - lastBearing) < BEARING_UPDATE_THRESHOLD)` always returns `false`, cause value `lastBearing` doesn't change.

This is expected behaviour, as the check is done inside the `updateCurrentBearing` function, and `abs(bearings.last() - lastBearing) < BEARING_UPDATE_THRESHOLD)` always returns `false` means that we are returning `updateCurrentBearing` function without updating the animator and the puck properties, thus won't interrupt with the map-idle call.

> Also, differences between values of `bearings.last()` often less than `BEARING_UPDATE_THRESHOLD`, but this looks like minor problem.

We will only update the bearing if the delta is above the `BEARING_UPDATE_THRESHOLD`, do you have any suggestions on the threshold according to your experiments?

And could you please clarify the expected behaviour?

kifio:
@pengdev Thank you for clarification reply.
Let me investigate my problem with all this new info :)
I would answer your question a bit later.

PR #1477, opened 2022-07-05
Fix setting max fps and fps count for renderer

kiryldz:
### Summary of changes

This PR fixes 2 problems that are pretty close to each other.

 - when adding FPS listener via `MapView.setOnFpsChangedListener` previously callback was triggered immediately after render call was executed. That was actually not correct as docs stated that callback should return average amount of rendered frames during last second. Now this is fixed and FPS values are much more readable and reliable. Assuming we use rendering on demand (and not continuous) corner situations are also covered:
   - when not interacting with the map for some time and starting to interact after some delay - we will not report huge amount of missed VSYNC events, instead we will report last FPS update after 50 ms update of map being IDLE and reset counters.
   - when going to background we will also report last FPS update and reset counters so that when resuming the map number of VSYNC missed events will not be huge
 - when setting maximum custom FPS via `MapView.setMaximumFps` correct FPS value will be now applied. Previously without frame pacing algorithm when setting let's say 59 FPS on 60 Hz screen actual FPS was 30 (closest divider). Now we do have basic frame pacing algorithm in place that spreads moments of time when we actually wish to render the frame evenly across one second timeframe.

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -207,9 +207,6 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun setMaximumFps(fps: Int) {
-    if (fps <= 0) {
-      return
-    }
     renderer.setMaximumFps(fps)
   }
 
@@ -233,6 +230,18 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     renderer.renderThread.renderHandlerThread.post {
       renderer.renderThread.eglCore.removeRendererStateListener(rendererSetupErrorListener)
     }
+  }
+
+  internal fun setScreenRefreshRate(refreshRate: Int) {
+    if (refreshRate <= 0 || refreshRate >= MapView.MAX_POSSIBLE_FPS) {
+      logE(
+        TAG,
+        "Screen refresh rate could not be <= 0 or >= ${MapView.MAX_POSSIBLE_FPS}! " +
+          "Setting max fps and fps counter will not work properly."
+      )
+      return
+    }
+    renderer.renderThread.setScreenRefreshRate(refreshRate)
   }
 
   //
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -233,10 +233,10 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   internal fun setScreenRefreshRate(refreshRate: Int) {
-    if (refreshRate <= 0 || refreshRate >= MapView.MAX_POSSIBLE_FPS) {
+    if (refreshRate <= 0) {
       logE(
         TAG,
-        "Screen refresh rate could not be <= 0 or >= ${MapView.MAX_POSSIBLE_FPS}! " +
+        "Screen refresh rate could not be <= 0! " +
           "Setting max fps and fps counter will not work properly."
       )
       return
--- a/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
@@ -11,6 +11,7 @@ import com.mapbox.maps.renderer.MapboxSurfaceRenderer
 import com.mapbox.maps.renderer.OnFpsChangedListener
 import com.mapbox.maps.renderer.RendererSetupErrorListener
 import com.mapbox.maps.renderer.widget.Widget
+import java.lang.ref.WeakReference
 
 /**
  * A [MapSurface] provides an embeddable map interface.
@@ -29,13 +30,14 @@ import com.mapbox.maps.renderer.widget.Widget
  * @param mapInitOptions the init options to for map
  */
 class MapSurface @JvmOverloads constructor(
-  private val context: Context,
+  context: Context,
   val surface: Surface,
   mapInitOptions: MapInitOptions = MapInitOptions(context)
 ) : MapPluginProviderDelegate, MapControllable {
 
   private val mapController: MapController
   private val renderer: MapboxSurfaceRenderer = MapboxSurfaceRenderer(mapInitOptions.antialiasingSampleCount)
+  private val weakContext = WeakReference(context)
 
   init {
     this.mapController = MapController(
@@ -51,7 +53,7 @@ class MapSurface @JvmOverloads constructor(
   fun surfaceCreated() {
     renderer.surfaceCreated()
     // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
-    val screenRefreshRate = (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+    val screenRefreshRate = (weakContext.get()?.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
       .defaultDisplay?.refreshRate?.toInt() ?: MapView.DEFAULT_FPS
     mapController.setScreenRefreshRate(screenRefreshRate)
   }
--- a/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.view.MotionEvent
 import android.view.Surface
+import android.view.WindowManager
 import com.mapbox.maps.plugin.MapPlugin
 import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
 import com.mapbox.maps.renderer.MapboxSurfaceRenderer
@@ -28,7 +29,7 @@ import com.mapbox.maps.renderer.widget.Widget
  * @param mapInitOptions the init options to for map
  */
 class MapSurface @JvmOverloads constructor(
-  context: Context,
+  private val context: Context,
   val surface: Surface,
   mapInitOptions: MapInitOptions = MapInitOptions(context)
 ) : MapPluginProviderDelegate, MapControllable {
@@ -49,6 +50,11 @@ class MapSurface @JvmOverloads constructor(
    */
   fun surfaceCreated() {
     renderer.surfaceCreated()
+    mapController.setScreenRefreshRate(
+      (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+        .defaultDisplay
+        .refreshRate.toInt()
+    )
   }
 
   /**
@@ -139,9 +145,6 @@ class MapSurface @JvmOverloads constructor(
    * @param fps The maximum fps
    */
   override fun setMaximumFps(fps: Int) {
-    if (fps <= 0) {
-      return
-    }
     renderer.setMaximumFps(fps)
   }
 
--- a/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
@@ -11,7 +11,6 @@ import com.mapbox.maps.renderer.MapboxSurfaceRenderer
 import com.mapbox.maps.renderer.OnFpsChangedListener
 import com.mapbox.maps.renderer.RendererSetupErrorListener
 import com.mapbox.maps.renderer.widget.Widget
-import java.lang.ref.WeakReference
 
 /**
  * A [MapSurface] provides an embeddable map interface.
@@ -30,14 +29,14 @@ import java.lang.ref.WeakReference
  * @param mapInitOptions the init options to for map
  */
 class MapSurface @JvmOverloads constructor(
-  context: Context,
+  // could use strong ref here as MapInitOptions have strong ref in any case
+  private val context: Context,
   val surface: Surface,
   mapInitOptions: MapInitOptions = MapInitOptions(context)
 ) : MapPluginProviderDelegate, MapControllable {
 
   private val mapController: MapController
   private val renderer: MapboxSurfaceRenderer = MapboxSurfaceRenderer(mapInitOptions.antialiasingSampleCount)
-  private val weakContext = WeakReference(context)
 
   init {
     this.mapController = MapController(
@@ -53,7 +52,7 @@ class MapSurface @JvmOverloads constructor(
   fun surfaceCreated() {
     renderer.surfaceCreated()
     // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
-    val screenRefreshRate = (weakContext.get()?.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+    val screenRefreshRate = (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
       .defaultDisplay?.refreshRate?.toInt() ?: MapView.DEFAULT_FPS
     mapController.setScreenRefreshRate(screenRefreshRate)
   }
--- a/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapSurface.kt
@@ -50,11 +50,10 @@ class MapSurface @JvmOverloads constructor(
    */
   fun surfaceCreated() {
     renderer.surfaceCreated()
-    mapController.setScreenRefreshRate(
-      (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
-        .defaultDisplay
-        .refreshRate.toInt()
-    )
+    // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
+    val screenRefreshRate = (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+      .defaultDisplay?.refreshRate?.toInt() ?: MapView.DEFAULT_FPS
+    mapController.setScreenRefreshRate(screenRefreshRate)
   }
 
   /**
--- a/sdk/src/main/java/com/mapbox/maps/MapView.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapView.kt
@@ -264,7 +264,8 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   /**
    * Set new maximum FPS for map rendering.
    *
-   * @param fps new maximum FPS value that must be > 0 and less than [MAX_POSSIBLE_FPS].
+   * @param fps new maximum FPS value that must be > 0 and less than 1000.
+   * When setting this value higher than screen physically supports - max possible screen FPS rate will be used.
    */
   override fun setMaximumFps(@IntRange(from = 1L, to = MAX_POSSIBLE_FPS) fps: Int) {
     mapController.setMaximumFps(fps)
@@ -366,6 +367,11 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   companion object {
     internal const val DEFAULT_ANTIALIASING_SAMPLE_COUNT = 1
     internal const val MAX_POSSIBLE_FPS = 1000L
+
+    /**
+     * Fallback to this value if, for some reason, Android display is NULL.
+     * This will not affect rendering on displays with higher frame rate if [MapView.setMaximumFps] was not called.
+     */
     internal const val DEFAULT_FPS = 60
     /**
      * Static method to check if [MapView] could properly render on this device.
--- a/sdk/src/main/java/com/mapbox/maps/MapView.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapView.kt
@@ -113,6 +113,7 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   override fun onAttachedToWindow() {
     super.onAttachedToWindow()
     mapController.onAttachedToWindow(this)
+    mapController.setScreenRefreshRate(display.refreshRate.toInt())
   }
 
   @SuppressLint("CustomViewStyleable")
@@ -261,10 +262,10 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   /**
    * Set new maximum FPS for map rendering.
    *
-   * @param maximumFps new maximum FPS value that must be > 0.
+   * @param fps new maximum FPS value that must be > 0 and less than [MAX_POSSIBLE_FPS].
    */
-  override fun setMaximumFps(@IntRange(from = 1L, to = Int.MAX_VALUE.toLong()) maximumFps: Int) {
-    mapController.setMaximumFps(maximumFps)
+  override fun setMaximumFps(@IntRange(from = 1L, to = MAX_POSSIBLE_FPS) fps: Int) {
+    mapController.setMaximumFps(fps)
   }
 
   /**
@@ -362,6 +363,7 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    */
   companion object {
     internal const val DEFAULT_ANTIALIASING_SAMPLE_COUNT = 1
+    internal const val MAX_POSSIBLE_FPS = 1000L
     /**
      * Static method to check if [MapView] could properly render on this device.
      * This method may take some time on slow devices.
--- a/sdk/src/main/java/com/mapbox/maps/MapView.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapView.kt
@@ -113,7 +113,6 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   override fun onAttachedToWindow() {
     super.onAttachedToWindow()
     mapController.onAttachedToWindow(this)
-    mapController.setScreenRefreshRate(display.refreshRate.toInt())
   }
 
   @SuppressLint("CustomViewStyleable")
@@ -186,6 +185,9 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    * @see android.app.Fragment.onStart
    */
   override fun onStart() {
+    // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
+    val screenRefreshRate = display?.refreshRate?.toInt() ?: DEFAULT_FPS
+    mapController.setScreenRefreshRate(screenRefreshRate)
     mapController.onStart()
   }
 
@@ -364,6 +366,7 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   companion object {
     internal const val DEFAULT_ANTIALIASING_SAMPLE_COUNT = 1
     internal const val MAX_POSSIBLE_FPS = 1000L
+    internal const val DEFAULT_FPS = 60
     /**
      * Static method to check if [MapView] could properly render on this device.
      * This method may take some time on slow devices.
--- a/sdk/src/main/java/com/mapbox/maps/MapView.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapView.kt
@@ -264,10 +264,10 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   /**
    * Set new maximum FPS for map rendering.
    *
-   * @param fps new maximum FPS value that must be > 0 and less than 1000.
+   * @param fps new maximum FPS value that must be > 0 and less than max integer.
    * When setting this value higher than screen physically supports - max possible screen FPS rate will be used.
    */
-  override fun setMaximumFps(@IntRange(from = 1L, to = MAX_POSSIBLE_FPS) fps: Int) {
+  override fun setMaximumFps(@IntRange(from = 1, to = Int.MAX_VALUE.toLong()) fps: Int) {
     mapController.setMaximumFps(fps)
   }
 
@@ -366,7 +366,6 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    */
   companion object {
     internal const val DEFAULT_ANTIALIASING_SAMPLE_COUNT = 1
-    internal const val MAX_POSSIBLE_FPS = 1000L
 
     /**
      * Fallback to this value if, for some reason, Android display is NULL.
--- a/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
@@ -179,15 +179,22 @@ internal class FpsManager(
     fpsChangedListener?.let { listener ->
       val droppedFps = choreographerSkips.toDouble() / choreographerTicks
       val fps = (1.0 - droppedFps) * screenRefreshRate
-      val averageRenderTimeNs = frameRenderTimeAccumulatedNs.toDouble() / (choreographerTicks - choreographerSkips)
       listener.onFpsChanged(fps)
-      logI(
-        TAG,
-        "VSYNC based FPS is $fps," +
-          " average core rendering time is ${averageRenderTimeNs / ONE_MILLISECOND_NS} ms" +
-          " (or ${String.format("%.2f", screenRefreshPeriodNs / averageRenderTimeNs * screenRefreshRate)} FPS)," +
-          " missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
-      )
+      if (choreographerTicks == choreographerSkips) {
+        logI(
+          TAG,
+          "VSYNC based FPS is $fps, missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
+        )
+      } else {
+        val averageRenderTimeNs = frameRenderTimeAccumulatedNs.toDouble() / (choreographerTicks - choreographerSkips)
+        logI(
+          TAG,
+          "VSYNC based FPS is $fps," +
+            " average core rendering time is ${averageRenderTimeNs / ONE_MILLISECOND_NS} ms" +
+            " (or ${String.format("%.2f", screenRefreshPeriodNs / averageRenderTimeNs * screenRefreshRate)} FPS)," +
+            " missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
+        )
+      }
     }
     previousDrawnFrameIndex = 0
     frameRenderTimeAccumulatedNs = 0L
--- a/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
@@ -33,7 +33,7 @@ internal class FpsManager(
     this.screenRefreshRate = screenRefreshRate
     screenRefreshPeriodNs = ONE_SECOND_NS / screenRefreshRate
     if (userRefreshRate != USER_DEFINED_REFRESH_RATE_NOT_SET) {
-      userToScreenRefreshRateRatio = userRefreshRate.toDouble() / screenRefreshRate
+      userToScreenRefreshRateRatio = (userRefreshRate.toDouble() / screenRefreshRate).coerceIn(0.0, 1.0)
       logI(TAG, "User defined ratio is $userToScreenRefreshRateRatio")
     }
     if (LOG_STATISTICS) {
@@ -50,7 +50,7 @@ internal class FpsManager(
       userRefreshRate = refreshRate
       logI(TAG, "User set max FPS to $userRefreshRate")
       if (screenRefreshRate != SCREEN_METRICS_NOT_DEFINED) {
-        userToScreenRefreshRateRatio = userRefreshRate.toDouble() / screenRefreshRate
+        userToScreenRefreshRateRatio = (userRefreshRate.toDouble() / screenRefreshRate).coerceIn(0.0, 1.0)
         logI(TAG, "User defined ratio is $userToScreenRefreshRateRatio")
       }
     }
@@ -87,7 +87,7 @@ internal class FpsManager(
       // otherwise when updating the map after it was IDLE first update will report
       // huge delta between new frame and last frame (as we're using dirty rendering)
       handler.postDelayed(
-        VSYNC_COUNT_TILL_IDLE * (screenRefreshPeriodNs / 10.0.pow(6.0)).toLong(),
+        VSYNC_COUNT_TILL_IDLE * (screenRefreshPeriodNs / ONE_MILLISECOND_NS),
         fpsManagerToken
       ) {
         onRenderingPaused()
@@ -184,7 +184,7 @@ internal class FpsManager(
       logI(
         TAG,
         "VSYNC based FPS is $fps," +
-          " average core rendering time is ${averageRenderTimeNs / 10.0.pow(6.0)} ms" +
+          " average core rendering time is ${averageRenderTimeNs / ONE_MILLISECOND_NS} ms" +
           " (or ${String.format("%.2f", screenRefreshPeriodNs / averageRenderTimeNs * screenRefreshRate)} FPS)," +
           " missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
       )
--- a/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
@@ -1,6 +1,8 @@
 package com.mapbox.maps.renderer
 
 import android.os.Handler
+import android.os.Looper
+import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
 import androidx.core.os.postDelayed
 import com.mapbox.maps.logI
@@ -8,9 +10,8 @@ import com.mapbox.maps.logW
 import kotlin.math.pow
 
 @WorkerThread
-internal class FpsManager(
+internal class FpsManager {
   private val handler: Handler
-) {
   private var userRefreshRate = USER_DEFINED_REFRESH_RATE_NOT_SET
   private var userToScreenRefreshRateRatio = USER_TO_SCREEN_REFRESH_RATIO_NOT_SET
 
@@ -25,6 +26,15 @@ internal class FpsManager(
   private var choreographerSkips = 0
 
   internal var fpsChangedListener: OnFpsChangedListener? = null
+
+  constructor(handler: Handler) {
+    this.handler = handler
+  }
+
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  constructor() {
+    this.handler = Handler(Looper.getMainLooper())
+  }
 
   fun setScreenRefreshRate(screenRefreshRate: Int) {
     if (this.screenRefreshRate == screenRefreshRate) {
@@ -77,28 +87,24 @@ internal class FpsManager(
   }
 
   fun postRender() {
-    fpsChangedListener?.let { listener ->
-      val frameRenderTimeNs = System.nanoTime() - preRenderTimeNs
-      frameRenderTimeAccumulatedNs += frameRenderTimeNs
-      // normally we update FPS counter once a second
-      if (choreographerTicks >= screenRefreshRate) {
-        calculateFps(listener)
-      } else {
-        // however to produce correct values we also update FPS after IDLE_TIMEOUT_MS
-        // otherwise when updating the map after it was IDLE first update will report
-        // huge delta between new frame and last frame (as we're using dirty rendering)
-        handler.postDelayed(IDLE_TIMEOUT_MS, fpsManagerToken) {
-          calculateFpsAndReset(listener)
-        }
+    val frameRenderTimeNs = System.nanoTime() - preRenderTimeNs
+    frameRenderTimeAccumulatedNs += frameRenderTimeNs
+    // normally we update FPS counter and reset counters once a second
+    if (choreographerTicks >= screenRefreshRate) {
+      calculateFpsAndReset()
+    } else {
+      // however to produce correct values we also update FPS after IDLE_TIMEOUT_MS
+      // otherwise when updating the map after it was IDLE first update will report
+      // huge delta between new frame and last frame (as we're using dirty rendering)
+      handler.postDelayed(IDLE_TIMEOUT_MS, fpsManagerToken) {
+        onRenderingPaused()
       }
-      preRenderTimeNs = -1L
     }
+    preRenderTimeNs = -1L
   }
 
   fun onSurfaceDestroyed() {
-    fpsChangedListener?.let { listener ->
-      calculateFpsAndReset(listener)
-    }
+    onRenderingPaused()
   }
 
   fun destroy() {
@@ -127,9 +133,9 @@ internal class FpsManager(
     }
   }
 
-  private fun calculateFpsAndReset(listener: OnFpsChangedListener) {
+  private fun onRenderingPaused() {
     handler.removeCallbacksAndMessages(fpsManagerToken)
-    calculateFps(listener)
+    calculateFpsAndReset()
     previousFrameTimeNs = -1
   }
 
@@ -173,39 +179,36 @@ internal class FpsManager(
     return false
   }
 
-  private fun calculateFps(listener: OnFpsChangedListener) {
+  private fun calculateFpsAndReset() {
     if (choreographerTicks == 0) {
       return
     }
-    val droppedFps = choreographerSkips.toDouble() / choreographerTicks
-    val fps = (1.0 - droppedFps) * screenRefreshRate
-    val averageRenderTimeNs = frameRenderTimeAccumulatedNs.toDouble() / (choreographerTicks - choreographerSkips)
-    listener.onFpsChanged(fps)
-    logI(
-      TAG,
-      "VSYNC based FPS is $fps," +
-        " average core rendering time is ${averageRenderTimeNs / 10.0.pow(6.0)} ms" +
-        " (or ${String.format("%.2f", screenRefreshPeriodNs / averageRenderTimeNs * screenRefreshRate)} FPS)," +
-        " missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
-    )
+    fpsChangedListener?.let { listener ->
+      val droppedFps = choreographerSkips.toDouble() / choreographerTicks
+      val fps = (1.0 - droppedFps) * screenRefreshRate
+      val averageRenderTimeNs = frameRenderTimeAccumulatedNs.toDouble() / (choreographerTicks - choreographerSkips)
+      listener.onFpsChanged(fps)
+      logI(
+        TAG,
+        "VSYNC based FPS is $fps," +
+          " average core rendering time is ${averageRenderTimeNs / 10.0.pow(6.0)} ms" +
+          " (or ${String.format("%.2f", screenRefreshPeriodNs / averageRenderTimeNs * screenRefreshRate)} FPS)," +
+          " missed $choreographerSkips out of $choreographerTicks VSYNC pulses"
+      )
+    }
     previousDrawnFrameIndex = 0
     frameRenderTimeAccumulatedNs = 0L
     choreographerTicks = 0
     choreographerSkips = 0
   }
 
-  private companion object {
+  internal companion object {
     private const val TAG = "Mbgl-FpsManager"
     private const val USER_DEFINED_REFRESH_RATE_NOT_SET = -1
     private const val USER_TO_SCREEN_REFRESH_RATIO_NOT_SET = 0.0
     private const val SCREEN_METRICS_NOT_DEFINED = -1
-    private const val LOG_STATISTICS = true
-    private const val IDLE_TIMEOUT_MS = 50L
-
-    /**
-     * We use same handler and looper as our render thread and we need to make sure that
-     * we clear messages related to FPS manager only.
-     */
+    private const val LOG_STATISTICS = false
+    internal const val IDLE_TIMEOUT_MS = 50L
     private val fpsManagerToken = Any()
     private val ONE_SECOND_NS = 10.0.pow(9.0).toLong()
     private val ONE_MILLISECOND_NS = 10.0.pow(6.0).toLong()
--- a/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
@@ -12,11 +12,11 @@ internal class FpsManager(
   private val handler: Handler
 ) {
   private var userRefreshRate = USER_DEFINED_REFRESH_RATE_NOT_SET
-  private var userDefinedRatio = USER_DEFINED_RATIO_NOT_SET
+  private var userToScreenRefreshRateRatio = USER_TO_SCREEN_REFRESH_RATIO_NOT_SET
 
   private var screenRefreshRate = SCREEN_METRICS_NOT_DEFINED
   private var screenRefreshPeriodNs = -1L
-  private var previousRefreshNs = -1L
+  private var previousFrameTimeNs = -1L
   private var previousDrawnFrameIndex = 0
   private var frameRenderTimeAccumulatedNs = 0L
 
@@ -33,8 +33,8 @@ internal class FpsManager(
     this.screenRefreshRate = screenRefreshRate
     screenRefreshPeriodNs = ONE_SECOND_NS / screenRefreshRate
     if (userRefreshRate != USER_DEFINED_REFRESH_RATE_NOT_SET) {
-      userDefinedRatio = userRefreshRate.toDouble() / screenRefreshRate
-      logI(TAG, "User defined ratio is $userDefinedRatio")
+      userToScreenRefreshRateRatio = userRefreshRate.toDouble() / screenRefreshRate
+      logI(TAG, "User defined ratio is $userToScreenRefreshRateRatio")
     }
     if (LOG_STATISTICS) {
       logI(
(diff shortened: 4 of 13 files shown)


sdk/src/main/java/com/mapbox/maps/MapSurface.kt
```diff
@@ -49,6 +50,7 @@ class MapSurface @JvmOverloads constructor(
    */
   fun surfaceCreated() {
     renderer.surfaceCreated()
+    mapController.setScreenRefreshRate((context.getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay.refreshRate)
```
pengdev, line 53: How about multi-display use cases?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,180 @@
...
+import android.os.Looper
+import androidx.annotation.AnyThread
+import androidx.annotation.WorkerThread
+import androidx.core.os.postDelayed
+import com.mapbox.maps.BuildConfig
+import com.mapbox.maps.logI
+import com.mapbox.maps.logW
+import kotlin.math.pow
+
+internal class FpsManager {
```
pengdev, line 13: nit: add JvmSynthetic annotation

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
pengdev: In a typical Android Auto use case, the MapSurface is not the default display, and does not necessarily have the same refresh rate as the primary display.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: why do we need it for an internal class?

kiryldz: got it, thanks, will do

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,180 @@
...
+    } ?: logW(
+      TAG,
+      "Looper could not be found for Mapbox render thread! " +
+        "Some FPS reports may be inaccurate."
+    )
+  }
+
+  @WorkerThread
+  fun setScreenRefreshRate(screenRefreshRate: Int) {
+    deviceRefreshPeriodNs = ONE_SECOND_NS / screenRefreshRate
```
ank27, line 45: Can `screenRefreshRate` be 0? Maybe add a condition just to be safe ?

kiryldz: good point, will do

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -558,11 +550,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   }
 
   companion object {
-    private const val TAG = "Mbgl-RenderThread"
-
-    private val ONE_SECOND_NS = 10.0.pow(9.0).toLong()
-    private val ONE_MILLISECOND_NS = 10.0.pow(6.0).toLong()
-
+    internal const val TAG = "Mbgl-RenderThread"
```
alexander-kulikovskii, line 553: should it be internal?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: @pengdev `@ JvmSynthetic` could not be actually applied to whole class, any ideas how you see it should be applied then?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -450,6 +432,13 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     prepareRenderFrame(creatingSurface = true)
   }
 
+  @MainThread
+  fun setScreenRefreshRate(refreshRate: Float) {
```
alexander-kulikovskii, line 436: any reasons for `float` parameter?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: should be private, reverted

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -558,11 +550,7 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
   }
 
   companion object {
```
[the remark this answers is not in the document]
kiryldz: made internal

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,180 @@
...
+    return false
+  }
+
+  private fun calculateFps(listener: OnFpsChangedListener) {
+    val droppedFps = choreographerSkips.toDouble() / choreographerTicks
+    val fps = (1.0 - droppedFps) * deviceRefreshRate
+    val averageRenderTimeNs = frameRenderTimeAccumulatedNs.toDouble() / choreographerTicks
+    listener.onFpsChanged(fps)
+    if (BuildConfig.DEBUG) {
+      logW(
```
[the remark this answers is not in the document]
kiryldz: decided to actiually hide under static bool flag instead that will be false but could be turned locally to true to quickly understand how pacing works in particular case

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,180 @@
...
+    // no need to perform neither pacing nor FPS calculation when setMaxFps / setOnFpsChangedListener was called by the user
+    if (userDefinedRatio == 0.0 && fpsChangedListener == null) {
+      return true
+    }
+    preRenderTimeNs = System.nanoTime()
+    var skippedNow = 0
+    // check if we did miss VSYNC deadline meaning too much work was done in previous doFrame
+    if (previousRefreshNs != -1L && frameTimeNs - previousRefreshNs > deviceRefreshPeriodNs + ONE_MILLISECOND_NS) {
+      skippedNow =
+        ((frameTimeNs - previousRefreshNs) / (deviceRefreshPeriodNs + ONE_MILLISECOND_NS)).toInt()
```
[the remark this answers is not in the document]
kiryldz: discussed offline, will not happen based on business logic

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: discussed offline, should be treated separately as setting `@file:JvmSynthetic` in the beginning of the file actually does not matter for API checks (for some reason, needs investigation)

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: cause Android SDK actually returns `float`, see https://developer.android.com/reference/android/view/Display#getRefreshRate()
we use int later then, just because it's a bit easier in calculation

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+  private var choreographerSkips = 0
+
+  @Volatile
+  private var fpsChangedListener: OnFpsChangedListener? = null
+
+  private var handler: Handler? = null
+
+  @WorkerThread
+  fun prepareHandler() {
+    Looper.myLooper()?.let { looper ->
```
yunikkk, line 34: Is it valid case at all, renderThread without Looper? Shouldn't we throw an exception?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+      handler = Handler(looper)
+    } ?: logW(
+      TAG,
+      "Looper could not be found for Mapbox render thread! " +
+        "Some FPS reports may be inaccurate."
+    )
+  }
+
+  @WorkerThread
+  fun setScreenRefreshRate(screenRefreshRate: Int) {
```
yunikkk, line 44: Add https://developer.android.com/reference/kotlin/androidx/annotation/IntRange annotation?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+          calculateFps(listener)
+          previousRefreshNs = -1
+        }
+      }
+      preRenderTimeNs = -1L
+    }
+  }
+
+  @WorkerThread
+  fun onSurfaceDestroyed() {
```
alexander-kulikovskii, line 147: I guess it's not critical but nice to add `handler?.removeCallbacksAndMessages(null)`

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -104,15 +96,19 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     mapboxRenderer: MapboxRenderer,
     mapboxWidgetRenderer: MapboxWidgetRenderer,
     translucentSurface: Boolean,
-    antialiasingSampleCount: Int,
```
yunikkk, line 107: why

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -242,17 +240,14 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
   }
 
-  private fun draw() {
-    val renderTimeNsCopy = renderTimeNs
-    val currentTimeNs = SystemClock.elapsedRealtimeNanos()
-    val expectedEndRenderTimeNs = currentTimeNs + renderTimeNsCopy
-    if (expectedVsyncWakeTimeNs > currentTimeNs) {
+  private fun draw(frameTimeNanos: Long) {
+    val proceedFrameRender = fpsManager.preRender(frameTimeNanos)
+    if (!proceedFrameRender) {
```
yunikkk, line 245: Just `if (!fpsManager.preRender(frameTimeNanos)) {` ?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
alexander-kulikovskii, line 44: with `Int.MAX_VALUE` map breaks

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+    }
+    previousDrawnFrameIndex = 0
+    frameRenderTimeAccumulatedNs = 0L
+    choreographerTicks = 0
+    choreographerSkips = 0
+  }
+
+  @AnyThread
+  fun destroy() {
+    fpsChangedListener = null
```
yunikkk, line 195: Clean handler?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
yunikkk: Also - maybe just pass prepared handler from outside?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -104,15 +96,19 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
     mapboxWidgetRenderer: MapboxWidgetRenderer,
     translucentSurface: Boolean,
-    antialiasingSampleCount: Int,
+    antialiasingSampleCount: Int
   ) {
     this.translucentSurface = translucentSurface
     this.mapboxRenderer = mapboxRenderer
     this.widgetRenderer = mapboxWidgetRenderer
     this.eglCore = EGLCore(translucentSurface, antialiasingSampleCount)
+    this.fpsManager = FpsManager()
```
yunikkk, line 105: Nit : add Handler as a parameter and init after `RenderHandlerThread().apply { start()`

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+      preRenderTimeNs = -1L
+    }
+  }
+
+  @WorkerThread
+  fun onSurfaceDestroyed() {
+    fpsChangedListener?.let { listener ->
+      calculateFps(listener)
+      previousRefreshNs = -1
+    }
```
yunikkk, line 151: Clean and reset handler

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -462,14 +452,16 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
 
   @AnyThread
   fun setMaximumFps(fps: Int) {
-    renderTimeNs = ONE_SECOND_NS / fps
+    renderHandlerThread.post {
+      fpsManager.updateMaxFps(fps)
```
yunikkk, line 456: I'd name it somehow similar to `setScreenRefreshRate`, either `setScreenRefreshRate` / `setUserRefreshRate` or `setUserFps` / `setScreenFps`

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+    this.fpsChangedListener = fpsChangedListener
+  }
+
+  /**
+   * Return true if rendering should happen this frame and false otherwise.
+   *
+   * @param frameTimeNs time value from Choreographer in [System.nanoTime] timebase.
+   */
+  @WorkerThread
+  fun preRender(frameTimeNs: Long): Boolean {
```
yunikkk, line 94: `shouldRender` ?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+      )
+    }
+  }
+
+  @WorkerThread
+  fun updateMaxFps(updatedMaxFps: Int) {
+    if (userDefinedRefreshRate != updatedMaxFps) {
+      userDefinedRefreshRate = updatedMaxFps
+      logI(TAG, "User set max FPS to $userDefinedRefreshRate")
+      if (screenRefreshRate != SCREEN_METRICS_NOT_DEFINED) {
```
yunikkk, line 76: Is the otherwise possible?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,208 @@
...
+        userDefinedRatio = userDefinedRefreshRate.toDouble() / screenRefreshRate
+        logI(TAG, "User defined ratio is $userDefinedRatio")
+      }
+    }
+  }
+
+  @AnyThread
+  fun setFpsChangedListener(fpsChangedListener: OnFpsChangedListener?) {
```
yunikkk, line 84: Any reasons to not handle everything on RenderThread?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: > Is it valid case at all, renderThread without Looper? Shouldn't we throw an exception?

no, guess it's not but my idea is that `FpsManager` is generally pretty rarely used class and I don't want it be the root cause for runtime exceptions for whole render thread

 > Also - maybe just pass prepared handler from outside?

passing handler from the outside requires this handler be created while I did design `FpsManager` been created in render thread ctor so that even if user does pass max FPS value _before_ render thread instance is indeed prepared - it will be also working as expected

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: will not work properly in Java; will be adding range check in code itself here and in `setMaxFps` method as well

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: 👍

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: will revert 😞

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: good catch, will change boundaries and add code check for range (also will support Java properly then)

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: 👍

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: this will actually not init `fpsManager` as an object in ctor, see https://github.com/mapbox/mapbox-maps-android/pull/1477#discussion_r928672237

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: why not

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
kiryldz: thanks, will align

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: indeed

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: we don't need else branch, setting `userDefinedRefreshRate = updatedMaxFps` is enough then

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: if user sets fps changed listener immediately when map view is created (and render thread thread instance is not yet created) - then setting fps changed listener via `post` (in current implementation) will simply discard this this event. We can either implement it the way it's implemented now (create `FpsManager` in `MapboxRenderThread` ctor and attach handler later) or pass setting fps changed listener via user defined non-render event so that this task does not get lost. Not sure what's better here, perhaps handling everything makes more sense, will think about it

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: I like concept of `pre` / `post` render more

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: refactored it all

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+
+  private var preRenderTimeNs = -1L
+  private var choreographerTicks = 0
+  private var choreographerSkips = 0
+
+  internal var fpsChangedListener: OnFpsChangedListener? = null
+
+  fun setScreenRefreshRate(screenRefreshRate: Int) {
+    if (this.screenRefreshRate == screenRefreshRate) {
+      return
```
alexander-kulikovskii, line 31: nit:
nice to add boolean flag as a result or throw some exception

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    if (LOG_STATISTICS) {
+      logI(
+        TAG,
+        "Device screen frequency is $screenRefreshRate," +
+          " user defined refresh rate is ${if (userRefreshRate == USER_DEFINED_REFRESH_RATE_NOT_SET) "not set" else userRefreshRate}"
+      )
+    }
+  }
+
+  fun setUserRefreshRate(refreshRate: Int) {
```
alexander-kulikovskii, line 48: is it possible to set `refreshRate` to 0?
maybe check it?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -182,7 +182,10 @@ internal abstract class MapboxRenderer : MapClient {
 
   @AnyThread
   fun setMaximumFps(fps: Int) {
-    renderThread.setMaximumFps(fps)
+    if (fps <= 0 || fps >= MapView.MAX_POSSIBLE_FPS) {
```
alexander-kulikovskii, line 185:
```suggestion
    if (fps <= 0 || fps > MapView.MAX_POSSIBLE_FPS) {
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -226,6 +223,18 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
   }
 
+  internal fun setScreenRefreshRate(refreshRate: Int) {
+    if (refreshRate <= 0 || refreshRate >= MapView.MAX_POSSIBLE_FPS) {
```
alexander-kulikovskii, line 227: but not critical :)
```suggestion
    if (refreshRate <= 0 || refreshRate > MapView.MAX_POSSIBLE_FPS) {
```

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: we have a check for zero before we get here

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: why? if we did get here - it means that device notified again that screen refresh rate has arrived (which happens e.g. if we minimize / maximize app). It's valid situation that we simply return.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
kiryldz: don't see difference, added warning log messages for those situations

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
kiryldz: @pengdev I did not manage to actually simulate primary display with one refresh rate and secondary with another but I did make sure that applying different user refresh rates (via `setMaximumFps`) to map on primary and secondary displays works fine. Guess that's all we could do now, if some extraordinary case will pop-up (although it's indeed very specific use-case but it should be handled).

Also I did check via our `SurfaceActivity` that methods `mapSurface.setMaximumFps` and `mapSurface.setOnFpsChangedListener` do work as expected.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    private const val USER_DEFINED_REFRESH_RATE_NOT_SET = -1
+    private const val USER_DEFINED_RATIO_NOT_SET = 0.0
+    private const val SCREEN_METRICS_NOT_DEFINED = -1
+    private const val LOG_STATISTICS = true
+    private const val IDLE_TIMEOUT_MS = 50L
+
+    /**
+     * We use same handler and looper as our render thread and we need to make sure that
+     * we clear messages related to FPS manager only.
+     */
```
yunikkk, line 181: Huh?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -226,6 +223,18 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+      logE(
+        TAG,
+        "Screen refresh rate could not be <= 0 or >= ${MapView.MAX_POSSIBLE_FPS}! " +
```
yunikkk, line 230: Nit - since user calls setMaxFPS - I'd written `Max fps could not ...`

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+import androidx.core.os.postDelayed
+import com.mapbox.maps.logI
+import com.mapbox.maps.logW
+import kotlin.math.pow
+
+@WorkerThread
+internal class FpsManager(
+  private val handler: Handler
+) {
+  private var userRefreshRate = USER_DEFINED_REFRESH_RATE_NOT_SET
```
yunikkk, line 14: Null seems more natural for `non-set` values, but up to you

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+  private var userDefinedRatio = USER_DEFINED_RATIO_NOT_SET
+
+  private var screenRefreshRate = SCREEN_METRICS_NOT_DEFINED
```
yunikkk, line 17: Nit : comment describing of the FPS calculation / management would be helpful

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    // clear any scheduled task as new render call is about to happen
+    handler.removeCallbacksAndMessages(fpsManagerToken)
+    // no need to perform neither pacing nor FPS calculation when setMaxFps / setOnFpsChangedListener was not called by the user
+    if (userDefinedRatio == USER_DEFINED_RATIO_NOT_SET && fpsChangedListener == null) {
+      return true
+    }
+    preRenderTimeNs = System.nanoTime()
+    var skippedNow = 0
+    // check if we did miss VSYNC deadline meaning too much work was done in previous doFrame
+    if (previousRefreshNs != -1L && frameTimeNs - previousRefreshNs > screenRefreshPeriodNs + ONE_MILLISECOND_NS) {
```
yunikkk, line 74: Nit - rename `previousRefreshNs` to `previousFrameTimeNs`?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+      skippedNow =
+        ((frameTimeNs - previousRefreshNs) / (screenRefreshPeriodNs + ONE_MILLISECOND_NS)).toInt()
```
yunikkk, line 76: I'd split the function for readability, e.g. 
```
updateFpsStats()
if (userDefinedRatio != USER_DEFINED_RATIO_NOT_SET) {
  return performPacing()
}
```

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+      return performPacing()
+    }
+    return true
+  }
+
+  fun postRender() {
+    fpsChangedListener?.let { listener ->
+      val frameRenderTimeNs = System.nanoTime() - preRenderTimeNs
+      frameRenderTimeAccumulatedNs += frameRenderTimeNs
+      // normally we update FPS counter once a second
```
yunikkk, line 99: Where is this `every second` checked?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    choreographerTicks = 0
+    choreographerSkips = 0
+  }
+
+  private companion object {
+    private const val TAG = "Mbgl-FpsManager"
+    private const val USER_DEFINED_REFRESH_RATE_NOT_SET = -1
+    private const val USER_DEFINED_RATIO_NOT_SET = 0.0
+    private const val SCREEN_METRICS_NOT_DEFINED = -1
+    private const val LOG_STATISTICS = true
```
yunikkk, line 175: Should be off by default? Also - I'd made it param of the consturctor

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
```diff
@@ -29,10 +29,12 @@ internal class RenderHandlerThread {
...
   }
 
-  fun start() {
-    handlerThread = HandlerThread(HANDLE_THREAD_NAME, THREAD_PRIORITY_DISPLAY).apply {
-      start()
-      handler = Handler(this.looper)
+  fun start(): Handler {
+    handlerThread = HandlerThread(HANDLE_THREAD_NAME, THREAD_PRIORITY_DISPLAY)
+    handlerThread.start()
+    Handler(handlerThread.looper).also {
```
yunikkk, line 35: Nit : 
```
return Handler(handlerThread.looper).also {
      handler = it
}
```

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: primitive type could not be null

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: I'll re-phrase a bit

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: user will not have control over it, it's more for us as adding logs for fps is not that easy and I do not want to actually do this again if needed

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
yunikkk: Yeah but you can use Kotlin wrappers right

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+import com.mapbox.maps.logI
+import com.mapbox.maps.logW
+import kotlin.math.pow
+
+@WorkerThread
+internal class FpsManager(
+  private val handler: Handler
+) {
+  private var userRefreshRate = USER_DEFINED_REFRESH_RATE_NOT_SET
+  private var userDefinedRatio = USER_DEFINED_RATIO_NOT_SET
```
yunikkk, line 15: Nit : `userToScreenRefreshRateRatio` ?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
yunikkk: Yes, I understand that, just wanted to highlight that if this var is used to turn smth on/off or change the behaviour of the class - this variable could be part of the constructor. Up to you though

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,186 @@
...
+    fpsChangedListener = null
+  }
+
+  private fun calculateFpsAndReset(listener: OnFpsChangedListener) {
+    handler.removeCallbacksAndMessages(fpsManagerToken)
+    calculateFps(listener)
+    previousRefreshNs = -1
+  }
+
+  private fun performPacing(): Boolean {
```
yunikkk, line 131: I would expect we could write tests for the every frame and simulate throttle / frame skips with low / high fps, right? Do you plan to write tests for the FpsManager itself?

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
```diff
@@ -98,9 +98,16 @@ internal abstract class MapboxRendererTest {
...
-  fun setMaximumFpsTest() {
+  fun setValidMaximumFpsTest() {
     mapboxRenderer.setMaximumFps(10)
-    verify { renderThread.setMaximumFps(10) }
+    verify { renderThread.setUserRefreshRate(10) }
+  }
+
+  @Test
+  fun setInvalidMaximumFpsTest() {
+    mapboxRenderer.setMaximumFps(-1)
```
alexander-kulikovskii, line 108: it should be two separate tests

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
```diff
@@ -98,9 +98,16 @@ internal abstract class MapboxRendererTest {
...
+    mapboxRenderer.setMaximumFps(MapView.MAX_POSSIBLE_FPS.toInt() + 1)
+    verify(exactly = 0) { renderThread.setUserRefreshRate(any()) }
```
alexander-kulikovskii, line 110: nit:
```suggestion
    verifyNo { renderThread.setUserRefreshRate(any()) }
```

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -260,9 +261,12 @@ class MapControllerTest {
 
   @Test
   fun setMaximumFpsInvalid() {
+    every { mockRenderer.setMaximumFps(any()) } just Runs
+
     testMapController.setMaximumFps(-1)
 
-    verify(exactly = 0) { mockRenderer.setMaximumFps(any()) }
+    // range check is performed within MapboxRenderer
+    verify { mockRenderer.setMaximumFps(any()) }
```
alexander-kulikovskii, line 269:
```suggestion
    verifyNo { mockedRenderThread.setUserRefreshRate(any()) }
```

kiryldz: it should be verify here, I did even leave a comment

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
kiryldz: I'm using verify with 0 counter, imo it's ok to leave as is, don't see a big issue here

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: I like it current way more.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: I'll drop it somewhere, sure

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: nice, indeed

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: cause `every second` is actually when number of VSYNC impulses is equal to given screen frequency. So when we have 60 choreographer ticks - it means that 1 second did pass.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: didn't think of anything for now, as for now we should just always keep it off. Main idea was that we might want to revisit pacing algorithm at some point and then those logs will help a lot.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: nice!

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: yeah, it is planned

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: alright

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: just removed, actually it's pretty clear why it is needed

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,216 @@
...
+  constructor() {
+    this.handler = Handler(Looper.getMainLooper())
+  }
+
+  fun setScreenRefreshRate(screenRefreshRate: Int) {
+    if (this.screenRefreshRate == screenRefreshRate) {
+      return
+    }
+    this.screenRefreshRate = screenRefreshRate
+    screenRefreshPeriodNs = ONE_SECOND_NS / screenRefreshRate
```
Chaoba, line 44: Could `screenRefreshRate` be 0 here?

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -311,4 +314,32 @@ class MapControllerTest {
...
+    unmockkStatic("com.mapbox.maps.MapboxLogger")
+  }
+
+  @Test
+  fun setInvalidScreenRefreshRate() {
+    mockkStatic("com.mapbox.maps.MapboxLogger")
+    every { logE(any(), any()) } just Runs
+    val mockRenderThread = mockk<MapboxRenderThread>(relaxed = true)
+    every { mockRenderer.renderThread } returns mockRenderThread
+    testMapController.setScreenRefreshRate(-1)
```
Chaoba, line 338: Please add a case for  `testMapController.setScreenRefreshRate(0)`

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -362,6 +365,8 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    */
   companion object {
     internal const val DEFAULT_ANTIALIASING_SAMPLE_COUNT = 1
+    internal const val MAX_POSSIBLE_FPS = 1000L
```
pengdev, line 368: why 1000L?

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -261,10 +264,10 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   /**
    * Set new maximum FPS for map rendering.
    *
-   * @param maximumFps new maximum FPS value that must be > 0.
+   * @param fps new maximum FPS value that must be > 0 and less than [MAX_POSSIBLE_FPS].
```
pengdev, line 267: since `MAX_POSSIBLE_FPS` is internal, probably this is not visible to user?

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
```diff
@@ -49,6 +50,10 @@ class MapSurface @JvmOverloads constructor(
    */
   fun surfaceCreated() {
     renderer.surfaceCreated()
+    // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
+    val screenRefreshRate = (context.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
+      .defaultDisplay?.refreshRate?.toInt() ?: MapView.DEFAULT_FPS
+    mapController.setScreenRefreshRate(screenRefreshRate)
```
pengdev, line 56: Is there other way to get the refresh rate? the MapSurface is also used for Android Auto, I'm not sure we should always take the default display's refresh rate.

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
```diff
@@ -28,7 +29,7 @@ import com.mapbox.maps.renderer.widget.Widget
  * @param mapInitOptions the init options to for map
  */
 class MapSurface @JvmOverloads constructor(
-  context: Context,
+  private val context: Context,
```
pengdev, line 32: nit: shall we use weak reference?

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
kiryldz: Could you please clarify a bit where should we get default screen refresh rate from if not from an Android service? I'm not that familiar with AA setup.
IMO we could always expose (if needed) new method on `MapSurface`, e.g. `MapSurface#setScreenRefreshRate()` to allow control that from outside but for now I doubt doing that as the setup is not fully clear to me

sdk/src/main/java/com/mapbox/maps/MapView.kt
kiryldz: just some big value (who knows what screens will appear on devices in the nearest future) - but not big enough (as let's say MAX_INT that breaks the code). Setting 1000 as max fps will actually simply mean that device will render with max possible speed (most likely 60, 90 or 120 fps).

sdk/src/main/java/com/mapbox/maps/MapView.kt
kiryldz: hard to say however imo it's not that important, general idea should be clean enough

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
kiryldz: yeah, I think it's worth it, will do

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
kiryldz: ok, will do

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
kiryldz: check for 0 performed before in `MapController#setScreenRefreshRate`

sdk/src/main/java/com/mapbox/maps/MapView.kt
pengdev: Why do we even need a max possible fps if the maximum is capped with the device's supported refresh rate? The argument is long type, I guess that's enough already?

sdk/src/main/java/com/mapbox/maps/MapView.kt
pengdev: I mean this documentation uses an internal var, which means the generated doc will have no reference to this value, without source code, the `MAX_POSSIBLE_FPS` is meaningless.

sdk/src/main/java/com/mapbox/maps/MapView.kt
kiryldz: cause user could potentially pass MAX_INT and break all the things, that's why I did add some limits

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -453,6 +451,13 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
     prepareRenderFrame(creatingSurface = true)
   }
 
+  @MainThread
```
yunikkk, line 454: `@UiThread` is used below, align?

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderer.kt
```diff
@@ -182,7 +182,11 @@ internal abstract class MapboxRenderer : MapClient {
...
+      logE(TAG, "Maximum FPS could not be <= 0 or >= ${MapView.MAX_POSSIBLE_FPS}")
```
yunikkk, line 185:
```suggestion
      logE(TAG, "Maximum FPS could not be <= 0 or >= ${MapView.MAX_POSSIBLE_FPS}, ignoring $fps value")
```

sdk/src/test/java/com/mapbox/maps/renderer/FpsManagerTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    pauseHandler()
+    val frameTotalNumber = 10
+    var frameRenderedCount = 0
+    for (i in 0 until frameTotalNumber) {
+      Choreographer.getInstance().postFrameCallback {
+        assert(fpsManager.preRender(it))
+        frameRenderedCount++
+      }
+    }
+    idleHandler(vsyncCount = frameTotalNumber)
```
yunikkk, line 48: It seems that all the `postFrameCallback` will be executed on the first Choreographer tick, since you post it before idling handler and all the runnables are executed together on the first onFrame callback. 
Checked the test locally and can confirm that `dleHandler(vsyncCount = 1)` still passes.

sdk/src/main/java/com/mapbox/maps/renderer/MapboxRenderThread.kt
```diff
@@ -86,8 +77,27 @@ internal class MapboxRenderThread : Choreographer.FrameCallback {
...
+    new?.let {
+      if (old != it) {
+        /**
+         * Consider setting [OnFpsChangedListener] as non-render event to make sure that
+         * it does not get dropped if user has set it right after MapView creation before render thread
+         * is actually fully prepared for rendering.
+         */
+        postNonRenderEvent(
+          RenderEvent(
+            { fpsManager.fpsChangedListener = it },
```
yunikkk, line 91: Does this mean previous `fpsManager.fpsChangedListener` will remain if null is set?

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,216 @@
...
+
+  internal var fpsChangedListener: OnFpsChangedListener? = null
+
+  constructor(handler: Handler) {
+    this.handler = handler
+  }
+
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  constructor() {
+    this.handler = Handler(Looper.getMainLooper())
```
yunikkk, line 36: Cant this handler just be passed from tests?

sdk/src/test/java/com/mapbox/maps/renderer/MapboxRendererTest.kt
```diff
@@ -98,9 +98,27 @@ internal abstract class MapboxRendererTest {
...
-  fun setMaximumFpsTest() {
+  fun setValidMaximumFpsTest() {
     mapboxRenderer.setMaximumFps(10)
-    verify { renderThread.setMaximumFps(10) }
+    verify(exactly = 1) { renderThread.setUserRefreshRate(10) }
+  }
+
+  @Test
+  fun setMoreThanMaximumFpsTest() {
+    mapboxRenderer.setMaximumFps(MapView.MAX_POSSIBLE_FPS.toInt() + 1)
```
alexander-kulikovskii, line 108: it's first not allowed value
```suggestion
    mapboxRenderer.setMaximumFps(MapView.MAX_POSSIBLE_FPS.toInt())
```

sdk/src/test/java/com/mapbox/maps/MapControllerTest.kt
```diff
@@ -376,4 +379,36 @@ class MapControllerTest {
...
+  }
+
+  @Test
+  fun setNegativeScreenRefreshRate() {
+    screenRefreshRateTest(0, -1)
+  }
+
+  @Test
+  fun setMoreThanMaxScreenRefreshRate() {
+    screenRefreshRateTest(0, MapView.MAX_POSSIBLE_FPS.toInt() + 1)
```
alexander-kulikovskii, line 400:
```suggestion
    screenRefreshRateTest(0, MapView.MAX_POSSIBLE_FPS.toInt())
```

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,213 @@
...
+    frameRenderTimeAccumulatedNs += frameRenderTimeNs
+    // normally we update FPS counter and reset counters once a second
+    if (choreographerTicks >= screenRefreshRate) {
+      calculateFpsAndReset()
+    } else {
+      // however to produce correct values we also update FPS after IDLE_TIMEOUT_MS
+      // otherwise when updating the map after it was IDLE first update will report
+      // huge delta between new frame and last frame (as we're using dirty rendering)
+      handler.postDelayed(
+        VSYNC_COUNT_TILL_IDLE * (screenRefreshPeriodNs / 10.0.pow(6.0)).toLong(),
```
alexander-kulikovskii, line 90: no need to calculate it each time
```suggestion
        VSYNC_COUNT_TILL_IDLE * (screenRefreshPeriodNs / ONE_MILLISECOND_NS).toLong(),
```

alexander-kulikovskii (approved):

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -261,10 +264,11 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   /**
    * Set new maximum FPS for map rendering.
    *
-   * @param maximumFps new maximum FPS value that must be > 0.
+   * @param fps new maximum FPS value that must be > 0 and less than 1000.
+   * When setting this value higher than screen physically supports - max possible screen FPS rate will be used.
    */
-  override fun setMaximumFps(@IntRange(from = 1L, to = Int.MAX_VALUE.toLong()) maximumFps: Int) {
-    mapController.setMaximumFps(maximumFps)
+  override fun setMaximumFps(@IntRange(from = 1L, to = MAX_POSSIBLE_FPS) fps: Int) {
```
pengdev, line 270: can we keep remove the `MAX_POSSIBLE_FPS` and use `Int.MAX_VALUE` instead? it feels pretty weird to have 1000 here, actually tested with `Int.MAX_VALUE` and the map renders pretty ok on my device.

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,213 @@
...
+  internal var fpsChangedListener: OnFpsChangedListener? = null
+
+  fun setScreenRefreshRate(screenRefreshRate: Int) {
+    if (this.screenRefreshRate == screenRefreshRate) {
+      return
+    }
+    this.screenRefreshRate = screenRefreshRate
+    screenRefreshPeriodNs = ONE_SECOND_NS / screenRefreshRate
+    if (userRefreshRate != USER_DEFINED_REFRESH_RATE_NOT_SET) {
+      userToScreenRefreshRateRatio = userRefreshRate.toDouble() / screenRefreshRate
```
pengdev, line 36: In the doc, we mentioned 
> When setting this value higher than screen physically supports - max possible screen FPS rate will be used.
So in theory the `userToScreenRefreshRateRatio` shouldn't exceed 1.0, maybe we can cap the `userToScreenRefreshRateRatio` below 1.0?

```suggestion
      userToScreenRefreshRateRatio = (userRefreshRate.toDouble() / screenRefreshRate).coerceIn(0.0, 1.0)
```

sdk/src/main/java/com/mapbox/maps/renderer/FpsManager.kt
```diff
@@ -0,0 +1,213 @@
...
+      )
+    }
+  }
+
+  fun setUserRefreshRate(refreshRate: Int) {
+    if (userRefreshRate != refreshRate) {
+      userRefreshRate = refreshRate
+      logI(TAG, "User set max FPS to $userRefreshRate")
+      if (screenRefreshRate != SCREEN_METRICS_NOT_DEFINED) {
+        userToScreenRefreshRateRatio = userRefreshRate.toDouble() / screenRefreshRate
```
pengdev, line 53: Similarly here:

```suggestion
        userToScreenRefreshRateRatio = (userRefreshRate.toDouble() / screenRefreshRate).coerceIn(0.0, 1.0)
```

sdk/src/main/java/com/mapbox/maps/MapSurface.kt
```diff
@@ -49,6 +52,10 @@ class MapSurface @JvmOverloads constructor(
    */
   fun surfaceCreated() {
     renderer.surfaceCreated()
+    // display should not be null at this point but to be sure we will fallback to DEFAULT_FPS
+    val screenRefreshRate = (weakContext.get()?.getSystemService(Context.WINDOW_SERVICE) as WindowManager)
```
yunikkk, line 56: Btw, wonder can this weak reference be released here? I mean, could MapSurface be detached and attached to different activity?

kiryldz: reverted to strong as strong ref is used in `MapInitOptions` anyway

sdk/src/test/java/com/mapbox/maps/renderer/FpsManagerTest.kt
```diff
@@ -77,6 +77,42 @@ class FpsManagerTest {
...
+  @Test
+  fun userRefreshRateMaxIntTest() {
+    fpsManager.setUserRefreshRate(Int.MAX_VALUE)
+    pauseHandler()
+    val frameRenderedPattern = mutableListOf<Boolean>()
+    idleHandler(vsyncCount = 10) {
+      frameRenderedPattern.add(fpsManager.preRender(it))
+    }
+    Assert.assertArrayEquals(
+      arrayOf(
```
yunikkk, line 108: `Array(10) { true }`

sdk/src/test/java/com/mapbox/maps/renderer/FpsManagerTest.kt
```diff
@@ -0,0 +1,196 @@
...
+  @Test
+  fun userRefreshRate1Test() {
+    fpsManager.setUserRefreshRate(1)
+    pauseHandler()
+    val vsyncCount = 120
+    val frameRenderedPattern = mutableListOf<Boolean>()
+    val expectedRenderedPattern = BooleanArray(vsyncCount)
+    for (i in 0 until 120) {
+      expectedRenderedPattern[i] = i == 59 || i == 119
+    }
```
yunikkk, line 89: `Array(vsyncCount) { it == 59 || it == 119 }`

sdk/src/test/java/com/mapbox/maps/renderer/FpsManagerTest.kt
```diff
@@ -0,0 +1,196 @@
...
+  @Test
+  fun userRefreshRate7Test() {
+    fpsManager.setUserRefreshRate(7)
+    pauseHandler()
+    val frameRenderedPattern = mutableListOf<Boolean>()
+    idleHandler(vsyncCount = 18) {
+      frameRenderedPattern.add(fpsManager.preRender(it))
+    }
+    Assert.assertArrayEquals(
+      arrayOf(
```
yunikkk, line 72: `Array(vsyncCount) { it == 9 || it == 18 }`

yunikkk: Hmm, and why 9/18 if FPS is 7?

yunikkk: Overall I would write those tests to be at least 1 or 2 seconds, e.g. use vsyncCount = 120 everywhere

sdk/src/main/java/com/mapbox/maps/renderer/RenderHandlerThread.kt
```diff
@@ -11,7 +11,6 @@ internal class RenderHandlerThread {
 
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal lateinit var handlerThread: HandlerThread
-  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal var handler: Handler? = null
```
yunikkk, line 14:
```suggestion
  private var handler: Handler? = null
```
?

kiryldz: It is accessed from map controller, so can't be private

yunikkk (approved):

Chaoba (approved):

sdk/src/test/java/com/mapbox/maps/renderer/FpsManagerTest.kt
kiryldz: > Hmm, and why 9/18 if FPS is 7?

A bit of math - to achieve 7 FPS on 60 Hz screen we have following ratio: 7 / 60 = 0.11(6) => thus we render new frame roughly each 9th frame. I don't see much sense in adding bigger vsync interval but just for you I'll add one additional test with some strange rate and with range of 120 frames.

PR #1552, opened 2022-08-01
Fix onAnnotationDragStarted event is still fired when annotation is not draggable - MAPSAND-356

Chaoba:
### Summary of changes
AnnotationManager fires `onAnnotationDragStarted` events even the annotation is not draggable.  In this pr, we will check the `isDraggable` property when start dragging and return directly if the annotation is not draggable so that `onAnnotationDragStarted` event will not be fired.

## Pull request checklist:
    
### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
+++ b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
@@ -714,6 +714,9 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     }
 
     private fun startDragging(annotation: T): Boolean {
+      if (!annotation.isDraggable) {
+        return false
+      }
       dragListeners.forEach { it.onAnnotationDragStarted(annotation) }
       draggingAnnotation = annotation
       return true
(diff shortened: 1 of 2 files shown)


ank27 (approved):

plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt
```diff
@@ -714,6 +714,9 @@ abstract class AnnotationManagerImpl<G : Geometry, T : Annotation<G>, S : Annota
     }
 
     private fun startDragging(annotation: T): Boolean {
+      if (!annotation.isDraggable) {
```
yunikkk, line 717: @Chaoba the only place where `startDragging` is called from is `onMoveBegin`, which also runs `queryMapForFeatures` before. Could we check `isDraggable` before and skip the query render features call then? Will save couple thread switches

yunikkk: E.g. [here ](https://github.com/mapbox/mapbox-maps-android/blob/4a633d6c118424ffa852dd6fde158ab48ddbb10b/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationManagerImpl.kt#L647)

Chaoba: Only after `queryMapForFeatures` can we get the annotation and know whether it's draggable or not.

yunikkk: Oh, indeed, somehow missed the obvious part

yunikkk (approved):

PR #1565, opened 2022-08-08
Fix the git release tag matching for CI/CD.

pengdev:
### Summary of changes

Fix the git release tag matching for continuous deployment.

## Pull request checklist:
    
### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ parameters:
     default: "owner/repo"
   git_release_tag:
     type: string
-    default: v10\.[0-9]*\.[0-9]*$
+    default: ^v[0-9]+\.[0-9]+\.[0-9]+.*$
 
 workflows:
   version: 2
@@ -194,7 +194,7 @@ commands:
       - run:
           name: Update version name
           command: |
-            if [[ $CIRCLE_TAG == << pipeline.parameters.git_release_tag >> ]]; then
+            if [[ $CIRCLE_TAG =~ << pipeline.parameters.git_release_tag >> ]]; then
               sed -i -e "s/^VERSION_NAME=.*/VERSION_NAME=${CIRCLE_TAG:1}/" gradle.properties
             elif [[ $CIRCLE_TAG == extension-androidauto-v* ]]; then
               sed -i -e "s/^MODULE_VERSION_NAME=.*/MODULE_VERSION_NAME=${CIRCLE_TAG:23}/" extension-androidauto/gradle.properties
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ parameters:
     default: "owner/repo"
   git_release_tag:
     type: string
-    default: v10\.[0-9]*\.[0-9]*$
+    default: v[0-9]+\.[0-9]+\.[0-9]+[\S]*$
```
alexander-kulikovskii, line 14: nice fix
maybe add only allowed words?
because such tag also will work `v10.4.4-_+beta..rc-123`

pengdev: I think we shouldn't limit the string after the version name, as we may want to introduce other variants of version name, e.g. alpha. 
And matching the initial part of the version name also matches the previous behavior.

yunikkk: Do we need to add regex for the starting part `v10.` ? It'll match `v0.` for example.

.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ commands:
       - run:
           name: Start internal CI release flow
           command: |
-            if [[ $CIRCLE_TAG == << pipeline.parameters.git_release_tag >> ]]; then
+            if [[ $CIRCLE_TAG =~ << pipeline.parameters.git_release_tag >> ]]; then
```
yunikkk, line 211: Hmm, but how did it work previously? We've had regex for quite a long time it seems

yunikkk: Ah, or it had worked with simple wildcard but when we've switched to the full regexps it stopped right

.circleci/config.yml
pengdev: I guess it's our responsibility to cut the release with correct version name? I think the expression here is primarily used for distinguishing the release tags from the other tags we might create?

.circleci/config.yml
pengdev: Need to run some real-word test though.. probably try it out in the next beta release.

alexander-kulikovskii (approved):

Chaoba (approved):

.circleci/config.yml
yunikkk: Yeah, however why allow the incorrect and broad regex patterns? E.g. would `v10\.[0-9]+\.[0-9]+[\S]*$` be more strict and better?

pengdev: We will eventually have version 11 and so on, it's easy to miss the ci config when releasing new versions?

yunikkk: Hmm, from this point of view... Not sure that we won't change CI on the major release update, but ok, on with leaving as is

yunikkk (approved):

ank27 (approved):

Issue #1019, opened 2022-01-03
Crash in StandardGestureListener.onFling


florianPOLARSTEPS:
## Environment

- Android OS version: 9,10,11,12
- Devices affected: irrelevant
- Maps SDK Version: 10.2.0

## Observed behavior and steps to reproduce

I am not able to reproduce this issue but it seems to be an issue with the inherited nullability of MotionEvent fields from the Android framework. (android.view.GestureDetector.GestureListener)

It may be wise to assume that all MotionEvents in the GestureDetector.OnGestureListener callbacks can be null.

```
Fatal Exception: java.lang.NullPointerException: Parameter specified as non-null is null: method com.mapbox.maps.plugin.gestures.GesturesPluginImpl$StandardGestureListener.onFling, parameter e1
       at com.mapbox.maps.plugin.gestures.GesturesPluginImpl$StandardGestureListener.onFling(GesturesPluginImpl.kt:2)
       at com.mapbox.android.gestures.StandardGestureDetector$1.onFling(StandardGestureDetector.java:56)
       at android.view.GestureDetector.onTouchEvent(GestureDetector.java:672)
       at androidx.core.view.GestureDetectorCompat$GestureDetectorCompatImplJellybeanMr2.onTouchEvent(GestureDetectorCompat.java:480)
       at androidx.core.view.GestureDetectorCompat.onTouchEvent(GestureDetectorCompat.java:543)
       at com.mapbox.android.gestures.StandardGestureDetector.analyzeEvent(StandardGestureDetector.java:147)
       at com.mapbox.maps.plugin.gestures.GesturesPluginImpl.<init>(GesturesPluginImpl.kt:61)
       at com.mapbox.android.gestures.BaseGesture.analyze(BaseGesture.java:61)
       at com.mapbox.android.gestures.BaseGesture.onTouchEvent(BaseGesture.java:39)
       at com.mapbox.android.gestures.AndroidGesturesManager.onTouchEvent(AndroidGesturesManager.java:193)
       at com.mapbox.maps.plugin.gestures.GesturesPluginImpl.onTouchEvent(GesturesPluginImpl.kt:268)
       at com.mapbox.maps.plugin.MapPluginRegistry.onTouch(MapPluginRegistry.kt:128)
       at com.mapbox.maps.MapController.onTouchEvent(MapController.kt:167)
       at com.mapbox.maps.MapView.onTouchEvent(MapView.kt:289)
[...]

PR #1608, opened 2022-08-17
Fix possible crash if null touch event is delivered.

yunikkk:
### Summary of changes

Fixes crash due to the null MotionEvent passed to Kotlin non-nullable var. 
Reported by some users at https://github.com/mapbox/mapbox-maps-android/issues/1019.

## Pull request checklist:
    
### DIFF
--- a/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
+++ b/plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
@@ -15,6 +15,7 @@ import androidx.annotation.VisibleForTesting.PRIVATE
 import androidx.core.animation.addListener
 import androidx.interpolator.view.animation.LinearOutSlowInInterpolator
 import com.mapbox.android.gestures.*
+import com.mapbox.android.gestures.StandardGestureDetector.StandardOnGestureListener
 import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.ScreenCoordinate
 import com.mapbox.maps.StylePropertyValueKind
@@ -29,15 +30,10 @@ import com.mapbox.maps.plugin.animation.CameraAnimatorOptions.Companion.cameraAn
 import com.mapbox.maps.plugin.animation.MapAnimationOptions
 import com.mapbox.maps.plugin.animation.MapAnimationOptions.Companion.mapAnimationOptions
 import com.mapbox.maps.plugin.animation.MapAnimationOwnerRegistry
-import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
-import com.mapbox.maps.plugin.delegates.MapDelegateProvider
-import com.mapbox.maps.plugin.delegates.MapPluginProviderDelegate
-import com.mapbox.maps.plugin.delegates.MapProjectionDelegate
-import com.mapbox.maps.plugin.delegates.MapTransformDelegate
+import com.mapbox.maps.plugin.delegates.*
 import com.mapbox.maps.plugin.gestures.generated.GesturesAttributeParser
 import com.mapbox.maps.plugin.gestures.generated.GesturesSettings
 import com.mapbox.maps.plugin.gestures.generated.GesturesSettingsBase
-import java.util.*
 import java.util.concurrent.CopyOnWriteArraySet
 import kotlin.math.*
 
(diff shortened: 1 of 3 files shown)


plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -454,19 +506,25 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
...
      */
     override fun onFling(
-      e1: MotionEvent,
-      e2: MotionEvent,
+      e1: MotionEvent?,
+      e2: MotionEvent?,
       velocityX: Float,
       velocityY: Float
     ): Boolean {
+      if (e1 == null || e2 == null) {
```
kiryldz, line 525: `e1` is not even used in `handleFlingEvent` - thus we can make `handleFlingEvent` take 3 params instead of 4 and also not check for `e1 == null` then

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -415,37 +411,93 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
     centerScreen = ScreenCoordinate((width / 2).toDouble(), (height / 2).toDouble())
   }
 
+  /**
+   * Used instead of [com.mapbox.android.gestures.StandardGestureDetector.SimpleStandardOnGestureListener]
+   * since it adds NonNull annotations while its parent [android.view.GestureDetector.OnGestureListener] does not
+   * have those annotations in the release Android versions (added recently at https://android.googlesource.com/platform/frameworks/base/+/68c65e58b4f4daf79c9ffab518a826a506799db2/core/java/android/view/GestureDetector.java).
+    */
+  private open class SimpleStandardOnGestureListener : StandardOnGestureListener {
```
kiryldz, line 419: could similar approach be used for other listeners from that library (if any)?

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -415,37 +411,93 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
...
-  private inner class StandardGestureListener internal constructor(private val doubleTapMovementThreshold: Float) :
-    StandardGestureDetector.SimpleStandardOnGestureListener() {
+  private inner class StandardGestureListener(private val doubleTapMovementThreshold: Float) :
+    SimpleStandardOnGestureListener() {
 
     /**
      * Called when an on down gesture was detected.
      */
-    override fun onDown(motionEvent: MotionEvent): Boolean {
+    override fun onDown(motionEvent: MotionEvent?): Boolean {
```
kiryldz, line 472: why don't return `true` in superclass and not even overwrite it here then?

kiryldz:
Guess couple of unit tests will be suitable here to check for nullable values coming from gesture detector

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
yunikkk: Dont want to mix logic here. 
Logic is added in the `StandardGestureListener`, while `SimpleStandardOnGestureListener` is just a class implementing everything and returning `false`. If/when we'll fix in in the gestures lib (which seams not feasible right now since will be non-backward compatible) we'll get rid of out Simple listener

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
yunikkk: All right

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
yunikkk: MotionEvent is not passed to us in other detectors. 
However, in fact this issue should be fixed in the upstream Gestures [library](https://github.com/mapbox/mapbox-gestures-android/blob/6fc34d91fc8d8758c51830127c49223624d446aa/library/src/main/java/com/mapbox/android/gestures/StandardGestureDetector.java#L98). 

Created an issue in the gesture repo https://github.com/mapbox/mapbox-gestures-android/issues/100

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
pengdev: > which seams not feasible right now since will be non-backward compatible)

I think if this is a bug, e.g the MotionEvent in Android GestureDetector is not marked as NonNull
```
    public interface OnGestureListener {

        /**
         * Notified when a tap occurs with the down {@link MotionEvent}
         * that triggered it. This will be triggered immediately for
         * every down event. All other events should be preceded by this.
         *
         * @param e The down motion event.
         */
        boolean onDown(MotionEvent e);
```
and it is marked as NonNull in the SimpleStandardOnGestureListener in the mapbox-gestures library
```
    @Override
    public boolean onSingleTapConfirmed(@NonNull MotionEvent e) {
      return false;
    }
```
We can probably just remove the NonNull annotation and consider it a bug fix, instead of a API breaking change. wdyt?

yunikkk: Well, still a breaking change, not sure we need to introduce it if its possible to fix on our side. wdyt? cc @mapbox/maps-android

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
```diff
@@ -1275,7 +1334,6 @@ class GesturesPluginImpl : GesturesPlugin, GesturesSettingsBase, MapStyleObserve
   }
 
   internal fun handleFlingEvent(
-    e1: MotionEvent,
     e2: MotionEvent,
```
Chaoba, line 1337: Rename e2?

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
Chaoba: I would prefer we fix the bug on our side and add the issue link of gesture lib in the comment.

plugin-gestures/src/test/java/com/mapbox/maps/plugin/gestures/GesturesPluginTest.kt
```diff
@@ -1121,6 +1124,19 @@ class GesturesPluginTest {
     assertEquals(cameraOptionsSlot.captured.zoom!!, startZoom + expectedZoomDelta * 3, EPS)
   }
 
+  @Test
+  fun verifyDoesNotCrashIfNullMotionEventPassed() {
+    gestureListener.captured.onSingleTapConfirmed(null)
+    gestureListener.captured.onDoubleTap(null)
+    gestureListener.captured.onDoubleTapEvent(null)
```
alexander-kulikovskii, line 1131: nit:

```suggestion
    gestureListener.captured.apply {
         onSingleTapConfirmed(null)
         onDoubleTap(null)
         onDoubleTapEvent(null)
         ...
```

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
pengdev: > still a breaking change

Maybe we can release a major version of the gestures library? but yeah, we have gestures library as an api dependency, so these APIs are also part of our public APIs.

Imo if something is clearly broken without API changes, it would be dangerous to leave it untouched, user still will be able to find this class and use it. Maybe @LukasPaczos has some suggestions/recommendations as well?

LukasPaczos: I agree with @pengdev that this should be fixed in the gestures library. The current behavior is incorrect and cannot be justified as a use-case that we'd be breaking.

The gestures library is still at v0.x (probably shouldn't at this point :upside_down_face:) so to me the best solution is a v0.8.0 with fixed annotations and a clear communication in both gestures lib and Maps SDK changelogs that the API compatibility can be impacted for someone that overrides the listeners.

LukasPaczos: Looks like the annotations themselves are a regression from https://github.com/mapbox/mapbox-gestures-android/pull/95, we shouldn't have inferred the nullity for non-annotated Java APIs from the Android SDK.

yunikkk: Okay, then lets move this fix to the gestures instead.

yunikkk:
Closing, will reopen with the new gestures lib instead

plugin-gestures/src/main/java/com/mapbox/maps/plugin/gestures/GesturesPluginImpl.kt
yunikkk: Moved fix there https://github.com/mapbox/mapbox-gestures-android/pull/101

PR #1616, opened 2022-08-19
Use on draw listener in view annotation manager for more granular control

kiryldz:
### Summary of changes

In view annotation manager we were relying on `ViewTreeObserver.OnGlobalLayoutListener` listener to control 2 state changes for the view annotation:
 - dimension change caused by `WRAP_CONTENT`
 - visibility changes based on user calling `View.setVisibility` in order to update core with new visible status

However it turned out that changing view visibility from `VISIBLE` to `INVISIBLE` does not actually trigger `OnGlobalLayoutListener` as the view still remains in the view tree. That actually was breaking view annotation logic when user was simply setting view annotation visibility to `INVISIBLE`, dragging the map and making it `VISIBLE` again.

The solution found here is to actually check view visibility inside `ViewTreeObserver.OnDrawListener` that is called on each frame for the view. In order to save a bit of CPU cycles only visibility check was migrated to `OnDrawListener` while checking for `WRAP_CONTENT` dimensions did remain in `OnGlobalLayoutListener`.

## Pull request checklist:
    
### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/infowindow/MarkerManager.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/infowindow/MarkerManager.kt
@@ -111,7 +111,7 @@ class MarkerManager(
         offsetX(0)
       }
     )
-    view.visibility = View.GONE
+    view.visibility = View.INVISIBLE
   }
 
   // if info window view is shown near screen edge - we adjust offsetX so that it fully appears on the screen
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone.
 # main
+## Bug fixes 🐞
+* Fix broken view annotation positioning when marking them `View.INVISIBLE` and then making `View.VISIBLE`. ([1616](https://github.com/mapbox/mapbox-maps-android/pull/1616))
+
+# 10.8.0-rc.1
 ## Features ✨ and improvements 🏁
 
 ## Bug fixes 🐞
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,11 @@
 
 Mapbox welcomes participation and contributions from everyone.
 # main
-## Bug fixes 🐞
-* Fix broken view annotation positioning when marking them `View.INVISIBLE` and then making `View.VISIBLE`. ([1616](https://github.com/mapbox/mapbox-maps-android/pull/1616))
-
-# 10.8.0-rc.1
 ## Features ✨ and improvements 🏁
 
 ## Bug fixes 🐞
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/infowindow/MarkerManager.kt
```diff
@@ -111,7 +111,7 @@ class MarkerManager(
         offsetX(0)
       }
     )
-    view.visibility = View.GONE
+    view.visibility = View.INVISIBLE
```
yunikkk, line 114: Actually a bit suprising this doesn't work since docs clearly state that `Interface definition for a callback to be invoked when the global layout state or the visibility of views within the view tree changes. ` [here](https://developer.android.com/reference/android/view/ViewTreeObserver.OnGlobalLayoutListener)

kiryldz: well, I guess the docs are not precisely correct then... that happens 😄

pengdev: Or this is actually a bug of Android that we can report to Google 😉

yunikkk: Would be nice to report, agree

yunikkk: Also - couldn't we instead if listening for `onDraw` keep using GONE? Seems we haven't been able to find broken cases with it?

kiryldz: @yunikkk what do you mean? We're not controlling that - users may hide their views (mark them `View.INVISIBLE`) and then make them visible again and expect everything works as expected. Now that use-case is completely broken and that's what we're trying to fix.
I'm reverting L114 to `INVISIBLE` instead of `GONE` because that's how that this example was designed back in time and to showcase that it's working again now as expected.

yunikkk (approved):
Ok, do not forget to file an issue in Google tracker 🦖

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/infowindow/MarkerManager.kt
kiryldz: created an issue https://issuetracker.google.com/issues/244067928

ank27 (approved):

app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/infowindow/MarkerManager.kt
pengdev: Shall we also add some inline docs to this example to add more context and suggest to use INVISIBLE instead of GONE?

PR #1620, opened 2022-08-23
Rs fix scale bar truncated - MAPSAND-305

jush:
### Summary of changes

Fixes issue where the scale bar is truncated at high zoom levels near the poles. The issues is easier to reproduce when the scale bar `ratio` property is set to small number like `0.3`.
See `Before` in following capture where the map is at long=0.0, lat=81.0 and zoom level 21.5.

After these changes, if the distance in the pre-defined scale bar tables (see `ScaleBarConstants` [metricTable](https://github.com/mapbox/mapbox-maps-android/blob/63535428955bf5d4a968ddbe7f3de145e3b8e795/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarConstant.kt#L12) and [imperialTable](https://github.com/mapbox/mapbox-maps-android/blob/63535428955bf5d4a968ddbe7f3de145e3b8e795/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarConstant.kt#L51)) does not fit in the view then the maximum distance that would fit will be used.

| Before | After |
| ----- | ----- |
| scale_bar_truncated | scale_bar_truncated_fix |

## Pull request checklist:
    

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -312,20 +301,53 @@ class ScaleBarImpl : ScaleBar, View {
     canvas.drawText(text, x, y, textPaint)
   }
 
+  /**
+   * @return A tuple with the scale bar segment distance and the amount of bar segments
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun findSuitableScaleBarSegments(maxDistance: Float): Pair<Float, Int> {
```
Chaoba, line 308: Is there any reason to return `Pair<Float, Int>` instead of `Pair<Int, Int>` here?

Chaoba (commented):
Does iOS have the same issue?

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: Yes, we can now return values smaller than `1` in the case where `maxDistance` it is already smaller than `1`. So the scale bar will render a scale bar segment with decimal distance (see second screenshot in the PR description).

pengdev:
> Does iOS have the same issue?
Good catch, we should probably check iOS and align the behaviour across platforms.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -312,20 +301,53 @@ class ScaleBarImpl : ScaleBar, View {
...
+    var pair: Pair<Int, Int> = scaleTable[0]
+    for (i in 1 until scaleTable.size) {
```
alexander-kulikovskii, line 310: how often we call this method? Could there be performance issues with this code? maybe cache somehow index from scaleTable? or refactor it and store range for getting value with O(1)

plugin-scalebar/src/test/java/com/mapbox/maps/plugin/scalebar/ScaleBarImplTest.kt
```diff
@@ -98,25 +99,29 @@ class ScaleBarImplTest {
...
   fun getDistanceText() {
-    assertEquals("0", scaleBarView.getDistanceText(0))
-    assertEquals("100 m", scaleBarView.getDistanceText(100))
-    assertEquals("999 m", scaleBarView.getDistanceText(999))
-    assertEquals("1 km", scaleBarView.getDistanceText(1000))
-    assertEquals("1 km", scaleBarView.getDistanceText(1001))
-    assertEquals("4.4 km", scaleBarView.getDistanceText(4444))
-    assertEquals("5.6 km", scaleBarView.getDistanceText(5555))
-    assertEquals("10 km", scaleBarView.getDistanceText(10000))
+    assertEquals("0", scaleBarView.getDistanceText(0F))
```
alexander-kulikovskii, line 102: nice to add parametrized tests. same for `verify below and above scale bar segments`

plugin-scalebar/src/test/java/com/mapbox/maps/plugin/scalebar/ScaleBarImplTest.kt
```diff
@@ -140,4 +145,44 @@ class ScaleBarImplTest {
...
+    `verify below and above scale bar segments`(metricTable)
+    // Then verify for imperial units
+    scaleBarSettings.isMetricUnits = false
+    scaleBarView.settings = scaleBarSettings
+    `verify below and above scale bar segments`(imperialTable)
+  }
+
+  private fun `verify below and above scale bar segments`(scaleTable: List<Pair<Int, Int>>) {
+    // Special case where max distance is smaller than the first entry segment in the table
+    val belowFirstUnitBarDistance = scaleTable[0].first.toFloat() / scaleTable[0].second - 0.1F
```
alexander-kulikovskii, line 163: nit:
just for clarity
```suggestion
    val belowFirstUnitBarDistance = (scaleTable[0].first.toFloat() / scaleTable[0].second) - 0.1F
```

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -312,20 +301,53 @@ class ScaleBarImpl : ScaleBar, View {
...
    *
    * @param distance original distance
    * @return Formatted distance text
    */
-  internal fun getDistanceText(distance: Int): String {
-    return if (distance == 0) {
+  internal fun getDistanceText(distance: Float): String {
+    return if (distance == 0F) {
       "0"
     } else if (METER_UNIT == unit) {
```
alexander-kulikovskii, line 346: nit:
```suggestion
    } else {
       val (unitMinDistance, unitName) = if (METE_UNIT == unit) { KILOMETER to KILOMETER_UNIT } else { FEET_PER_MILE to MILE_UNIT }
        return if (distance < unitMinDistance) {
           decimalFormat.format(distance) + unit
        } else {
           decimalFormat.format(distance * 1.0 / unitMinDistance) + unitName
       }
      } 
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 Mapbox welcomes participation and contributions from everyone.
 # main
+* Fix scale bar truncated at high zoom levels near the poles. ([1620](https://github.com/mapbox/mapbox-maps-android/pull/1620))
```
kiryldz, line 5: Should go under `## Bug fixes 🐞` sub-block

jush:
> Does iOS have the same issue?

I quickly checked and actually the way the scale bar is rendered is different than in Android.
In Android we enforce a maximum width (given by the `ratio`). In iOS they don't have such rule, the scale bar can grow indefinitely. That results in a visual bug (MAPSIOS-301) where the scale bar is wider than the map view.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: This method is called on each draw. Unless `ScaleBar.useContinuousRendering` is set to true the view is invalidated per need basis.

AFAIK there hasn't been issues with this piece of code (see previous implementation at [ScaleBarImpl#onDraw](https://github.com/mapbox/mapbox-maps-android/blob/63535428955bf5d4a968ddbe7f3de145e3b8e795/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt#L235) before.
I think that before we do any optimization we should measure it. I would leave it out of this PR.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: Somehow I see your suggestion as harder/more difficult to read. (nit: moreover, it will create a new object (`Pair`) every time is called)

Maybe others can also give their opinion.

yunikkk: I'm fine with the current notation

CHANGELOG.md
jush: Fixed in 94d16073

Chaoba (approved):

plugin-scalebar/src/test/java/com/mapbox/maps/plugin/scalebar/ScaleBarImplTest.kt
jush: Please see 69fb987 and baca659 and let me know if that's what you were suggesting.

yunikkk (approved):

scripts/java-api-check.json
```diff
@@ -52,7 +52,8 @@
             "com\\.mapbox\\.maps\\.plugin\\.locationcomponent\\.ModelLayerRenderer",
             "com\\.mapbox\\.maps\\.plugin\\.locationcomponent\\.ModelLayerWrapper",
             "com\\.mapbox\\.maps\\.plugin\\.locationcomponent\\.ModelSourceWrapper",
-            "com\\.mapbox\\.maps\\.plugin\\.scalebar\\.RefreshHandler",
+            "com\\.mapbox\\.maps\\.plugin\\.scalebar\\.ScaleBarImpl",
+            "com\\.mapbox\\.maps\\.plugin\\.scalebar\\.ScaleBarImpl$RefreshHandler",
```
jush, line 56: @Chaoba My understanding is that previously the `RefreshHandler` was not picked up properly since it was a `ScaleBarImpl` inner class.

kiryldz (approved):
Very cool, LGTM!

alexander-kulikovskii (approved):

PR #1622, opened 2022-08-24
MAPSAND-379 Style getter extension functions do not save property values

paulrenn67:
### Summary of changes

This branch adds support to preserve cached layer properties during the getLayer call.

We then use any cached properties when the layer is later added to the style.

See related PR in mapbox-maps-android-internal which update the tests in this PR.

### User impact (optional)

Should not be any.

## Pull request checklist:


kiryldz:
@paulrenn67 I added `bug` label - `verify-code` will not pass then until you add the changelog entry. However in your case it's failing because of `ktlint`. Could you please fix that so that CI will start?

kiryldz:
Unit tests are failing with `ClassCastException`s...

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -100,6 +100,11 @@ abstract class Layer : StyleContract.StyleLayerExtension {
    */
   fun bindTo(delegate: StyleInterface) {
     bindTo(delegate, null)
+    (this.getCachedLayerProperties().contents as? Value?)?.let { value ->
+      (value.contents as HashMap<String, Value>).forEach {
+        delegate.addStyleLayer(it.value, null)
+      }
+    }
```
pengdev, line 107: shouldn't these logic be added to the `fun bindTo(delegate: StyleInterface, position: LayerPosition?)` instead of here?

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -119,6 +124,12 @@ abstract class Layer : StyleContract.StyleLayerExtension {
 
   // Layer Properties
 
+  internal fun setCachedLayerProperties(properties: HashMap<String, Value>) {
+    properties.forEach {
+      layerProperties[it.key] = PropertyValue(it.key, it.value)
```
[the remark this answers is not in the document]
paulrenn67: Yes that might be better, the API is a bit problematic and unit tests cause problems if they have no values at all

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 Mapbox welcomes participation and contributions from everyone.
 # main
+
+## Bug fixes 🐞
+* Preserve cached layer properties during the getLayer call. ([1622](https://github.com/mapbox/mapbox-maps-android/pull/1622))
```
kiryldz, line 7: based on the delta here this PR was not rebased on latest main for quite a while, could you please do that?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 * Fix scale bar truncated at high zoom levels near the poles. ([1620](https://github.com/mapbox/mapbox-maps-android/pull/1620))
 * Fix broken view annotation positioning when marking them `View.INVISIBLE` and then making `View.VISIBLE`. ([1616](https://github.com/mapbox/mapbox-maps-android/pull/1616))
+* Preserve cached layer properties during the getLayer call. ([1622](https://github.com/mapbox/mapbox-maps-android/pull/1622))
```
kiryldz, line 11:
```suggestion
* Preserve cached properties if applied to the layer before during the `Style#getLayer` call. ([1622](https://github.com/mapbox/mapbox-maps-android/pull/1622))
```

extension-style-app/src/androidTest/java/com/mapbox/maps/testapp/style/layers/generated/CircleLayerTest.kt
```diff
@@ -706,6 +706,72 @@ class CircleLayerTest : BaseStyleTest() {
...
+    val circleStrokeColorTestValue = "rgba(0, 0, 0, 1)"
+    val circleStrokeOpacityTestValue = 1.0
+    val circleStrokeWidthTestValue = 1.0
+    val circleTranslateTestValue = listOf(0.0, 1.0)
+    val circleTranslateAnchorTestValue = CircleTranslateAnchor.MAP
+
+    val layer = circleLayer("id", "source") {
+      sourceLayer("test")
+      minZoom(10.0)
+      maxZoom(10.0)
```
kiryldz, line 735: why is this `10.0` not stored in constants as all the other properties?

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -18,23 +19,30 @@ import com.mapbox.maps.logE
...
+    else -> {
+      logE(TAG, "Layer type: $type unknown.")
+      null
+    }
+  }?.also { result ->
+    result.delegate = this
+    try {
+      result.cachedLayerPropertiesValue = getStyleLayerProperties(layerId).value
+    } catch (e: ClassCastException) {
+      // Exception can be thrown during unit tests (impossible to mock getStyleLayerProperties)
```
kiryldz, line 45: why couldn't that be mocked? can you provide a bit more context?

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -18,23 +19,30 @@ import com.mapbox.maps.logE
...
+    "symbol" -> SymbolLayer(layerId, source())
+    else -> {
+      logE(TAG, "Layer type: $type unknown.")
+      null
+    }
+  }?.also { result ->
+    result.delegate = this
+    try {
+      result.cachedLayerPropertiesValue = getStyleLayerProperties(layerId).value
+    } catch (e: ClassCastException) {
```
jush, line 44: Probably not worth it but, we could avoid calling three times the native side by calling `getStyleLayerProperties(...)` in the beginning and then get the required properties directly.

Something like this:
```suggestion
  val layerProperties = getStyleLayerProperties(layerId).value
  val styleLayerProperties: Map<String, Any> = layerProperties!!.unwrapToTyped()
  val source: String? = styleLayerProperties["type"] as String?
  val type = styleLayerProperties["type"]!! as String
  return when (type) {
    "background" -> BackgroundLayer(layerId)
    "location-indicator" -> LocationIndicatorLayer(layerId)
    "sky" -> SkyLayer(layerId)
    "circle" -> CircleLayer(layerId, source!!)
    "fill-extrusion" -> FillExtrusionLayer(layerId, source!!)
    "fill" -> FillLayer(layerId, source!!)
    "heatmap" -> HeatmapLayer(layerId, source!!)
    "hillshade" -> HillshadeLayer(layerId, source!!)
    "line" -> LineLayer(layerId, source!!)
    "raster" -> RasterLayer(layerId, source!!)
    "symbol" -> SymbolLayer(layerId, source!!)
    else -> {
      logE(TAG, "Layer type: $type unknown.")
      null
    }
  }?.also { result ->
    result.delegate = this
    try {
      result.cachedLayerPropertiesValue = layerProperties
    } catch (e: ClassCastException) {
```

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
paulrenn67: `getStyleLayerProperties` returns `Expected<String, Value>`, but was unable to mock getStyleLayerProperties to return a valid Expected object since Expected constructor is not exposed.  Maybe I misunderstood something, but at least trying the following does not compile: 
`every { style.getStyleLayerProperties("id") } returns Expected(null, null)
`

But now I noticed we have this in some tests:
`private val valueExpected = mockk<Expected<String, Value>>(relaxUnitFun = true, relaxed = true)
`

and then it would be possible to do:
`every { style.getStyleLayerProperties("id") } returns valueExpected
`

So I could remove this try catch and add this each of the Layer tests.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -28,6 +28,8 @@ abstract class Layer : StyleContract.StyleLayerExtension {
 
   internal var delegate: StyleManagerInterface? = null
 
+  internal var cachedLayerPropertiesValue: Value? = null
```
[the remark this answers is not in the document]
paulrenn67: We can rename it

extension-style-app/src/androidTest/java/com/mapbox/maps/testapp/style/layers/generated/CircleLayerTest.kt
paulrenn67: This was copied from other tests, will make these consistent.

CHANGELOG.md
paulrenn67: ok

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
paulrenn67: This is now fixed and the try-catch block has been removed

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
paulrenn67: It's a fair point but think we don't need to go that far at the moment, also tests would need further refactoring since there are often not real values provided and !! causes the test to fail.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -1,4 +1,5 @@
 @file:JvmName("LayerUtils")
+
```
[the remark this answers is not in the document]
paulrenn67: Actually yes, if I delete that line in the IDE and then run the auto-formatter, it forces a newline there.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -111,7 +113,9 @@ abstract class Layer : StyleContract.StyleLayerExtension {
   override fun bindTo(delegate: StyleInterface, position: LayerPosition?) {
     this.delegate = delegate
 
-    val expected = delegate.addStyleLayer(getCachedLayerProperties(), position)
+    val expected = cachedLayerPropertiesValue?.let {
+      delegate.addStyleLayer(it, position)
+    } ?: delegate.addStyleLayer(getCachedLayerProperties(), position)
```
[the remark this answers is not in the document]
paulrenn67: That's looks cleaner 👍

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
```diff
@@ -18,24 +19,26 @@ import com.mapbox.maps.logE
...
-      "heatmap" -> HeatmapLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
-      "hillshade" -> HillshadeLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
-      "line" -> LineLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
-      "raster" -> RasterLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
-      "symbol" -> SymbolLayer(layerId, this.getStyleLayerProperty(layerId, "source").unwrap()).also { it.delegate = this }
-      else -> {
-        logE(TAG, "Layer type: $type unknown.")
-        null
-      }
+  val source by lazy { getStyleLayerProperty(layerId, "source").unwrap<String>() }
```
pengdev, line 22: There're layers which do not have a "source" property, and that's why we specifically call this function inside the layer constructors.
i.e. background/location-indicator/sky layer doesn't have/support "source" property, and will report error here if you unwrap it to a String.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -111,7 +113,8 @@ abstract class Layer : StyleContract.StyleLayerExtension {
   override fun bindTo(delegate: StyleInterface, position: LayerPosition?) {
     this.delegate = delegate
 
-    val expected = delegate.addStyleLayer(getCachedLayerProperties(), position)
+    val propertiesValue = appliedLayerPropertiesValue ?: getCachedLayerProperties()
+    val expected = delegate.addStyleLayer(propertiesValue, position)
```
pengdev, line 117: once set the `appliedLayerPropertiesValue`, we should also be able to set the runtime-cached property values in the `getCachedLayerProperties`, i.e.
```
val properties = layerProperties.values.filter { it.propertyName != "id" &&  it.propertyName != "type" && it.propertyName != "source"}
properties.forEach { 
  delegate?.setStyleLayerProperty(layerId, it.propertyName, it.value)
}
```

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/LayerExt.kt
paulrenn67: I used lazy evaluation for `source` so it should not be evaluated until the variable is actually used, so I think it should be safe to do this since source is not used in the BackgroundLayer, LocationIndicatorLayer, SkyLayer constructors.

pengdev: ah, yes, sorry I missed lazy evaluation, sounds good!

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
paulrenn67: done

kiryldz (approved):

pengdev (approved):
LGTM with one comment 👍 

As discussed offline, please also cut a ticket regarding the potential tail work to refactor and merge `appliedLayerPropertiesValue` and `layerProperties`.

extension-style/src/main/java/com/mapbox/maps/extension/style/layers/Layer.kt
```diff
@@ -111,14 +113,21 @@ abstract class Layer : StyleContract.StyleLayerExtension {
...
     expected.error?.let {
       throw MapboxStyleException("Add layer failed: $it")
     }
+
+    if (appliedLayerPropertiesValue != null) {
+      layerProperties.values.filter { it.propertyName != "id" && it.propertyName != "type" && it.propertyName != "source" }
+        .forEach {
+          delegate.setStyleLayerProperty(layerId, it.propertyName, it.value)
+        }
+    }
```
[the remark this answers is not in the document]
paulrenn67: > LGTM with one comment 👍
> 
> As discussed offline, please also cut a ticket regarding the potential tail work to refactor and merge `appliedLayerPropertiesValue` and `layerProperties`.

Created https://mapbox.atlassian.net/browse/MAPSAND-418

PR #1623, opened 2022-08-24
Bump gl-native to 10.8-rc.1

pengdev:
## Bug fixes 🐞
* Relax LOD requirements for maps with insets and zero terrain exaggeration, leading to more content shown in easily readable map areas
* Make CameraManager.setCamera method exception free. In cases when incorrect CameraOptions are provided, error would be logged.
* Fix black holes in the globe view when edge insets are used 

## Dependencies
* Upgrade mapbox-common version to v23.0.0-rc.2

## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. If tests were not written, please explain why.
 - [ ] Optimize code for java consumption (`@JvmOverloads`, `@file:JvmName`, etc).
 - [ ] Add example if relevant.
 - [ ] Document any changes to public APIs.
 - [ ] Run `make update-api` to update generated api files, if there's public API changes, otherwise the `verify-api-*` CI steps might fail.
 - [x] Update [CHANGELOG.md](../CHANGELOG.md) or use the label 'skip changelog', otherwise `check changelog` CI step will fail.
 - [x] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.[version]` release branch.

Fixes: < Link to related issues that will be fixed by this pull request, if they exist >

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-android/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1624, opened 2022-08-24
[v10.9] revert setCamera exception fix #652

ank27:
### Summary of changes
- After gl-native bump to v10.8.0-rc.1, exceptions in set camera will be handled in native, which makes the fix in https://github.com/mapbox/mapbox-maps-android/pull/652 redundant. Reverting old fix in this pr. 

More context can be found in MAPSAND-389.

## Pull request checklist:
    
 - [x] If this PR is a `v10.9` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.9` release branch.

### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
@@ -133,7 +133,8 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     }
   }
 
-  private var lastCameraOptions: CameraOptions? = null
+  @VisibleForTesting(otherwise = PRIVATE)
+  internal var lastCameraOptions: CameraOptions? = null
   private var cameraOptionsBuilder = CameraOptions.Builder()
 
   private lateinit var mapDelegateProvider: MapDelegateProvider
@@ -180,15 +181,23 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
     handler.removeCallbacks(commitChangesRunnable)
   }
 
-  private fun performMapJump(cameraOptions: CameraOptions) {
+  @VisibleForTesting(otherwise = PRIVATE)
+  internal fun performMapJump(cameraOptions: CameraOptions) {
     if (lastCameraOptions == cameraOptions) {
       return
     }
     // move native map to new position
-    mapCameraManagerDelegate.setCamera(cameraOptions)
-    // notify listeners with actual values
-    notifyListeners(mapCameraManagerDelegate.cameraState)
-    lastCameraOptions = cameraOptions
+    try {
+      mapCameraManagerDelegate.setCamera(cameraOptions)
+      // notify listeners with actual values
+      notifyListeners(mapCameraManagerDelegate.cameraState)
+      lastCameraOptions = cameraOptions
+    } catch (e: Exception) {
+      logE(
+        TAG,
+        "Exception while setting camera options : ${e.message} CameraOptions = $cameraOptions"
+      )
+    }
   }
 
   private fun updateAnimatorValues(cameraAnimator: CameraAnimator<*>): Boolean {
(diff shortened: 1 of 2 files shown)


ank27:
will add changelog once https://github.com/mapbox/mapbox-maps-android/pull/1623 is merged.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -177,28 +176,15 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
-      mapCameraManagerDelegate.setCamera(cameraOptions)
-      // notify listeners with actual values
-      notifyListeners(mapCameraManagerDelegate.cameraState)
-      lastCameraOptions = cameraOptions
-    } catch (e: Exception) {
-      logE(
-        TAG,
-        "Exception while setting camera options : ${e.message} CameraOptions = $cameraOptions"
-      )
-    } catch (error: Error) {
```
pengdev, line 196: we were catching both `Exception` and `Error`, would it behave differently since gl-native only catches `Error` I guess?

ank27: that's correct, we are also catching `Exceptions`. while native only cath `Error`. 
should we just keep exceptions here and let native handle `Error`? cc @mapbox/maps-android

kiryldz: imo we should not catch anything - if user does pass lat as NaN - it's pretty logical to actually fail with exception thrown from gl-native.
even if we decide to catch it - then it's logical to move catching it in `MapboxMap` and not only in `CameraAnimationsPluginImpl.kt`.
I would lean to removing catching exceptions from our end - e.g. in gestures code we already check points for NaN values and that's the correct path

ank27:
As discussed offline, this pr will be merged with v10.9-beta.1

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 * Fix scale bar truncated at high zoom levels near the poles. ([1620](https://github.com/mapbox/mapbox-maps-android/pull/1620))
 
+# 10.9.0-beta.1
+## Features ✨ and improvements 🏁
+* Remove redundant exception handling for `CameraManager.setCamera` method. Exception in `CameraManager.setCamera` now handled in native.([1624](https://github.com/mapbox/mapbox-maps-android/pull/1624))
```
kiryldz, line 12: We did remove redundant exception handling in _camera animation plugin_ and not in camera manager. Moreover if someone was not using for some reason our animation system and was calling `MapboxMap#setCamera` directly - now they could be sure that this method is exception free.

pengdev (approved):

ank27:
we are keeping catching `Exceptions` as it is as native catches `java.lang.error` and not `java.lang.exception`.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -197,12 +197,10 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
       )
-    } catch (error: Error) {
-      logE(
-        TAG,
-        "Error while setting camera options : ${error.message} CameraOptions = $cameraOptions"
-      )
     }
+    // notify listeners with actual values
+    notifyListeners(mapCameraManagerDelegate.cameraState)
+    lastCameraOptions = cameraOptions
```
kiryldz, line 203: either I'm crazy or I do see the same code L192-194

ank27: nice catch. forgot to remove those lines 🙈

kiryldz (approved):

yunikkk (approved):

PR #1635, opened 2022-08-31
MAPSAND-221 setting puck bearing enabled to false does not reset bearing

paulrenn67:
### Summary of changes

Disabling the puck in LocationPuckManager will now cause the puck direction to snap back to north.  This was requested by a customer.

### User impact (optional)

Calling LocationPuckManager.applySettings2 with LocationComponentSettings2.puckBearingEnabled == false will cause the puck to instantly animate to the north position.  When calling with puckBearingEnabled == true, the puck will animate smoothly back to the last reported heading.

## Pull request checklist:
        | <video src="https://user-images.githubusercontent.com/16490463/187616284-da6a58c8-9cc9-459c-9530-6c4b3e82487a.mp4" width = 250/>

Fixes: 

https://mapbox.atlassian.net/jira/software/c/projects/MAPSAND/issues/MAPSAND-221

### DIFF
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
@@ -180,13 +180,17 @@ internal class LocationPuckManager(
       animationManager.setEnabled(true)
       animateToBearing(bearings, options, forceUpdate)
     } else {
-      animateToBearing(doubleArrayOf(0.0), options = {
-        options?.invoke(this)
-        duration = 0
-        doOnEnd {
-          animationManager.setEnabled(false)
-        }
-      }, forceUpdate)
+      animateToBearing(
+        doubleArrayOf(0.0),
+        options = {
+          options?.invoke(this)
+          duration = 0
+          doOnEnd {
+            animationManager.setEnabled(false)
+          }
+        },
+        forceUpdate
+      )
     }
   }
 
--- a/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
+++ b/plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
@@ -176,38 +176,18 @@ internal class LocationPuckManager(
     options: (ValueAnimator.() -> Unit)? = null,
     forceUpdate: Boolean = false
   ) {
-    when (settings2.puckBearingEnabled) {
-      false -> {
-        // Puck bearing is disabled, snap back to north.
-        snapToNorth(options, false)
-      }
-      true -> {
-        // Puck bearing is enabled, animate to new bearing.
-        animateToBearing(bearings, options, forceUpdate)
-      }
-    }
-  }
-
-  @VisibleForTesting(otherwise = PRIVATE)
-  fun snapToNorth(
-    options: (ValueAnimator.() -> Unit)? = null,
-    forceUpdate: Boolean
-  ) {
-    // Skip bearing updates if the change from the lastBearing is too small, thus avoid unnecessary calls to gl-native.
-    if (!forceUpdate && lastBearing < BEARING_UPDATE_THRESHOLD) {
-      return
-    }
-    val targets = doubleArrayOf(lastBearing, 0.0)
-    animationManager.animateBearing(
-      *targets,
-      options = {
+    if (settings2.puckBearingEnabled) {
+      animationManager.setEnabled(true)
+      animateToBearing(bearings, options, forceUpdate)
+    } else {
+      animateToBearing(doubleArrayOf(0.0), options = {
         options?.invoke(this)
         duration = 0
         doOnEnd {
-          animationManager.setDisabled()
+          animationManager.setEnabled(false)
         }
-      }
-    )
+      }, forceUpdate)
+    }
   }
 
   @VisibleForTesting(otherwise = PRIVATE)
@@ -221,7 +201,6 @@ internal class LocationPuckManager(
       return
     }
     val targets = doubleArrayOf(lastBearing, *bearings)
-    animationManager.setEnabled()
     animationManager.animateBearing(
       *targets,
       options = options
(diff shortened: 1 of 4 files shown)


plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
```diff
@@ -173,12 +175,53 @@ internal class LocationPuckManager(
...
+      }
+      true -> {
+        // Puck bearing is enabled, animate to new bearing.
+        animateToBearing(bearings, options, forceUpdate)
+      }
+    }
+  }
+
+  @VisibleForTesting(otherwise = PRIVATE)
+  fun snapToNorth(
```
kiryldz, line 192: just as a thought - why can't we actually re-use `animateToBearing` function with options.duration = 0 and target = 0?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimatorManager.kt
```diff
@@ -117,9 +117,14 @@ internal class PuckAnimatorManager(
...
-      enabled = settings2.puckBearingEnabled
-    }
+  }
+
+  internal fun setDisabled() {
+    bearingAnimator.enabled = false
+  }
+
+  internal fun setEnabled() {
+    bearingAnimator.enabled = true
```
pengdev, line 127: nit: maybe unify these two methods to `setEnabled(enabled: Boolean)`?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
paulrenn67: hmmm I guess we could do so

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimatorManager.kt
paulrenn67: I was going to do that but usually we have separate setters for each state, but since this is purely internal, maybe it is ok to unify

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
```diff
@@ -173,9 +175,33 @@ internal class LocationPuckManager(
...
     forceUpdate: Boolean = false
+  ) {
+    if (settings2.puckBearingEnabled) {
+      animationManager.setEnabled(true)
+      animateToBearing(bearings, options, forceUpdate)
+    } else {
+      animateToBearing(
+        doubleArrayOf(0.0),
+        options = {
+          options?.invoke(this)
```
kiryldz, line 186: actually not sure we need to use options provided from above, it should be sufficient to use zero duration + zero target bearing

paulrenn67: Yeah I guess we should discard any user-provided options at this point since we want a very specific behaviour. Good catch.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimatorManager.kt
```diff
@@ -117,9 +117,10 @@ internal class PuckAnimatorManager(
       accuracyCircleColor = settings2.accuracyRingColor
       accuracyCircleBorderColor = settings2.accuracyRingBorderColor
     }
-    bearingAnimator.apply {
-      enabled = settings2.puckBearingEnabled
-    }
+  }
+
+  internal fun setEnabled(enabled: Boolean) {
```
kiryldz, line 122: `setEnabled` naming for the method that controls only bearing animator sounds a bit strange. Perhaps rename?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
```diff
@@ -173,9 +175,32 @@ internal class LocationPuckManager(
     vararg bearings: Double,
     options: (ValueAnimator.() -> Unit)? = null,
     forceUpdate: Boolean = false
+  ) {
+    if (settings2.puckBearingEnabled) {
+      animationManager.setEnabled(true)
```
kiryldz, line 180: just mentioned one thing that concerns me here: `updateCurrentBearing` is technically getting called very often not depending if `puckBearingEnabled` is true or false. That actually means that e.g. when `puckBearingEnabled = false` - we will still be calling 0-duration animation each time and also `animationManager.setEnabled(false)`. Similarly  `animationManager.setEnabled(true)` is called each time.
Can we actually incapsulate that logic and perform it only when `puckBearingEnabled` did actually change?
The only thing we truly need to do on each bearing update is `animateToBearing(bearings, options, forceUpdate)` when `settings2.puckBearingEnabled = true`

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/animators/PuckAnimatorManager.kt
paulrenn67: Do you prefer `setPuckAnimationEnabled` (in next commit)?

kiryldz (approved):

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/LocationPuckManager.kt
paulrenn67: Resolved by filtering out calls to animateToBearing when puckBearingEnabled is false.

alexander-kulikovskii (approved):

PR #1664, opened 2022-09-12
MAPSAND-397 Attempt to detect invalid Activity when calling MapView.getMapboxMap

paulrenn67:
Covers the situation where a MapView can be in a Fragment which is removed from the Activity before onAttachedToWindow is called. Not sure whether we should really be trying to handle this case in the SDK, looks more like an issue with the user's application logic.

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapView.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapView.kt
@@ -1,7 +1,9 @@
 package com.mapbox.maps
 
 import android.annotation.SuppressLint
+import android.app.Activity
 import android.content.Context
+import android.content.ContextWrapper
 import android.graphics.Bitmap
 import android.opengl.GLES20
 import android.util.AttributeSet
@@ -115,6 +117,7 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
     mapController.onAttachedToWindow(this)
   }
 
+
   @SuppressLint("CustomViewStyleable")
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal fun parseTypedArray(context: Context, attrs: AttributeSet?): MapInitOptions {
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -227,7 +230,21 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
    * @return [MapboxMap] object to interact with the map.
    */
-  override fun getMapboxMap(): MapboxMap = mapController.getMapboxMap()
+  override fun getMapboxMap(): MapboxMap {
+    val activity = context.getActivity()
+    if (activity == null || activity.isFinishing || activity.isDestroyed) {
+      if (viewAnnotationManagerDelegate.isInitialized()) {
+        (viewAnnotationManager as ViewAnnotationManagerImpl).destroy()
+      }
+      mapController.onDestroy()
```
kiryldz, line 239: just call `onDestroy()` here from `MapView`? no need to duplicate the code

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -410,4 +428,7 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
       }
     }
   }
-}
\ No newline at end of file
+}
+
+private tailrec fun Context.getActivity(): Activity? = this as? Activity
```
kiryldz, line 433: never even heard of `tailrec` 😄

sdk/src/test/java/com/mapbox/maps/MapViewTest.kt
```diff
@@ -67,6 +67,13 @@ class MapViewTest {
   @Test
   fun getMapboxMap() {
     mapView.getMapboxMap()
+    verify { mapController.onDestroy() }
```
kiryldz, line 70: kinda surprised why is it even called here...
guess due to some mocking but perhaps we can add more details why is it called here

kiryldz (commented):
I wonder if this could be covered by instrument test-case, from first glance seems that it should be doable

kiryldz (commented):
Also please not forget to the `bug` label + changelog if we decide to proceed

sdk/src/test/java/com/mapbox/maps/MapViewTest.kt
paulrenn67: In this case, it is indeed because the objects are mocked and the Activity is null.  Not a great test, but it does behave as expected in this case. Maybe should add some kind of comments.

sdk/src/main/java/com/mapbox/maps/MapView.kt
paulrenn67: Didn't want to call `onDestroy` just in case anything in the superclass is affected, but agree it would be better to extract this functionality to its own method.

PR #1672, opened 2022-09-15
Integrate core telemetry - MAPSAND-244

pengdev:
### Summary of changes

This PR migrated telemetry implementation from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK.

## Pull request checklist:
    

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapBaseEvent.kt
```diff
@@ -1,16 +1,13 @@
...
 import com.google.gson.annotations.SerializedName
-import com.mapbox.android.telemetry.Event
 
 /**
  * Base event class for telemetry events.
  */
 @SuppressLint("ParcelCreator")
-internal abstract class MapBaseEvent(phoneState: PhoneState) :
-  Event() {
+internal abstract class MapBaseEvent(phoneState: PhoneState) {
```
pengdev, line 10: Need to double check the Event class in MME and see if we miss anything.

pengdev: refs https://github.com/mapbox/mapbox-events-android/blob/d73ca293b194376516997b73c5139b28430be671/libtelemetry/src/main/java/com/mapbox/android/telemetry/Event.java

plugin-locationcomponent/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ android {
 }
 
 dependencies {
-  api(Dependencies.mapboxAndroidCore)
```
pengdev, line 26: Need to remember to add this dependency back to avoid api breaking change, internally we are migrated to the common sdk compat package.

app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
pengdev, line 8: Do we really need this new permission? should it be added from common SDK? cc @tarigo

tarigo: Required by `TelemetryUtils` to  obtain cellular network type.

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
```diff
@@ -191,8 +190,8 @@ class DefaultLocationProvider @VisibleForTesting(otherwise = PRIVATE) internal c
     private val locationProviderWeakReference: WeakReference<DefaultLocationProvider> =
       WeakReference(locationProvider)
 
-    override fun onSuccess(result: LocationEngineResult) {
-      result.lastLocation?.let { location ->
+    override fun onSuccess(result: LocationEngineResult?) {
+      result!!.lastLocation?.let { location ->
```
yunikkk, line 194: Can null arrive here?

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/PermissionsListener.java
```diff
@@ -0,0 +1,14 @@
+package com.mapbox.maps.plugin.locationcomponent;
+
+import java.util.List;
+
+/**
+ * Callback used in PermissionsManager
+ */
+
```
yunikkk, line 8:
```suggestion
```

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/PermissionsManager.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.List;
+
+/**
+ * Helps request permissions at runtime.
+ */
+public class PermissionsManager {
+
+    private static final String LOG_TAG = "PermissionsManager";
+    private static final String COARSE_LOCATION_PERMISSION = Manifest.permission.ACCESS_COARSE_LOCATION;
+    private static final String FINE_LOCATION_PERMISSION = Manifest.permission.ACCESS_FINE_LOCATION;
```
yunikkk, line 25: Since they're already constants - is there any benefit to create our own? Maybe just inline `Manifest.permission.ACCESS_FINE_LOCATION` ?

yunikkk: Ah, this file seems to be copied from other lib, nevermind then

module-telemetry/src/test/java/com/mapbox/maps/module/telemetry/MapTelemetryTest.kt
```diff
@@ -32,49 +76,53 @@ class MapTelemetryTest {
...
-    verify { mapboxTelemetry.updateSessionIdRotationInterval(any()) }
+    assertFalse(telemetry.setSessionIdRotationInterval(22))
   }
 
   @Test
+  @Ignore("testSetDebugLoggingEnabled has been deprecated")
   fun testSetDebugLoggingEnabled() {
-    telemetry.setDebugLoggingEnabled(true)
-    verify { mapboxTelemetry.updateDebugLoggingEnabled(true) }
+    // no-ops
```
yunikkk, line 114: Verify it's no-op?

app/src/main/AndroidManifest.xml
yunikkk: Hmm, but was it used in the same manner before in TelemetryUtils? Why do we add the permission now? As far as I can see if no permission is present the exception will be just swallowed https://github.com/mapbox/mapbox-sdk-common/blob/137e07ef970156322ecad99d3ea642a60ee0c428/components/telemetry_service/platform/android/src/main/java/com/mapbox/common/TelemetrySystemUtils.java#L154

plugin-locationcomponent/src/main/java/com/mapbox/maps/plugin/locationcomponent/DefaultLocationProvider.kt
pengdev: This is outdated, common SDK introduced this breaking change and has reverted it back.

app/src/main/AndroidManifest.xml
pengdev: Discussed offline with @tarigo , I think we can just remove it, this is not required field, and we only gather it if user has already granted it for other purposes.

sdk-base/src/main/java/com/mapbox/common/MapboxMapsAndroidLogger.kt
```diff
@@ -1,6 +1,6 @@
 package com.mapbox.common
 
-internal object MapboxCommonLogger {
+internal object MapboxMapsAndroidLogger {
```
pengdev, line 3: For more context, renaming this internal class because common SDK introduced the same class in the same package, renaming wouldn't be an issue since this is an internal change.

tatiana-yan: And are we going to have this change in maps-android separately from this PR? Cause we released CommonSDK 23.1.0-beta.1 and 23.1.0-beta.2 coming soon... There may be issues with adoption

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapBaseEvent.kt
tatiana-yan: Event type is determined from `event`  key in json, for MapLoad value should be `map.load`. Event should respect the [schema](https://github.com/mapbox/event-schema/blob/master/lib/base-schemas/map.load.js). Which means that valid event should have `type = map.load`, and also valid `userId`  and `created`  fields and may have some of optional fields. We do not serialize class type, I see no issues here.

valid `map.load` event example:
```
[{"wifi":false,"cellularNetworkType":"Unknown","orientation":"Portrait","userId":"d1cec2d6-5966-4608-9e85-59ac51178de0","sdkVersion":"10.8.0-SNAPSHOT","created":"2022-09-13T14:47:47.944+0300","sdkIdentifier":"mapbox-maps-android","event":"map.load","operatingSystem":"Android - 13","resolution":2.75,"carrier":"T-Mobile","model":"sdk_gphone64_arm64","pluggedIn":false,"accessibilityFontScale":1.0,"batteryLevel":100}]
```

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -17,7 +22,8 @@ class MapTelemetryImpl : MapTelemetry {
 
   private val appContext: Context
   private val accessToken: String
-  private val telemetry: MapboxTelemetry
+  private val eventsService: EventsServiceInterface
+  private val telemetryService: TelemetryService
```
tatiana-yan, line 26: nit: consider adding a comment, something like `// We need telemetryService for location and metrics event which will be sent automatically if telemetryService is initialized.`

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -28,39 +34,78 @@ class MapTelemetryImpl : MapTelemetry {
...
+
+    // TODO deferredDeliveryServiceOptions?
+    val eventsServiceOptions = EventsServerOptions(eventsToken, BuildConfig.MAPBOX_EVENTS_USER_AGENT, null)
+    this.eventsService = EventsService.getOrCreate(eventsServiceOptions)
+
+    this.telemetryService = TelemetryService.getOrCreate(eventsServiceOptions)
+
+    val coreTelemetryState = TelemetryUtils.getEventsCollectionState()
+    if (coreTelemetryState) {
+      enableTelemetryCollection(true)
```
tatiana-yan, line 51: we had more tricky code with pause/resume collection and simultaneous management of MME & Core Telemetry here, but now we do not need 

```
if (coreTelemetryState) {
      enableTelemetryCollection(true)
```
cause it is 
```
if (TelemetryUtils.getEventsCollectionState() == true)
  TelemetryUtils.setEventsCollectionState(true)
```

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -110,11 +152,18 @@ class MapTelemetryImpl : MapTelemetry {
...
     )
+    sendEvent(Gson().toJson(performanceEvent))
+  }
+
+  /**
+   * Static variables and methods.
+   */
+  private companion object {
+    private const val TAG = "MapTelemetryImpl"
+    private const val EVENTS_ACCESS_TOKEN_RESOURCE_NAME = "mapbox_events_access_token"
```
tatiana-yan, line 167: we used it for mme vs core telemetry testing to have separate tokens for mme and core telemetry (prod for mme, staging for core telemetry)
but in case of complete switch to core telemetry I think we do not need it, we need to use the same token we used for mme

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapBaseEvent.kt
pengdev: checked the event body we sent and I see no issue either, thanks for the confirmation!

sdk-base/src/main/java/com/mapbox/common/MapboxMapsAndroidLogger.kt
pengdev: we will rename this class when we do gl-native/common dependency bump, so shouldn't be an issue.

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -28,39 +34,78 @@ class MapTelemetryImpl : MapTelemetry {
...
-    if (TelemetryEnabler.State.ENABLED == telemetryState) {
-      telemetry.enable()
+
+    // check in the resources
+    val eventsTokenResId = getEventsResId(appContext, EVENTS_ACCESS_TOKEN_RESOURCE_NAME)
+    val eventsToken =
+      if (eventsTokenResId != 0) appContext.getString(eventsTokenResId) else accessToken
+
+    // TODO deferredDeliveryServiceOptions?
+    val eventsServiceOptions = EventsServerOptions(eventsToken, BuildConfig.MAPBOX_EVENTS_USER_AGENT, null)
```
pengdev, line 44: @tatiana-yan could you confirm that we can simply set null to the deferredDeliveryServiceOptions? Or do we need some extra logic here?

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -28,39 +34,78 @@ class MapTelemetryImpl : MapTelemetry {
...
+  private fun getEventsResId(context: Context, resourceName: String): Int = context.resources.getIdentifier(
+    resourceName,
+    "string",
+    context.packageName
+  )
+
+  private fun sendEvent(event: String) {
+    val eventAttributes = Value.fromJson(event)
+    // TODO: deferredOptions?
+    val mapEvent = eventAttributes.value?.let { Event(EventPriority.IMMEDIATE, it, null) }
```
pengdev, line 97: same here, can we just use null for deferredOptions or need some extra logic here? cc @tatiana-yan @tarigo

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
tatiana-yan: I think `null` ok for now because deferred telemetry is a new feature and it is safe to just switch from MME to CoreTelemetry now and probably introduce deferred options later.
It's very new and I think we do not have clear guidelines for it now (it's in progress).
@DorkMatter WDYT?

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: enableTelemetryCollection is also used in several other places, but I've removed `coreTelemetryState` variable.

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: could you clarify which token should we use now? just the access token passed from constructor?

tatiana-yan: Access token passed from constructor.

I have some additional code for setting token in my branches but it's just to have an easy filters for was events retrieval. For production we use just access token passed from constructor.

pengdev: thanks! updated 👍

module-telemetry/src/test/java/com/mapbox/maps/module/telemetry/MapTelemetryTest.kt
pengdev: telemetry and  mapboxTelemetry has been removed from the dependency and  `setDebugLoggingEnabled` is removed from codebase, so there wouldn't be a way to validate the no-op 🤷‍♂️

yunikkk: Ah, so method is not available at all, gotcha

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
tatiana-yan: I talked to @DorkMatter, she confirmed that we need `null` here

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
tatiana-yan: we need `null` here

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: Thanks for the confirmation!

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -17,50 +23,84 @@ class MapTelemetryImpl : MapTelemetry {
 
   private val appContext: Context
   private val accessToken: String
-  private val telemetry: MapboxTelemetry
+  private val eventsService: EventsServiceInterface
+  // We need telemetryService for location and metrics event which will be sent automatically if telemetryService is initialized.
```
kiryldz, line 27:
```suggestion
  // We need telemetryService for location and metrics events which will be sent automatically if telemetryService is initialized.
```

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -17,50 +23,84 @@ class MapTelemetryImpl : MapTelemetry {
...
+        BuildConfig.MAPBOX_SDK_IDENTIFIER,
+        BuildConfig.MAPBOX_SDK_VERSION
+      )
+    ) { eventsServiceError ->
+      eventsServiceError?.let {
+        logE(TAG, "EventsServiceError: $it")
+      }
+    }
+
+    val mapLoadEvent = MapEventFactory.buildMapLoadEvent(PhoneState(appContext))
```
kiryldz, line 82: a bit curious that in `onAppUserTurnstileEvent` we send turnstile event + additional one; if it's OK - shouldn't we actually send  `mapLoadEvent` if `eventsServiceError != null` above? otherwise the data may look a bit inconsistent..

tatiana-yan: we sent two events (turnstile + map.load) with mme, so please let's not modify it while `mme`->`core telemetry` switch

pengdev: this is the original logic, we shouldn't modify these two events during the migration.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 ## Features ✨ and improvements 🏁
 * Deprecated gesture settings `increaseRotateThresholdWhenPinchingToZoom` property. ([1632](https://github.com/mapbox/mapbox-maps-android/pull/1632))
+* Migrated telemetry APIs from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK implementation. ([1672](https://github.com/mapbox/mapbox-maps-android/pull/1672))
```
pengdev, line 8: @sbma44 could you help to review this release note?

plugin-locationcomponent/build.gradle.kts
yunikkk: Remove?

sdk-base/src/main/java/com/mapbox/maps/module/MapTelemetry.kt
```diff
@@ -23,8 +23,11 @@ interface MapTelemetry {
   fun disableTelemetrySession()
 
   /**
-   * Set the end-user selected state to participate or opt-out in telemetry collection.
+   * Set the debug logging enabled states.
```
yunikkk, line 26: Maybe remove comment if it's deprecated and no-op?

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -100,8 +135,10 @@ class MapTelemetryImpl : MapTelemetry {
    * @param interval the selected session interval
    * @return true if rotation session id was updated
    */
+  @Deprecated("setSessionIdRotationInterval has been deprecated and will do no operations")
```
yunikkk, line 138: Remove comment except for the Deprecated note?

buildSrc/src/main/kotlin/Project.kt
```diff
@@ -39,8 +39,7 @@ object Dependencies {
   const val mapboxGlNative = "com.mapbox.maps:android-core:${Versions.mapboxGlNative}"
   const val mapboxCoreCommon = "com.mapbox.common:common:${Versions.mapboxCommon}"
   const val mapboxOkHttp = "com.mapbox.common:okhttp:${Versions.mapboxCommon}"
-  const val mapboxAndroidCore = "com.mapbox.mapboxsdk:mapbox-android-core:${Versions.mapboxAndroidCore}"
-  const val mapboxAndroidTelemetry = "com.mapbox.mapboxsdk:mapbox-android-telemetry:${Versions.mapboxAndroidTelemetry}"
+//  const val mapboxAndroidCore = "com.mapbox.mapboxsdk:mapbox-android-core:${Versions.mapboxAndroidCore}"
```
yunikkk, line 42: Remove?

pengdev: We need to revert this one back as we had api dependency on the mapboxAndroidCore, otherwise it would be a breaking change. Will actually revert it now..

plugin-locationcomponent/build.gradle.kts
kiryldz: @pengdev if we're about to merge it - let's uncomment all this code related to this dependency

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: imo we shouldn't remove doc even its deprecated, but maybe state the API is deprecated in the API doc to make it clear. We also have checks to validate all the public APIs have documentation..

sdk-base/src/main/java/com/mapbox/maps/module/MapTelemetry.kt
pengdev: imo we shouldn't remove doc even its deprecated, but maybe state the API is deprecated in the API doc to make it clear. We also have checks to validate all the public APIs have documentation..

plugin-locationcomponent/build.gradle.kts
pengdev: yes, this is done now, this api dependency is brought back.

CHANGELOG.md
pengdev: got the 👍 offline, resolving the comment now.

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -17,50 +23,84 @@ class MapTelemetryImpl : MapTelemetry {
...
-    if (TelemetryEnabler.State.ENABLED == telemetryState) {
-      telemetry.enable()
+
+    val eventsServiceOptions = EventsServerOptions(accessToken, BuildConfig.MAPBOX_EVENTS_USER_AGENT, null)
+    this.eventsService = EventsService.getOrCreate(eventsServiceOptions)
+
+    this.telemetryService = TelemetryService.getOrCreate(eventsServiceOptions)
+
+    if (TelemetryUtils.getEventsCollectionState()) {
+      enableTelemetryCollection(true)
```
tatiana-yan, line 46: My point here was that `TelemetryUtils.getEventsCollectionState()` and `TelemetryUtils.setEventsCollectionState(...)` which happens inside `enableTelemetryCollection` gets/sets option to the same place. That's why it does not make sense to call 
```
if (TelemetryUtils.getEventsCollectionState()) {
      enableTelemetryCollection(true)
```
TelemetryUtils work with persistent settings and events service also checks `TelemetryUtils.getEventsCollectionState()` to determine if we need to send events. So no extra setting of events collection state is needed.

For MME we need this because event service did not check persistent settings for event collection state by itself.

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
yunikkk: Well what's the benefit of having some description and then saying method is no-op? IMO will only confuse users. Just having a short one saying `No-up, deprecated, do not use this method` will have a clear indication that this method should not be used

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
```diff
@@ -3,9 +3,15 @@ package com.mapbox.maps.module.telemetry
 import android.content.Context
 import android.os.Bundle
 import androidx.annotation.VisibleForTesting
-import com.mapbox.android.telemetry.*
+import androidx.annotation.VisibleForTesting.PRIVATE
+import com.google.gson.Gson
 import com.mapbox.annotation.module.MapboxModule
 import com.mapbox.annotation.module.MapboxModuleType
+import com.mapbox.bindgen.Value
+import com.mapbox.common.*
```
ank27, line 11: nit: maybe remove wildcard improts

ank27 (approved):

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: 👍 do you suggest we should remove 
```
if (TelemetryUtils.getEventsCollectionState()) {
      enableTelemetryCollection(true)
}
```
all together?

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: If we enforce this rule, we should probably enforce it with ktlint or detect, but will fix it for this PR anyway.

yunikkk (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Deprecated gesture settings `increaseRotateThresholdWhenPinchingToZoom` property. ([1632](https://github.com/mapbox/mapbox-maps-android/pull/1632))
 * Enable asynchronous tile uploader by default to improve animation performance. ([1679](https://github.com/mapbox/mapbox-maps-android/pull/1679))
 * Vector tiles without symbols are not hold for fade-out animation so that less amount of vector tiles are managed at a time. ([1679](https://github.com/mapbox/mapbox-maps-android/pull/1679))
+* Migrated telemetry APIs from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK implementation. ([1672](https://github.com/mapbox/mapbox-maps-android/pull/1672))
```
yunikkk, line 10: Should we drop a note about known issues with bearing = 0 ?

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
tatiana-yan: Yes. These get/set use the same place to get and set that's why this call does nothing. And I double-checked that everything works without this `enableTelemetryCollection` call: we get telemetry if it's enabled and we get turnstiles only if it's disabled.

tatiana-yan (approved):

CHANGELOG.md
pengdev: I don't think we ever linked known issues to the change log, we will patch the issue shortly in beta.2 or rc release.

module-telemetry/src/main/java/com/mapbox/maps/module/telemetry/MapTelemetryImpl.kt
pengdev: I also double checked, seems removing that logic doesn’t change the behaviour, will just remove it.

evil159 (approved):

PR #1675, opened 2022-09-17
Remove update android docs from update script

SevaZhukov:
Update android-docs will be executed by the Release train app. This approach is better than the current one because the app checks the version in https://docs.mapbox.com/android/maps/api/VERSION_NAME before creating a PR in the android-docs. It guarantees green checks and easy merge.


yunikkk:
@SevaZhukov could you share the link to the code which is used to update the android-docs with the new approach?

kiryldz (approved):

yunikkk (approved):

pengdev (approved):

PR #1681, opened 2022-09-21
Throw exception when constructing Value with NaN or infinite numbers.

pengdev:
### Summary of changes

Throw exception when constructing `Value` with not a finite number, so we get meaningful crash stacktrace early.

According to [JSON](https://www.json.org/json-en.html)  NAN isn't part of the JSON specification, and it will cause issues when we pass the Double.NaN to gl-native, where we serialise it to a JSON string.

## Pull request checklist:
    
### DIFF
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
@@ -55,23 +55,23 @@ internal object TypeUtils {
         Value(value)
       }
       is IntArray -> {
-        val valueArray = value.map { Value(it.toLong()) }
+        val valueArray = value.map { wrapToValue(it.toLong()) }
         Value(valueArray)
       }
       is BooleanArray -> {
-        val valueArray = value.map(::Value)
+        val valueArray = value.map(::wrapToValue)
         Value(valueArray)
       }
       is DoubleArray -> {
-        val valueArray = value.map(::Value)
+        val valueArray = value.map(::wrapToValue)
         Value(valueArray)
       }
       is FloatArray -> {
-        val valueArray = value.map { Value(it.toDouble()) }
+        val valueArray = value.map { wrapToValue(it.toDouble()) }
         Value(valueArray)
       }
       is LongArray -> {
-        val valueArray = value.map(::Value)
+        val valueArray = value.map(::wrapToValue)
         Value(valueArray)
       }
       is Array<*> -> {
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
@@ -44,6 +44,7 @@ internal object TypeUtils {
         Value(value)
       }
       is Double -> {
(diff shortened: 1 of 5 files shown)


extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
```diff
@@ -44,6 +44,7 @@ internal object TypeUtils {
         Value(value)
       }
       is Double -> {
+        require(!value.isNaN()) { "Value can not be Double.NaN" }
```
jush, line 47: Should we do the same for `Float` below? and their array counterparts?

jush (changes requested):

extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
jush: Actually, the same will happen if INFINITY is passed.
See Moshi tests: https://github.com/square/moshi/blob/master/moshi/src/test/java/com/squareup/moshi/MoshiTest.java#L296

pengdev: Fixed, please take a look again.

jush (approved):

kiryldz (approved):

extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
```diff
@@ -44,32 +44,34 @@ internal object TypeUtils {
...
       is Float -> {
+        require(value.isFinite()) { "Value can not be Float.NaN, Float.POSITIVE_INFINITY or Float.NEGATIVE_INFINITY" }
         Value(value.toDouble())
       }
       is Long -> {
         Value(value)
       }
       is IntArray -> {
-        val valueArray = value.map { Value(it.toLong()) }
+        val valueArray = value.map { wrapToValue(it.toLong()) }
```
yunikkk, line 58: Why do you need wrap int though?

extension-style/src/test/java/com/mapbox/maps/extension/style/utils/TypeUtilsTest.kt
```diff
@@ -58,12 +58,42 @@ class TypeUtilsTest {
...
+  fun wrapToValue_Float_POSITIVE_INFINITY() {
+    testwrapToValue(Float.POSITIVE_INFINITY)
+  }
+
+  @Test(expected = IllegalArgumentException::class)
+  fun wrapToValue_Float_NEGATIVE_INFINITY() {
+    testwrapToValue(Float.NEGATIVE_INFINITY)
+  }
+
   private fun testwrapToValue(value: Any) {
```
yunikkk, line 97:
```suggestion
  private fun testWrapToValue(value: Any) {
```

extension-style/src/main/java/com/mapbox/maps/extension/style/utils/TypeUtils.kt
pengdev: just want to make it consistent, so we can add future assertion to the primitive types as well.

yunikkk (approved):

Issue #1540, opened 2022-07-27
Crash in MapboxTelemetry : Caused by com.mapbox.common.module.provider.MapboxInvalidModuleException 


dh-hasif:
## Environment

- Android OS version: 10 , 11
- Devices affected: Realme, OPPO, Nubia
- Maps SDK Version: 10.3.0

## Observed behavior and steps to reproduce
We are reporting several crashes in production from some set of users, and the stack trace is as follows 

`Caused by com.mapbox.common.module.provider.MapboxInvalidModuleException
MapTelemetry has been excluded from build but a correct alternative was not provided. Make sure that: - Your custom module implements com.mapbox.maps.module.MapTelemetry. - Your custom module class is annotated with @a(b.MapTelemetry). - You've provided a `ModuleProvider` instance to Mapbox_MapTelemetryModuleConfiguration#setModuleProvider before initializing the library, unless `enableConfiguration` flag is set to false and your implementation has a public, non-arg constructor or is a Kotlin object.`

`com.mapbox.common.module.provider.MapboxModuleProvider.createModule (MapboxModuleProvider.kt:88)
com.mapbox.maps.MapController.dispatchTelemetryTurnstileEvent (MapController.kt:214)
com.mapbox.maps.MapController.<init> (MapController.kt:77)
com.mapbox.maps.MapView.<init> (MapView.kt:101)
com.mapbox.maps.MapView.<init> (MapView.kt:69)
com.mapbox.maps.MapView.<init> (MapView.kt:63)
java.lang.reflect.Constructor.newInstance0 (Constructor.java)
java.lang.reflect.Constructor.newInstance (Constructor.java:343)
android.view.LayoutInflater.createView (LayoutInflater.java:852)
android.view.LayoutInflater.createViewFromTag (LayoutInflater.java:1008)
android.view.LayoutInflater.createViewFromTag (LayoutInflater.java:959)
android.view.LayoutInflater.rInflate (LayoutInflater.java:1138)
android.view.LayoutInflater.rInflateChildren (LayoutInflater.java:1099)
android.view.LayoutInflater.inflate (LayoutInflater.java:680)
android.view.LayoutInflater.inflate (LayoutInflater.java:532)
com.foodora.courier.map.enabled.presentation.MapboxFragment.onCreateView (MapboxFragment.kt:153)
androidx.fragment.app.Fragment.performCreateView (Fragment.java:2995)
[...]

PR #1693, opened 2022-09-26
Remove Telemetry module setup 

tobrun:
### Summary of changes

This setup was driven by the mapbox-base modularization using annotations. The SDK depends directly on the MapTelemetryImpl versus the MapTelemetry interface module. This patch needs backporting to v10.9 release branch.

### User impact (optional)
Normal SDK usage will not be impacted, only custom setup will require a workaround.

## Pull request checklist:
    

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 Mapbox welcomes participation and contributions from everyone.
 # main
+* Remove module setup driven by the mapbox-base modularization using annotations. The SDK depends directly on the MapTelemetryImpl versus the MapTelemetry interface module. ([#1693](https://github.com/mapbox/mapbox-maps-android/pull/1693))
```
kiryldz, line 5: if we know it should be cherry-picked to v10.9 - perhaps create the header here already and place this in correct section? btw - is this bug fix or the feature? 🤔

kiryldz:
@tobrun can we do smth about failing `verify-api-metalava`? guess we don't need telemetry module anymore
```
* What went wrong:
A problem was found with the configuration of task ':module-telemetry:checkApiMetalava' (type 'JavaExec').
  - Type 'org.gradle.api.tasks.JavaExec' property '$1' specifies file '/src/module-telemetry/api/metalava.txt' which doesn't exist.
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -247,44 +253,6 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
-  private fun paramsProvider(type: MapboxModuleType): Array<ModuleProviderArgument> {
-    val context = mapInitOptions.context
-    return when (type) {
-      MapboxModuleType.CommonLibraryLoader -> arrayOf(
-        ModuleProviderArgument(
-          Context::class.java,
-          context.applicationContext
-        )
-      )
-      MapboxModuleType.CommonHttpClient -> arrayOf()
```
yunikkk, line 275: Wonder if we should also proceed and remove the module system completely?

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -72,10 +70,18 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     this.mapboxMap =
       MapProvider.getMapboxMap(nativeMap, nativeObserver, mapInitOptions.mapOptions.pixelRatio)
     this.mapboxMap.renderHandler = renderer.renderThread.renderHandlerThread.handler
+
+    // telemetry
+    val telemetry: MapTelemetry = MapTelemetryImpl(
```
pengdev, line 75: As we reference the `MapTelemetryImpl` directly in the sdk, do we want to actually move the code inside the `module-telemetry` to the sdk itself? considering telemetry is not going to be a module after the change, keeping the module separate seems a bit confusing.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: Also wonder if we don't fully remove the module setup, and there're still other modules in the codebase, we might still encounter the similar issue in other code paths, e.g https://github.com/mapbox/mapbox-maps-android/issues/1540

yunikkk: > As far as I know common SDK still uses the `CommonHttpClient` module setup, but seems they have their own `paramsProvider`.

Right - but if there is no clear reasons to do so it will be nice to align and avoid exact issues that are still possible with reflection

pengdev: The module setup is widely used across common and nav sdks, it exists for the flexibility to let the user to provide their own overwrites to some functionalities. And we have [this](https://docs.mapbox.com/android/maps/guides/migrate-to-v10/#example-replace-modular-http-stack) section in the migration guide to use this system to overwrite the http stack. If we want to totally remove the modular setup, it would be a bigger change, and probably in v11.

yunikkk: Hmm, indeed, then suppose we need to keep the public API for those users. 
Can we though somehow modify the way Modules are generated so that it wouldn't involve reflection? Should be possible, e.g. add some setters to the [CommonModuleProvider](https://github.com/mapbox/mapbox-sdk-common/blob/97257fc308917dd07a4568847b87f1fe201958c6/platform/android/common/src/main/java/com/mapbox/common/core/module/CommonSingletonModuleProvider.kt#L15) and call them from the generated modules

PR #1707, opened 2022-09-29
Skip checks for changelog branches

SevaZhukov:
image

It's not necessary to check changelog branches by build workflow. This hack can save 30 minutes in the release process


.circleci/config.yml
```diff
@@ -604,67 +716,83 @@ jobs:
...
-            python3 scripts/ci-circleci-start-pipeline.py --token ${MOBILE_METRICS_TOKEN} \
-              --origin-slug mapbox/mapbox-maps-android \
-              --current-branch $CIRCLE_BRANCH \
-              --target-slug mapbox/mapbox-maps-android-private \
-              --sync-internal True
+      - when:
+          condition:
+            not:
+              matches:
+                pattern: "^add-changelog.*"
```
alexander-kulikovskii, line 783: maybe move such pattern to the global parameters?

.circleci/config.yml
```diff
@@ -395,207 +395,319 @@ jobs:
...
-              echo "$(./mbx-ci github reader token)" > gh_token.txt
-              ./scripts/check-changelog.sh
-            fi
-      - save-gradle-cache
-      - store-results
+      - when:
+          condition:
+            not:
+              matches:
+                pattern: "^add-changelog.*"
```
pengdev, line 402: can we save this pattern as a global parameter for reuse?

.circleci/config.yml
```diff
@@ -395,207 +395,319 @@ jobs:
...
-          name: check changelog
-          command: |
-            if [[ ! -z "${CIRCLE_PULL_REQUEST}" ]]; then
-              echo "$(./mbx-ci github reader token)" > gh_token.txt
-              ./scripts/check-changelog.sh
-            fi
-      - save-gradle-cache
-      - store-results
+      - when:
+          condition:
```
alexander-kulikovskii, line 399: will it be enough to add condition for this job? all other jobs depends on it

SevaZhukov: all other jobs have the same condition

alexander-kulikovskii: I mean that `verify-code` is the initial point for all other jobs. And if you filter it, then all other jobs will not start.

pengdev: since our PR merge strategy requires the other jobs, these required jobs has to pass 🙈

pengdev: wonder if there's a way to simplify these conditions, could we add a base job/workflow?

SevaZhukov: All jobs execute because they have alternative step **exit 0**

SevaZhukov: I don't know have to avoid empty job runs because we have a strong branch rule

alexander-kulikovskii (approved):

.circleci/config.yml
jush: IMHO, it might be good to add a comment explaining why the `exit 0` is needed. Otherwise, I'm afraid we will forget it (especially when adding new jobs)

jush (approved):
LGTM with minor comment

yunikkk (approved):

PR #1708, opened 2022-09-29
MAPSAND-481 adding widget does not refresh map view

paulrenn67:
### Summary of changes

When a `Widget` was added to the `MapView`, there was no explicit refresh made, so if there were no render events already queued, then the widgets would not be immediately displayed.  Main fix here is simply to call `scheduleRepaint` in the `MapController.addWidget` call.

Also added an `onBitmapLoaded` callback for each Widget so that the caller can know when it is safe to call `Bitmap.recycle()`, if needed.

## Pull request checklist:
    

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
```diff
@@ -14,6 +14,7 @@ internal class BitmapWidgetRenderer(
   private val position: WidgetPosition,
   private val marginX: Float,
   private val marginY: Float,
+  internal val onRendered: (Bitmap) -> Unit,
```
yunikkk, line 17: Imo it's not really `onRendered` - bitmap is just loaded to the texture, which is rendered after it many times. Maybe rename like `onBitmapLoaded`?

kiryldz (commented):
add unit test for that?

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
paulrenn67: Renamed as suggested

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -218,6 +218,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
 
   override fun addWidget(widget: Widget) {
     renderer.renderThread.addWidget(widget)
+    renderer.scheduleRepaint()
   }
 
   override fun removeWidget(widget: Widget) = renderer.renderThread.removeWidget(widget)
```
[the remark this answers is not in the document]
jush: What about other things like `setRotation`, `setTranslation` and `updateBitmap` in the `BitmapWidget`?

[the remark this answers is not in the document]
paulrenn67: Good question, but I think after the widget has been added, then updates to any existing widgets should get drawn as part of `MapboxRenderThread.draw`?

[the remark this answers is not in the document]
paulrenn67: But true, it does not force a redraw.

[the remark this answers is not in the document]
paulrenn67: In the current architecture it is impossible to make a redraw when `BitmapWidget's` `setRotation`, `setTranslation` and `updateBitmap` methods are called directly, since there is no reference to the `MapController`.  Maybe we leave this for another day.

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidget.kt
```diff
@@ -10,6 +10,7 @@ import com.mapbox.maps.MapboxExperimental
  * @param position position of widget
  * @param marginX horizontal margin in pixels
  * @param marginY vertical margin in pixels
+ * @param onBitmapLoaded callback when widget is first rendered, the Bitmap may be recycled if desired.
```
jush, line 13:
```suggestion
 * @param onBitmapLoaded callback when widget is first loaded, the Bitmap may be recycled if desired.
```

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidget.kt
```diff
@@ -20,16 +21,20 @@ open class BitmapWidget @JvmOverloads constructor(
   ),
   marginX: Float = 0f,
   marginY: Float = 0f,
+  onBitmapLoaded: (Bitmap) -> Unit = {},
```
jush, line 24: Instead of lambda (which is seen as `FunctionN` in Java) should we use a proper [SAM](https://kotlinlang.org/docs/fun-interfaces.html)?

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
```diff
@@ -246,6 +247,7 @@ internal class BitmapWidgetRenderer(
       GLUtils.texImage2D(GLES20.GL_TEXTURE_2D, 0, it, 0)
       GLES20.glBindTexture(GLES20.GL_TEXTURE_2D, 0)
 
+      onBitmapLoaded(it)
```
jush, line 250: Could we add a test to check this is called? (I guess it needs to be a instrumentation test)
And it will most likely only work when building the widget for first time. `updateBitmap` won't trigger it since we don't have yet the force refresh.

jush (changes requested):

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidget.kt
jush: But, if you do so, don't we lose the nice benefit of being able to write:
```Kotlin
// Creating an instance using lambda
val isEven = IntPredicate { it % 2 == 0 }
```
instead of:
```Kotlin
// Creating an instance of a class
val isEven = object : IntPredicate {
   override fun accept(i: Int): Boolean {
       return i % 2 == 0
   }
}
```

paulrenn67: Hmmm `fun interface` does provide a nicer syntax for the caller. We'll use that for now at least.

jush (approved):

pengdev:
Note this PR might also affect the Android Auto extension's compatibility with Maps SDK, cc @kmadsen

yunikkk (approved):

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetListener.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Listener for BitmapWidget events.
+ */
+@MapboxExperimental
+fun interface BitmapWidgetListener {
+  /**
+   * Invoked when the supplied Bitmap has been loaded.
+   * The Bitmap object is no longer required by the BitmapWidget,
+   * and can be recycled if needed during the callback.
+   */
+  fun onLoaded(bitmap: Bitmap)
```
kmadsen, line 16: What is the reason for exposing this? It looks like good functionality to recycle the bitmap when it has loaded onto the map.

But why expose this to downstream? It seems a developer should "give ownership" of the bitmap to the `BitmapWidget`

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
```diff
@@ -246,6 +247,7 @@ internal class BitmapWidgetRenderer(
       GLUtils.texImage2D(GLES20.GL_TEXTURE_2D, 0, it, 0)
       GLES20.glBindTexture(GLES20.GL_TEXTURE_2D, 0)
 
+      listener?.onLoaded(it)
```
kmadsen, line 250:
```suggestion
      bitmap.recycle()
```

https://github.com/mapbox/mapbox-maps-android/pull/1708/files#r984878027
^ Related to this comment.

Alternative is to not expose a listener and just recycle it. If the developer wants to share the bitmap across multiple `BitmapWidgetRenderer` instances, then it seems Mapbox should build a `MultiBitmapWidgetRenderer`

kmadsen: Considering the desire to share a bitmap elsewhere, for example a bitmap in a View icon and the same icon for the map. But those are typically separate icons, so it does not seem like something to optimize for.

IMO if you turn this into a simple bug fix you could back port it to older versions. For example, what if you wanted to add this to the `10.4` release?

extension-androidauto/src/main/java/com/mapbox/maps/extension/androidauto/widgets/CompassWidget.kt
```diff
@@ -32,4 +33,5 @@ class CompassWidget(
   position = position,
   marginX = marginX,
   marginY = marginY,
+  listener = BitmapWidgetListener { it.recycle() }
```
kmadsen, line 36:
```suggestion
```

If the current approach is desired this will create a semver incompatibility and delay any potential stable releases

I think you should prefer solutions that would not cause semver. For example, the `BitmapWidget` is already an open class so why not give the ability to override the functionality there??

``` kotlin
@MapboxExperimental
open class BitmapWidget @JvmOverloads constructor(
  bitmap: Bitmap,
  position: WidgetPosition = WidgetPosition(
    vertical = WidgetPosition.Vertical.TOP,
    horizontal = WidgetPosition.Horizontal.LEFT,
  ),
  marginX: Float = 0f,
  marginY: Float = 0f,
) : Widget() {
  override val renderer = BitmapWidgetRenderer(
    bitmapWidget = this, // <<< now the renderer can call onLoaded
    bitmap = bitmap,
    position = position,
    marginX = marginX,
    marginY = marginY,
  )

  // This is called when the BitmapWidgetRenderer is done with the bitmap
  fun onLoaded() {
    bitmap.recycle()
  }

  /**
   * Update bitmap widget uses.
   */
  fun updateBitmap(bitmap: Bitmap) {
    renderer.updateBitmap(bitmap)
  }

  override fun setTranslation(translationX: Float, translationY: Float) {
    renderer.setTranslation(translationX = translationX, translationY = translationY)
  }

  override fun setRotation(angleDegrees: Float) {
    renderer.setRotation(angleDegrees = angleDegrees)
  }
}
```

kmadsen: This is why we really need to merge this first https://github.com/mapbox/mapbox-maps-android/pull/1706

That way the incompatibility is caught by the build system

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
kmadsen: If it is still desired to expose the `onLoaded`, can you please consider an alternative approach that will not create an incompatibility?

https://github.com/mapbox/mapbox-maps-android/pull/1708/files#r984894170

kmadsen (changes requested):
I don't like to block the change but please check alternative solutions before merging.

I'm imagining some workarounds for android auto, but this does not appear to be a good reason to cause an incompatibility between versions

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
paulrenn67: We don't own the Bitmap, and there is nothing in the API to say that we do.  The caller could be using the Bitmap for other purposes, so recycling it in the SDK is maybe not the best approach.

extension-androidauto/src/main/java/com/mapbox/maps/extension/androidauto/widgets/CompassWidget.kt
paulrenn67: Hi @kmadsen, since you mention that the class is `open`, it got me thinking whether we implement the `onLoaded` as you suggest, but have a subclassed `RecycledBitmapWidget` which calls `bitmap.recycle` in its `onLoaded`.  The user is then free to decide which class to use. Our own `CompassWidget` and `LogoWidget` could inherit from `RecycledBitmapWidget`. Would this be acceptable?

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
paulrenn67: Please review latest commit

extension-androidauto/api/extension-androidauto.api
```diff
@@ -71,12 +71,12 @@ public final class com/mapbox/maps/extension/androidauto/MapboxCarMapSurface {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class com/mapbox/maps/extension/androidauto/widgets/CompassWidget : com/mapbox/maps/renderer/widget/BitmapWidget {
+public final class com/mapbox/maps/extension/androidauto/widgets/CompassWidget : com/mapbox/maps/renderer/widget/RecyclingBitmapWidget {
```
kmadsen, line 74:
```suggestion
public final class com/mapbox/maps/extension/androidauto/widgets/CompassWidget : com/mapbox/maps/renderer/widget/BitmapWidget {
```

This is still a non-backwards compatible change.

the androidauto extension intends to use version `10.9` and "be upgradable to 10.10`

Changing the name of this will make it so the minimum version of androidauto extension is 10.10.

kmadsen: Can we make sure to add this to `10.9`?

extension-androidauto/src/main/java/com/mapbox/maps/extension/androidauto/widgets/LogoWidget.kt
```diff
@@ -27,7 +27,7 @@ class LogoWidget constructor(
   ),
   marginX: Float = 20f,
   marginY: Float = 20f,
-) : BitmapWidget(
+) : RecyclingBitmapWidget(
```
kmadsen, line 30:
```suggestion
) : BitmapWidget(
```

Implementations in `androidauto` will need to use the upgradable object.

We can adopt the new api when it becomes available but it is going to be locked on [10.5](https://github.com/mapbox/mapbox-maps-android/pull/1706), it'll depend stable release versions when they become available

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidgetRenderer.kt
kmadsen: Resolving here ! thanks for addressing

extension-androidauto/src/main/java/com/mapbox/maps/extension/androidauto/widgets/CompassWidget.kt
kmadsen: Yeah this is better with a `onLoaded` open function in my opinion.

`RecycledBitmapWidget` could have been good if it was the original implementation, unfortunately the bug fix also requires a code change. So technically will not use `RecycledBitmapWidget` until it becomes part of a final release like `10.9.0` or `10.10.0`

If you want to make it so this is a "free fix upon upgrade". Then make it part of the `BitmapWidget`

extension-androidauto/src/main/java/com/mapbox/maps/extension/androidauto/widgets/LogoWidget.kt
kmadsen: We won't use the `RecyclingBitmapWidget` until we upgrade the androidauto extension minimum version to a version that includes the new class.

Another option is to backport this to all the releases (for example, `10.5.1`, `10.6.1`, `10.7.1`, `10.8.1`). That is an obvious headache that would require a major issue, and I think we all agree this issue does not qualify for that.

So our other option is to just remove  `RecyclingBitmapWidget` from the AA changes. We can use it becomes available. Developers have the option to upgrade and implement it themselves.

pengdev (approved):

kiryldz (approved):

kmadsen (approved):
Thanks for addressing feedback!

yunikkk (approved):

PR #1716, opened 2022-10-03
Fix scale bar text being cut

jush:
### Summary of changes

Fixes rendering issues where the scale bar text was being cut on the sides:

Before:

After:

Also added a new example activity to showcase scale bar customization.

## Pull request checklist:
    
### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
+++ b/plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
@@ -270,7 +270,6 @@ class ScaleBarImpl : ScaleBar, View {
             textPaint.textAlign = Paint.Align.LEFT
             strokePaint.textAlign = Paint.Align.LEFT
           }
-
           else -> {
             textPaint.textAlign = Paint.Align.CENTER
             strokePaint.textAlign = Paint.Align.CENTER


paulrenn67:
Fix looks good, but wondering about the alignment of the start/end labels, at least the "100m" seems to be differently aligned than in the original screenshot.  Should we try to centrally align the labels above the start/end of the scalebar? In this case, the "0" should be more slightly to the left (does not look correctly aligned even in the original) and the "100m" should be more to the right.  The "50m" label seems to be correctly centred.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -294,11 +301,33 @@ class ScaleBarImpl : ScaleBar, View {
...
 
-  private fun drawText(canvas: Canvas, text: String, x: Float, y: Float) {
+  private fun drawText(canvas: Canvas, text: String, x: Float, y: Float, lastUnitBarText: Boolean) {
+    var safeX = x
+
+    // As an optimization only check if it goes beyond right view for the last unit bar text
+    if (lastUnitBarText) {
+      // Check if it goes beyond the right margin of the view
+      val textWidthPx = textPaint.measureText(text)
+      val textMaxRightPx = x + strokePaint.strokeWidth + when (textPaint.textAlign) {
```
yunikkk, line 311: As far as I understood `strokeWidth` was the main reason for the text being cut?

jush:
> Fix looks good, but wondering about the alignment of the start/end labels, at least the "100m" seems to be differently aligned than in the original screenshot. Should we try to centrally align the labels above the start/end of the scalebar? In this case, 
The alignment has not changed, but in order for the text to fix we need to "move" it away from the margins. See this video of the behaviour:

https://user-images.githubusercontent.com/691172/193564972-97b7e97a-3f4f-4aef-b9de-00d387599b6c.mp4

The scale bar layout has a defined that we should honour. Therefore, the only way I could think to _easily fix_ it is to move the label. If we want a more advanced solution we could try to figure out if the labels fit and if not then reduce the amount of unit bars we show. We will probably need to do that to fix the issue where the labels overlap each other. But that I'd say is for another PR.

> the "0" should be more slightly to the left (does not look correctly aligned even in the original) and the "100m" should be more to the right. The "50m" label seems to be correctly centred.

The "0" is left aligned but given the horizontal bearing it renders it a bit to the right. (See https://stackoverflow.com/a/7579469)

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -114,7 +114,7 @@ class ScaleBarImpl : ScaleBar, View {
       strokePaint.textSize = value.textSize
       scaleTable = if (value.isMetricUnits) metricTable else imperialTable
       unit = if (value.isMetricUnits) METER_UNIT else FEET_UNIT
-      strokePaint.strokeWidth = value.textBorderWidth
+      strokePaint.strokeWidth = if (value.showTextBorder) value.textBorderWidth else 0F
```
kiryldz, line 117: is this related to text being cut fix?

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: No, the main reason was that the `x` we are passing is based on the position of the current unit bar and that sometimes makes it grow beyond the right margin

jush: But `strokeWidth` needs to be taken into account also, since `measureText` doesn't.

app/src/main/res/values-w600dp/scale_bar_config.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="scaleBarRatio" format="float" type="fraction">0.2</item>
+    <dimen name="scaleBarMarginRight">60dp</dimen>
+    <dimen name="scaleBarMarginBottom">8dp</dimen>
+    <!-- "bottom | right" See plugin-scalebar/src/main/res/values/attrs.xml -->
+    <integer name="scaleBarGravity">0x55</integer>
```
kiryldz, line 7: are we actually suggesting the users to have similar config? looks not very user-friendly tbh 😄

app/src/main/res/layout/activity_scale_bar.xml
```diff
@@ -0,0 +1,24 @@
...
+    android:id="@+id/mapView"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    app:mapbox_cameraTargetLat="61.489557808394814"
+    app:mapbox_cameraTargetLng="23.735379149896005"
+    app:mapbox_cameraZoom="14.940602577615023"
+
+    app:mapbox_scaleBarBorderWidth="1dp"
+    app:mapbox_scaleBarHeight="5dp"
+    app:mapbox_scaleBarGravity="@integer/scaleBarGravity"
```
kiryldz, line 13: will actually `bottom|end` work here?

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: Yes, if we want to render border around the text we need to take it into account, otherwise we set it to 0 so we avoid an `if` check on every draw. See code a few lines below:
```Kotlin
val textMaxRightPx = x + strokePaint.strokeWidth + when (textPaint.textAlign) {
```

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -270,13 +270,20 @@ class ScaleBarImpl : ScaleBar, View {
...
           }
         }
 
-        drawText(canvas, distanceText, (unitBarWidth * rectIndex), textSize)
+        drawText(
+          canvas = canvas,
+          text = distanceText,
+          x = (unitBarWidth * rectIndex),
+          y = textSize,
+          lastUnitBarText = rectIndex == rectCount
```
kiryldz, line 285:
```suggestion
          lastUnitBarText = (rectIndex == rectCount)
```
to better align with L283

app/src/main/res/values-w600dp/scale_bar_config.xml
jush: I am not aware of any other way to use resources to define gravity.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -294,11 +301,33 @@ class ScaleBarImpl : ScaleBar, View {
...
+      // Check if it goes beyond the right margin of the view
+      val textWidthPx = textPaint.measureText(text)
+      val textMaxRightPx = x + strokePaint.strokeWidth + when (textPaint.textAlign) {
+        Paint.Align.LEFT -> textWidthPx
+        Paint.Align.CENTER -> textWidthPx / 2
+        Paint.Align.RIGHT, null -> 0F
+      }
+      if (textMaxRightPx > width) {
+        // Move it away from right margin enough to fit
+        safeX -= textMaxRightPx - width
```
kiryldz, line 318:
```suggestion
        safeX -= (textMaxRightPx - width)
```

app/src/main/res/layout/activity_scale_bar.xml
jush: I haven't tried. Moreover, I doubt we support RTL for this view.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -294,11 +301,33 @@ class ScaleBarImpl : ScaleBar, View {
...
+      }
+    }
+
+    // Check if it goes beyond the left margin of the view
+    if (x - strokePaint.strokeWidth / 2 < 0) {
```
kiryldz, line 323: is it ok that we check `x` here but operate with `safeX` below?

jush: You're right. It's better to use `safeX`.

kiryldz: and also for better readability I'd use `if (x < trokePaint.strokeWidth / 2)` similar to L316

jush: Hmmm I personally think that:
```Kotlin
if (x < trokePaint.strokeWidth / 2)
```
it's a bit obscure. 
I need to think more to figure out why we're comparing a position (`x` or `safeX`) with a width (`strokeWidth`). Somehow, to me, mine reads easier: to position `safeX` I remove a width and then check if it goes beyond the left minimum position (`0`).

But happy to change it to your proposal if others also prefer it.

app/src/main/res/layout/activity_scale_bar.xml
jush: Removed the separate xml values to avoid confusion.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Mapbox welcomes participation and contributions from everyone.
 # main
 
+## Bug fixes 🐞
```
kiryldz, line 6: this will be the part of upcoming rc - could prepare the header for it already? otherwise if we land some fix that we _don't_ want to cherry-pick - they will get shuffled under the `main` section

app/src/main/java/com/mapbox/maps/testapp/examples/ScaleBarActivity.kt
```diff
@@ -0,0 +1,16 @@
+package com.mapbox.maps.testapp.examples
+
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import com.mapbox.maps.testapp.R
+
+/**
+ * Activity to showcase scale bar custom configuration using xml attributes.
+ */
+class ScaleBarActivity : AppCompatActivity() {
```
kiryldz, line 10: should we create similar activity on iOS for parity and making sure that our scalebar customizations are aligned? @mapbox/maps-ios

kiryldz (approved):
Would be nice to have dedicated pixel match as the tailwork, otherwise LGTM

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -270,13 +270,20 @@ class ScaleBarImpl : ScaleBar, View {
             textPaint.textAlign = Paint.Align.LEFT
             strokePaint.textAlign = Paint.Align.LEFT
           }
+
```
ank27, line 273: nit: empty line

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -294,11 +301,33 @@ class ScaleBarImpl : ScaleBar, View {
...
+      // Check if it goes beyond the right margin of the view
+      val textWidthPx = textPaint.measureText(text)
+      val textMaxRightPx = x + strokePaint.strokeWidth + when (textPaint.textAlign) {
+        Paint.Align.LEFT -> textWidthPx
+        Paint.Align.CENTER -> textWidthPx / 2
+        Paint.Align.RIGHT, null -> 0F
+      }
+      if (textMaxRightPx > width) {
+        // Move it away from right margin enough to fit
+        safeX -= (textMaxRightPx - width)
```
ank27, line 318: can you also check if this works expected for RTL layout as well ?

ank27 (approved):

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: Seems to work ok

CHANGELOG.md
jush: Please check [5e38872](https://github.com/mapbox/mapbox-maps-android/pull/1716/commits/5e388728e870de7f86b8f0affb94b3a3489165f3). Not sure if that's what it's needed.

PR #1717, opened 2022-10-03
Fix raster dem crash style reload

yunikkk:
### Summary of changes

When we load styles many times in a row there's a possibility that `style data : sources` event from the previous style will arrive after new style has been started loading. Here we add the fix for it and actually wait for the `style data : style` event after the style load has been initiated and skip all the other events. 

This is actually part of MAPSNAT-516 that couldn't be fixed in gl-native. 

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 
 * Fix scale bar text being cut ([1716](https://github.com/mapbox/mapbox-maps-android/pull/1716))
+## Bug fixes 🐞
+* Fix possible crash adding terrain using Style DSL after consecutive style changes ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
 
 # 10.8.1 September 30, 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 
 * Fix scale bar text being cut ([1716](https://github.com/mapbox/mapbox-maps-android/pull/1716))
-## Bug fixes 🐞
 * Fix possible crash adding terrain using Style DSL after consecutive style changes ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
 
 # 10.8.1 September 30, 2022
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## Bug fixes 🐞
 
 * Fix scale bar text being cut ([1716](https://github.com/mapbox/mapbox-maps-android/pull/1716))
-* Fix possible crash adding terrain using Style DSL after consecutive style changes ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
+* Fix possible crash when adding terrain using Style DSL after consecutive style changes. ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
 
 # 10.8.1 September 30, 2022
 
--- a/sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
@@ -4,7 +4,10 @@ import androidx.test.annotation.UiThreadTest
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
+import com.mapbox.maps.extension.style.expressions.dsl.generated.literal
+import com.mapbox.maps.extension.style.sources.generated.rasterDemSource
 import com.mapbox.maps.extension.style.style
+import com.mapbox.maps.extension.style.terrain.generated.terrain
 import junit.framework.TestCase.*
 import org.junit.*
 import org.junit.runner.RunWith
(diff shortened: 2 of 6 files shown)


kiryldz (commented):
Could you please add
 - PR description
 - bug label
 - fix ktlint
 - add changelog

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Mapbox welcomes participation and contributions from everyone.
 # main
 
+## Bug fixes 🐞
```
kiryldz, line 6: should be included in rc so please rebase on latest main and include in rc section instead of main section

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
...
+* Fix possible crash adding terrain using Style DSL after consecutive style changes ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
```
kiryldz, line 7:
```suggestion
* Fix possible crash when adding terrain using Style DSL after consecutive style changes. ([1717](https://github.com/mapbox/mapbox-maps-android/pull/1717))
```

sdk/src/test/java/com/mapbox/maps/StyleObserverTest.kt
```diff
@@ -314,4 +315,32 @@ class StyleObserverTest {
 
     verify { styleSourcesCallback.onStyleLoaded(any()) }
   }
+
+  @Test
+  fun onStyleDataSourcesAndSpritesEventsAreIgnoredUntilStyleDataLoadedNotCalledOnStyleChange() {
```
kiryldz, line 320: looks a bit too short 😄 would be nice to add more details in the test name!

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
```diff
@@ -169,6 +172,53 @@ class StyleLoadTest {
...
+              mapView.postDelayed(this, 1)
+            }
+          }
+        }
+        mapView.post(runnable)
+
+        mapView.onStart()
+      }
+    }
+    countDownLatch.throwExceptionOnTimeoutMs()
```
kiryldz, line 219: based on Firebase this test fails with timeout now

sdk/src/test/java/com/mapbox/maps/StyleObserverTest.kt
```diff
@@ -314,4 +315,32 @@ class StyleObserverTest {
...
+    styleObserver.onStyleDataLoaded(StyleDataLoadedEventData(0, 0, StyleDataType.SPRITE))
+
+    verifyNo { styleSourcesCallback.onStyleLoaded(any()) }
+    verifyNo { styleSpritesCallback.onStyleLoaded(any()) }
+
+    styleObserver.onStyleDataLoaded(StyleDataLoadedEventData(0, 0, StyleDataType.STYLE))
+    styleObserver.onStyleDataLoaded(StyleDataLoadedEventData(0, 0, StyleDataType.SOURCES))
+    styleObserver.onStyleDataLoaded(StyleDataLoadedEventData(0, 0, StyleDataType.SPRITE))
+
+    verifyOnce { styleSourcesCallback.onStyleLoaded(any()) }
```
alexander-kulikovskii, line 343: and nice to add
```
    verifyOnce { styleCallback.onStyleLoaded(any()) }
```

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
```diff
@@ -169,6 +172,53 @@ class StyleLoadTest {
...
   }
 
+  @Test
+  fun testTerrainDoesNotCrashOnMultipleStyleLoad() {
+    fun loadStyleWithTerrain(
+      uri: String = Style.DARK,
+      terrainSource: String = "TERRAIN_SOURCE",
+      demSourceUri: String = "mapbox://mapbox.mapbox-terrain-dem-v1"
+    ) {
+      mapboxMap.loadStyle(
```
alexander-kulikovskii, line 182: nit:
maybe add extra delay? potentially we can have super fast internet and super fast device.

yunikkk: Well worst case if it happens without any delay - then the crashes actually had happenned. Could you elaborate more why do you think delay is needed?

alexander-kulikovskii: >Could you elaborate more why do you think delay is needed?

to guarantee a delay between different styles load

yunikkk: But we don't need a delay, that's the thing, for this specific test - it happens when there is no delay so that style load events could collide with higher probability

alexshalamov:
Would be nice to add a test for a following scenario (pseudo code):

```kotlin
map.loadStyle(URL) { DSL runtime modifications }
map.onStyleDataLoaded({
  map.loadStyleJson(InvalidStyleJSON)
})
// verify that original style (URL) is loaded and runtime modifications are applied 
```

yunikkk:
> Would be nice to add a test for a following scenario (pseudo code):
> 
> ```kotlin
> map.loadStyle(URL) { DSL runtime modifications }
> map.onStyleDataLoaded({
>   map.loadStyleJson(InvalidStyleJSON)
> })
> // verify that original style (URL) is loaded and runtime modifications are applied 
> ```

Such flow will not be handled correctly at the moment, turns out. 
We only keep the latest loadStyle DSL settings, overwriting the previous ones.

sdk/src/test/java/com/mapbox/maps/StyleObserverTest.kt
yunikkk: Added

kiryldz (approved):

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
alexander-kulikovskii: I mean some delay in style loaded callbacks. To make listeners crossover effect more noticeable.

alexander-kulikovskii (approved):

sdk/src/androidTest/java/com/mapbox/maps/StyleLoadTest.kt
yunikkk: But I can add bunch of randomly delayed calls, all right, no problem

yunikkk: Added 100 launches with random delay of 0-5 millis

PR #1728, opened 2022-10-05
Scale bar: avoid text overlapping and avoid rendering unless needed

jush:
### Summary of changes

There were cases where the scale bar labels could overlap each other.

This PR:
- Checks if there's an overlap and reduces the amount of unit bars if needed.
- As a performance improvement we now check if `distancePerPixel` change has an impact in the scale bar visualization before enqueueing a re-rendering.

### User impact (optional)

Impact on the overlapping:
| Before | After |
| ----- | ----- |
|  |   |

Impact on the performance:
- **Before**: Scale bar was rendered almost constantly: [legacy_scalebar.webm](https://user-images.githubusercontent.com/691172/194079419-54c66d26-c71a-4f50-83ca-f0357ffcca09.webm)
- **After**: Scale bar doesn't re-render since there are now significant changes: [new_scalebar.webm](https://user-images.githubusercontent.com/691172/194079004-860c26bd-506f-42f0-9511-c703c563eb6a.webm)

## Pull request checklist:
    

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -138,20 +125,37 @@ class ScaleBarImpl : ScaleBar, View {
...
 
+  private var currentSegmentsConfiguration: SegmentsConfiguration? = null
+
   /**
    * distancePerPixel in current location latitude
    */
   override var distancePerPixel = 0F
     set(value) {
       if (field != value) {
+        // TODO: Check the value is always different than field for Imperial units due to this conversion
```
kiryldz, line 136: do we want to actually merge it with following todo? perhaps we can create the Jira ticket and mention it here instead

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -138,20 +125,37 @@ class ScaleBarImpl : ScaleBar, View {
...
+            scaleTable = scaleTable,
+            textPaint = textPaint,
+            strokeWidth = strokePaint.strokeWidth,
+            unit = unit,
+            rightMarginPx = width
+          )
+          // Only force a re-draw if the bar UI has changed enough by checking current
+          // segmentsConfiguration against new ones
+          if (newSegmentsConfiguration != currentSegmentsConfiguration) {
+            if (!refreshHandler.hasMessages(MSG_RENDER_ON_DEMAND)) {
```
kiryldz, line 154: collapse into one `if` check?

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
jush: I just fixed it.

kiryldz (commented):
In general pretty hard to review the actual logic - just as a side note - can we make sure that for all division operations we don't divide by zero?

kiryldz (approved):
hard to say from the code if the algorithm is correct for all the cases but let's iterate 😄

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -300,81 +326,156 @@ class ScaleBarImpl : ScaleBar, View {
...
+    var distance: Float = pair.first.toFloat()
+    var rectCount: Int = pair.second
+    var unitDistance: Float = distance / rectCount
+
     // When maxDistance is small (i.e. high zoom levels near the poles) then
     // the `distance` might be bigger than maxDistance. This loop will keep removing
     // bar divisions (rectCount) until it fits
     while (unitDistance * rectCount > maxDistance && rectCount > 0) {
       rectCount--
+      distance = unitDistance * rectCount
```
yunikkk, line 367: Maybe move out of the loop below it?

jush: I'm not sure I follow, given the following scenario:
- `unitDistance * rectCount` is less than `maxDistance`
- `rectCount` is > 0
wouldn't that change the logic if I take it out or in the else?

The value could still be the same (given that `unitDistance` is `distance / rectCount`) but there could also be float conversion issues.

jush:
> In general pretty hard to review the actual logic

I agree... I'm happy to add more comments in the code if there's some area that would need some clarification.

> just as a side note - can we make sure that for all division operations we don't divide by zero?

There are only 12 divisions in that class, most of them with constants and then a couple for `rectCount` and `distancePerPixel`. `rectCount` is taken from the `scaleTable` and then if I decrement it I always try to check for `0`. `distancePerPixel` is provided by the map itself, I'm not sure if it can ever be `0`.

jush:
> hard to say from the code if the algorithm is correct for all the cases but let's iterate 😄

It is indeed hard to say. The input is a bunch of things coming from maps, developer settings, and also from the UI (width, margin...) 😭

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
yunikkk: Indeed, I didn't take into account that loop may not run at all.

plugin-scalebar/src/main/java/com/mapbox/maps/plugin/scalebar/ScaleBarImpl.kt
```diff
@@ -300,81 +326,156 @@ class ScaleBarImpl : ScaleBar, View {
...
+    text: String,
+    anchorXPx: Float,
+    alignment: Paint.Align,
+    textPaint: Paint,
+    strokeWidth: Float,
+    rightMarginPx: Int,
+  ): Triple<Float, Float, Float> {
+    var safeAnchorX = anchorXPx
+
+    val textWidthPx = textPaint.measureText(text)
```
yunikkk, line 440: Wonder how fast is this nowadays, remember there were issues with the speed of text measurement, though it was probably long time ago. At the same time if it's called on every frame - maybe worth to optimize it somehow. Since the texts are actually the same most of the times - maybe introduce some cache for it?

jush: I could measure it but I don't think it's a problem.

I might remember wrong but I think `TextView` measurements were expensive and Android introduced [PrecomputedText](https://developer.android.com/reference/android/text/PrecomputedText) to help with it.

jush: I measure it in a Pixel 3 and it takes on average 0.02milliseconds (1386 times). I would not do any improvement yet.

alexander-kulikovskii (approved):

yunikkk (approved):

PR #1744, opened 2022-10-10
Fix view annotations triggering excessive onMeasure [MAPSAND-516]

yunikkk:
Problem was caused calling bringToFront() on every update of the annotations, that happens on any map animation or gesture.

bringToFront() was needed to guarantee that : 
- view annotations are displayed in the correct order (same as items in the `onViewAnnotationPositionsUpdate(@NonNull List<ViewAnnotationPositionDescriptor> positions);` callback). This order changes if some ViewAnnotation is being selected, for example
- plugins (compass, ruler etc) are displayed on top of ViewAnnotations

Also I've added separate FrameLayout to the MapView on top of the actual surface/texture view but below any other plugin - so that removing or adding new view annotation does not interfere with the plugins at all. 
MapView adds a SurfaceView/TextureView that actually displays map at the index 0 here, then additional views are added on top of it, thus before the fix the layout was looking the following way :

```
FrameLayout MapView : 
 - SurfaceView/TextureView
 - ViewAnnotation1
 - ViewAnnotation2
 - ...
 - CompassPlugin
 - LogoPlugin
```

After the fix it becomes :

```
FrameLayout MapView : 
 - SurfaceView/TextureView
 - FrameLayout for view annotations
   - ViewAnnotation1
   - ViewAnnotation2
   - ...
 - CompassPlugin
 - LogoPlugin
```

### User impact (optional)

Before changes, [simple map layout](https://github.com/mapbox/mapbox-maps-android/blob/main/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationShowcaseActivity.kt) with 1 view annotation, browsing map (so onMeasure is called every frame): 
image

onMeasure calls take `117.767` nanos out of `1.409.936` that choreographer.doFrame spent, roughly 8.3%. 

After changes, same layout and map : 
image

Tested on the Android emulator.
No onMeasure calls at all.
The difference will be more noticeable the more complex layout is used.

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 # main
 ## Bug fixes 🐞
-* Fix frequent layout invalidation caused by view annotations calling View.bringToFront(). ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
+* Fix frequent layout invalidation caused by view annotations calling `View.bringToFront()`. ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
 
 # 10.9.0-rc.1 October 07, 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,12 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+
+# 10.9.0
+
 ## Bug fixes 🐞
 * Fix frequent layout invalidation caused by view annotations calling `View.bringToFront()`. ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
+
 
 # 10.9.0-rc.1 October 07, 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
-
+## Bug fixes 🐞
+* Fix frequent layout invalidation caused by view annotations calling View.bringToFront(). ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
 
 # 10.9.0-rc.1 October 07, 2022
 
--- a/sdk/build.gradle.kts
+++ b/sdk/build.gradle.kts
@@ -75,7 +75,6 @@ dependencies {
   implementation(Dependencies.kotlin)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
-  implementation(Dependencies.androidxRecyclerView)
   testImplementation(Dependencies.junit)
   testImplementation(Dependencies.mockk)
   testImplementation(Dependencies.androidxTestCore)
--- a/sdk/build.gradle.kts
+++ b/sdk/build.gradle.kts
@@ -75,6 +75,7 @@ dependencies {
   implementation(Dependencies.kotlin)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
+  implementation(Dependencies.androidxRecyclerView)
   testImplementation(Dependencies.junit)
   testImplementation(Dependencies.mockk)
   testImplementation(Dependencies.androidxTestCore)
--- a/sdk/src/androidTest/java/com/mapbox/maps/TestUtils.kt
+++ b/sdk/src/androidTest/java/com/mapbox/maps/TestUtils.kt
@@ -1,14 +1,15 @@
 package com.mapbox.maps
 
 import android.view.View
+import android.widget.FrameLayout
 import androidx.core.view.children
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
-internal fun MapView.hasChildView(view: View) = children.any { it == view }
+internal fun FrameLayout.hasChildView(view: View) = children.any { it == view }
 
-internal fun MapView.getChildViewIndex(view: View) = children.indexOf(view)
+internal fun FrameLayout.getChildViewIndex(view: View) = children.indexOf(view)
 
 internal fun CountDownLatch.throwExceptionOnTimeoutMs(timeoutMs: Long = DEFAULT_LATCH_TIMEOUT_MS) {
   if (!await(timeoutMs, TimeUnit.MILLISECONDS)) {
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -15,21 +15,20 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArraySet
 
 internal class ViewAnnotationManagerImpl(
-  mapView: MapView
+  mapView: MapView,
+  private val viewAnnotationsLayout: FrameLayout = FrameLayout(mapView.context),
 ) : ViewAnnotationManager, ViewAnnotationPositionsUpdateListener {
 
-  private val annotationsRootView: FrameLayout = FrameLayout(mapView.context).also {
-    it.layoutParams = FrameLayout.LayoutParams(
-      ViewGroup.LayoutParams.MATCH_PARENT,
-      ViewGroup.LayoutParams.MATCH_PARENT,
-    )
-  }
   private val mapboxMap: MapboxMap = mapView.getMapboxMap()
   private val renderThread = mapView.mapController.renderer.renderThread
 
   init {
+    mapView.layoutParams = FrameLayout.LayoutParams(
+      ViewGroup.LayoutParams.MATCH_PARENT,
+      ViewGroup.LayoutParams.MATCH_PARENT,
+    )
     // place the view annotations above the map (index 0) but below the compass, ruler and other plugin views
-    mapView.addView(annotationsRootView, 1)
+    mapView.addView(viewAnnotationsLayout, 1)
     mapView.requestDisallowInterceptTouchEvent(false)
     mapboxMap.setViewAnnotationPositionsUpdateListener(this)
   }
@@ -54,7 +53,7 @@ internal class ViewAnnotationManagerImpl(
     asyncInflateCallback: (View) -> Unit
   ) {
     validateOptions(options)
-    asyncInflater.inflate(resId, annotationsRootView) { view, _, _ ->
+    asyncInflater.inflate(resId, viewAnnotationsLayout) { view, _, _ ->
       asyncInflateCallback.invoke(prepareViewAnnotation(view, options))
     }
   }
@@ -64,7 +63,7 @@ internal class ViewAnnotationManagerImpl(
     options: ViewAnnotationOptions
   ): View {
     validateOptions(options)
-    val view = LayoutInflater.from(annotationsRootView.context).inflate(resId, annotationsRootView, false)
+    val view = LayoutInflater.from(viewAnnotationsLayout.context).inflate(resId, viewAnnotationsLayout, false)
     return prepareViewAnnotation(view, options)
   }
 
@@ -253,7 +252,7 @@ internal class ViewAnnotationManagerImpl(
     // as OnGlobalLayoutListener does not cover cases for View.INVISIBLE properly
     val onDrawListener = ViewTreeObserver.OnDrawListener {
       if (viewAnnotation.handleVisibilityAutomatically) {
-        val isAndroidViewVisible = (inflatedView.visibility == View.VISIBLE)
+        val isAndroidViewVisible = inflatedView.visibility == View.VISIBLE
         if (
           (isAndroidViewVisible && viewAnnotation.isVisible) ||
           (!isAndroidViewVisible && viewAnnotation.visibility == ViewAnnotationVisibility.INVISIBLE) ||
@@ -329,22 +328,7 @@ internal class ViewAnnotationManagerImpl(
     // as per current implementation when view annotation was added with WRAP_CONTENT dimension AND
     // allowOverlap = false - core will notify only about this particular view and thus we will remove
     // all the others and then restore later resulting in a quick blink
-
-    // FIXME - other bugs are possible, this does not cover all cases, e.g. 3 annotations and the last one blinks :
-    // Positions : 3
-    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
-    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
-    //   pos : [identifier: 48, width: -2, height: -2, leftTopCoordinate: [x: -2.147483218E9, y: -4.2949667920000005E9]]
-    // Positions : 3
-    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
-    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
-    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
-    // Positions : 1
-    //   pos : [identifier: 49, width: -2, height: -2, leftTopCoordinate: [x: -2.147483362E9, y: -4.2949661230000005E9]]
-    // Positions : 3
-    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
-    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
-    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
+    // FIXME - other bugs are possible, this does not cover all cases
     val wrapContentWithAllowOverlapFalseViewAdded = positionDescriptorUpdatedList.size == 1 &&
       (positionDescriptorUpdatedList[0].width == WRAP_CONTENT || positionDescriptorUpdatedList[0].height == WRAP_CONTENT)
     // firstly delete views that do not belong to the viewport if it's not specific use-case from above
@@ -355,7 +339,7 @@ internal class ViewAnnotationManagerImpl(
             // if view is invisible / gone we don't remove it so that visibility logic could
             // still be handled by OnGlobalLayoutListener
             if (annotation.view.visibility == View.VISIBLE) {
-              annotationsRootView.removeView(annotation.view)
+              viewAnnotationsLayout.removeView(annotation.view)
               updateVisibilityAndNotifyUpdateListeners(annotation, ViewAnnotationVisibility.INVISIBLE)
             }
           }
@@ -380,9 +364,9 @@ internal class ViewAnnotationManagerImpl(
           }
         }
         if (!currentlyDrawnViewIdSet.contains(descriptor.identifier) &&
-          annotationsRootView.indexOfChild(annotation.view) == -1
+          viewAnnotationsLayout.indexOfChild(annotation.view) == -1
         ) {
-          annotationsRootView.addView(annotation.view, annotation.viewLayoutParams)
+          viewAnnotationsLayout.addView(annotation.view, annotation.viewLayoutParams)
           updateVisibilityAndNotifyUpdateListeners(
             annotation,
             if (annotation.view.visibility == View.VISIBLE)
@@ -459,7 +443,7 @@ internal class ViewAnnotationManagerImpl(
   }
 
   private fun remove(internalId: String, annotation: ViewAnnotation) {
-    annotationsRootView.removeView(annotation.view)
+    viewAnnotationsLayout.removeView(annotation.view)
     updateVisibilityAndNotifyUpdateListeners(annotation, ViewAnnotationVisibility.INVISIBLE)
     annotation.view.removeOnAttachStateChangeListener(annotation.attachStateListener)
     annotation.attachStateListener = null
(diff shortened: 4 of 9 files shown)


sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -308,23 +309,80 @@ internal class ViewAnnotationManagerImpl(
...
+          // don't care about the contents
+        }
+      }
+    )
+
     // as per current implementation when view annotation was added with WRAP_CONTENT dimension AND
     // allowOverlap = false - core will notify only about this particular view and thus we will remove
     // all the others and then restore later resulting in a quick blink
+
+    // FIXME - other bugs are possible, this does not cover all cases, e.g. 3 annotations and the last one blinks :
```
yunikkk, line 360: @kiryldz found similar bug with view annotations blinking

sdk/build.gradle.kts
```diff
@@ -75,6 +75,7 @@ dependencies {
   implementation(Dependencies.kotlin)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
+  implementation(Dependencies.androidxRecyclerView)
```
jush, line 78: what's your opinion on bringing just the `DiffUtil` and `ListUpdateCallback` (not sure it would be needed) source code instead of the whole recycler view library?

This adds one more dependency we (and consumers of our library) need to worry about.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -6,35 +6,40 @@ import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
...
+      ViewGroup.LayoutParams.MATCH_PARENT,
+    )
+  }
   private val mapboxMap: MapboxMap = mapView.getMapboxMap()
-  private val viewPlugins = mapView.mapController.pluginRegistry.viewPlugins
   private val renderThread = mapView.mapController.renderer.renderThread
 
   init {
+    // place the view annotations above the map (index 0) but below the compass, ruler and other plugin views
+    mapView.addView(annotationsRootView, 1)
```
paulrenn67, line 34: Question, as I'm not sure of the answer without coding it :D, if the user inflates the MapView from an XML which contains other Views, are we sure that we should add annotationsRootView at index 1? Or should we check how many Views the ViewGroup contains, and add it at the end?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -308,23 +309,80 @@ internal class ViewAnnotationManagerImpl(
...
   private fun positionAnnotationViews(
     positionDescriptorCoreList: List<ViewAnnotationPositionDescriptor>
   ) {
+    val positionDiffCallback = PositionDiffCallback(
+      oldList = positionDiffCallback.newList,
+      newList = positionDescriptorCoreList
+    )
+
+    var needToReorderZ = false
+    DiffUtil.calculateDiff(positionDiffCallback).dispatchUpdatesTo(
```
alexander-kulikovskii, line 333: wdyt about async diff calculation? 
maybe if you have 1000+ elements there might be some performance issues

sdk/build.gradle.kts
yunikkk: Controversial. 
Moving will mean we have to maintain the code ourselves, check whether new fixes are available upstream every release, port them etc. 
It's also `androidx` dependency and is pretty standard in the android world, most of the apps will already contain it. But let me check how will the method count change without core shrink to make more intentional decision.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: It's not the xml that user inflates, it's the FrameLayout of the MapView. MapView adds a SurfaceView/TextureView that actually displays map at the index 0 [here](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/MapView.kt#L109), then additional views are added on top of it [here](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt#L71), thus before the fix the layout was looking the following way : 
```
FrameLayout MapView : 
 - SurfaceView/TextureView
 - CompassPlugin
 - LogoPlugin
 - ViewAnnotation1
 - ViewAnnotation2
 - ...
 ```

After the fix it becomes : 
```
FrameLayout MapView : 
 - SurfaceView/TextureView
 - CompassPlugin
 - LogoPlugin
 - FrameLayout for view annotations
   - ViewAnnotation1
   - ViewAnnotation2
 - ...
```

File: sdk/build.gradle.kts (hunk 2 above)
Comment (jush) [reply sdk/build.gradle.kts:78]: Also, let's keep in mind that this might have an impact on users that already have recyclerview dependency but doesn't match ours.

File: sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerZOrderingTest.kt [hunk 5]
```diff
@@ -0,0 +1,264 @@
...
+          "1",
+          "3",
+        ),
+        /* new descriptors */
+        descriptors(
+          "1",
+          "2",
+        ),
+        /* change z order */
+        false,
```
jush, line 115: How come this is `false` but from empty to `1,3,2` is true (https://github.com/mapbox/mapbox-maps-android/pull/1744/files#diff-0e7802c1bcba87eaaea61f8bf7ba216fe00ab2f8eae240f5dfbcadd3429f9470R101) ?

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerZOrderingTest.kt
```diff
@@ -0,0 +1,264 @@
...
+  companion object {
+    private fun descriptors(vararg identifiers: String) = identifiers.map {
+      ViewAnnotationPositionDescriptor(
+        it, 0, 0, ScreenCoordinate(0.0, 0.0)
+      )
+    }
+
+    @JvmStatic
+    @Parameterized.Parameters
+    fun data() = listOf(
```
alexander-kulikovskii, line 47: - two empty lists
- equal lists

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -439,5 +455,57 @@ internal class ViewAnnotationManagerImpl(
...
+            // skip equal items
+            if (positionDescriptorCurrentList[currentIndex] == positionDescriptorUpdatedList[updatedIndex]) {
+              currentIndex++
+              updatedIndex++
+              continue
+            }
+
+            // if removed descriptors list contains the item from the new list - it means
+            // the element was not removed but moved, need to invalidate Z order
+            if (removedDescriptors.contains(positionDescriptorUpdatedList[currentIndex].identifier)) {
```
alexander-kulikovskii, line 488:
```suggestion
            if (removedDescriptors.contains(positionDescriptorUpdatedList[updatedIndex].identifier)) {
```

yunikkk: Nice catch! Added a test case that crashes here also.

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerZOrderingTest.kt
yunikkk: Added!

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerZOrderingTest.kt
yunikkk: Hmm, true is from 1-2-3 to 1-3-2, if you mean this test case
image

yunikkk: Ah, also it's true for any update that is adding more elements that there were before. So empty to anything non-empty will produce true.

yunikkk: 1-3 to 1-2 does also assumes that new view will be placed into the layout and onMeasure will be called, we just do not optimize calculations this particular edge case.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -250,16 +254,20 @@ internal class ViewAnnotationManagerImpl(
     val onDrawListener = ViewTreeObserver.OnDrawListener {
       if (viewAnnotation.handleVisibilityAutomatically) {
         val isAndroidViewVisible = (inflatedView.visibility == View.VISIBLE)
-        if ((isAndroidViewVisible && viewAnnotation.isVisible) ||
-          (!isAndroidViewVisible && viewAnnotation.visibility == ViewAnnotationVisibility.INVISIBLE)
+        if (
+          (isAndroidViewVisible && viewAnnotation.isVisible) ||
+          (!isAndroidViewVisible && viewAnnotation.visibility == ViewAnnotationVisibility.INVISIBLE) ||
+          (!isAndroidViewVisible && viewAnnotation.visibility == ViewAnnotationVisibility.VISIBLE_AND_NOT_POSITIONED)
```
alexander-kulikovskii, line 260: nit:
```suggestion
          (!isAndroidViewVisible && (viewAnnotation.visibility == ViewAnnotationVisibility.INVISIBLE) || (viewAnnotation.visibility == ViewAnnotationVisibility.VISIBLE_AND_NOT_POSITIONED))
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -308,31 +316,54 @@ internal class ViewAnnotationManagerImpl(
...
+    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
+    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
+    val wrapContentWithAllowOverlapFalseViewAdded = positionDescriptorUpdatedList.size == 1 &&
+      (positionDescriptorUpdatedList[0].width == WRAP_CONTENT || positionDescriptorUpdatedList[0].height == WRAP_CONTENT)
     // firstly delete views that do not belong to the viewport if it's not specific use-case from above
     if (!wrapContentWithAllowOverlapFalseViewAdded) {
-      currentViewsDrawnMap.keys.forEach { id ->
-        if (positionDescriptorCoreList.indexOfFirst { it.identifier == id } == -1) {
+      currentlyDrawnViewIdSet.forEach { id ->
+        if (positionDescriptorUpdatedList.indexOfFirst { it.identifier == id } == -1) {
```
alexander-kulikovskii, line 353: if I understand correctly in worst case it can be O(N)? 
maybe save somehow id + position?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -308,31 +316,54 @@ internal class ViewAnnotationManagerImpl(
...
+    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
+    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
+    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
+    // Positions : 1
+    //   pos : [identifier: 49, width: -2, height: -2, leftTopCoordinate: [x: -2.147483362E9, y: -4.2949661230000005E9]]
+    // Positions : 3
+    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
+    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
+    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
+    val wrapContentWithAllowOverlapFalseViewAdded = positionDescriptorUpdatedList.size == 1 &&
```
alexander-kulikovskii, line 348: ?
```suggestion
    val wrapContentWithAllowOverlapFalseViewAdded = positionDescriptorUpdatedList.size >= 1 &&
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: I found it less readable, actually, thought about it as well

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Don't think we can do that, workaround specifically handles case when there's single view. I'll also create a ticket to gl-native to fix the root issue

sdk/build.gradle.kts
yunikkk: Ended up implementing manual algo for calculations

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -308,31 +316,54 @@ internal class ViewAnnotationManagerImpl(
...
+    // Positions : 3
+    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
+    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
+    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
+    // Positions : 1
+    //   pos : [identifier: 49, width: -2, height: -2, leftTopCoordinate: [x: -2.147483362E9, y: -4.2949661230000005E9]]
+    // Positions : 3
+    //   pos : [identifier: 42, width: 282, height: 196, leftTopCoordinate: [x: 99.99999999998583, y: 368.9999999999887]]
+    //   pos : [identifier: 44, width: 282, height: 196, leftTopCoordinate: [x: 148.00000000000566, y: 616.9999999999659]]
+    //   pos : [identifier: 45, width: 282, height: 196, leftTopCoordinate: [x: 483.0000000000149, y: 566.9999999999759]]
```
kiryldz, line 347: please create the separate issue for that in gl-native and remove those comments from here

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
-
+## Bug fixes 🐞
+* Fix frequent layout invalidation caused by view annotations calling View.bringToFront(). ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
```
kiryldz, line 7:
```suggestion
* Fix frequent layout invalidation caused by view annotations calling `View.bringToFront()`. ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -13,28 +13,32 @@ import com.mapbox.maps.viewannotation.ViewAnnotation.Companion.USER_FIXED_DIMENS
...
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArraySet
-import kotlin.collections.HashMap
 
 internal class ViewAnnotationManagerImpl(
-  private val mapView: MapView
+  mapView: MapView
 ) : ViewAnnotationManager, ViewAnnotationPositionsUpdateListener {
 
+  private val annotationsRootView: FrameLayout = FrameLayout(mapView.context).also {
```
kiryldz, line 21:
```suggestion
  private val viewAnnotationsLayout: FrameLayout = FrameLayout(mapView.context).also {
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -372,24 +405,23 @@ internal class ViewAnnotationManagerImpl(
             }
           }
         }
-        hiddenViewMap[annotation.view]?.let { zIndex ->
-          annotation.view.translationZ = zIndex
-          hiddenViewMap.remove(annotation.view)
+
+        if (unpositionedViews.remove(annotation.view)) {
+          annotation.view.visibility = View.VISIBLE
```
kiryldz, line 410: checked the code - seems that if user has actually done the following:
1. change view from invisible -> visible.
2. view will be added to `unpositionedViews`.
3. before core callback user has made it visible -> invisible.
4. we will return from `OnDrawListener` then
5. core callback executed - and we will actually make the view visible.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Ok

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Removed

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Theoretically valid concern - though hard to reproduce and not sure what would be the fix. We don't have a way to detect that user called `invisible = true` while view is already invisible. Also, it seems the very same issue had existed before this PR so I suggest to ignore it for now.

kiryldz: Discussed offline that in current implementation fixing that seems impossible.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -13,28 +13,31 @@ import com.mapbox.maps.viewannotation.ViewAnnotation.Companion.USER_FIXED_DIMENS
...
   private val mapboxMap: MapboxMap = mapView.getMapboxMap()
-  private val viewPlugins = mapView.mapController.pluginRegistry.viewPlugins
   private val renderThread = mapView.mapController.renderer.renderThread
 
   init {
+    mapView.layoutParams = FrameLayout.LayoutParams(
+      ViewGroup.LayoutParams.MATCH_PARENT,
+      ViewGroup.LayoutParams.MATCH_PARENT,
+    )
+    // place the view annotations above the map (index 0) but below the compass, ruler and other plugin views
```
pengdev, line 30: Is it expected that view annotation should be placed below the compass/scalebar? Do we want to expose some configurations to adjust the z-order of the view annotations?

kiryldz: that was not possible before that and in order to control Z index we have the concept of selected state for view annotatotions so it seems that we're totally OK here.

kiryldz (approved):
One last thing before merging - will you be able to record CPU load and update PR description before / after this PR? I believe improvement should be visible.

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
-
+## Bug fixes 🐞
```
kiryldz, line 6: Also - as we're porting this to v10.9 branch - we should then create that header now and remove this entry from `#main`

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: @peng this kind of customisations makes sense to me, actually, though outside the scope of this PR I guess

yunikkk: Dropped a ticket https://mapbox.atlassian.net/browse/MAPSAND-535

CHANGELOG.md
yunikkk: I'll move it in the v10.9 branch probably

kiryldz: then we might forget about it in `main` branch and accidently include it in v10.10.0-beta.1 that will grab changelog from `#main` section

pengdev: yeah, agree to add 10.9.0 change log section now, so we can just cherry-pick this to the release branch, and our release script should be able to pick it up automatically.

sdk/src/androidTest/java/com/mapbox/maps/TestUtils.kt
```diff
@@ -1,14 +1,15 @@
...
 import androidx.core.view.children
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
-internal fun MapView.hasChildView(view: View) = children.any { it == view }
+internal fun FrameLayout.hasChildView(view: View) = children.any { it == view }
 
-internal fun MapView.getChildViewIndex(view: View) = children.indexOf(view)
+internal fun FrameLayout.getChildViewIndex(view: View) = children.indexOf(view)
```
pengdev, line 12: As discussed in the last kotlin group session, it's generally a bad practice to extend Android system APIs, but I guess it's less of a problem for internal methods, and in tests.

pengdev (approved):
LGTM % nits

CHANGELOG.md
yunikkk: Okay

PR #1745, opened 2022-10-12
Fix touch event with many annotations MAPSAND-459

alexander-kulikovskii:
### Summary of changes

Fix an issue when touch events didn't pass through the view annotation.

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Mapbox welcomes participation and contributions from everyone.
 
 # main
 
+## Bug fixes 🐞
+Fix an issue when touch events didn't pass through clickable view annotations. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
+
 
 # 10.9.0 October 21, 2022
 
@@ -16,10 +19,6 @@ Mapbox welcomes participation and contributions from everyone.
 * Add support to set location puck opacity. ([1659](https://github.com/mapbox/mapbox-maps-android/pull/1659))
 * Support `pitch` and `distanceFromCenter` filters in symbol layers. ([1662](https://github.com/mapbox/mapbox-maps-android/pull/1662))
 * Migrate telemetry APIs from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK implementation. ([1672](https://github.com/mapbox/mapbox-maps-android/pull/1672))
-## Bug fixes 🐞
-Fix an issue when touch events didn't pass through clickable view annotations. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
-
-# 10.9.0
 
 ## Bug fixes 🐞
 * Fix frequent layout invalidation caused by view annotations calling `View.bringToFront()`. ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Support `pitch` and `distanceFromCenter` filters in symbol layers. ([1662](https://github.com/mapbox/mapbox-maps-android/pull/1662))
 * Migrate telemetry APIs from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK implementation. ([1672](https://github.com/mapbox/mapbox-maps-android/pull/1672))
 ## Bug fixes 🐞
-Fix an issue when touch events didn't pass through the view annotation. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
+Fix an issue when touch events didn't pass through clickable view annotations. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
 
 # 10.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Mapbox welcomes participation and contributions from everyone.
 * Add support to set location puck opacity. ([1659](https://github.com/mapbox/mapbox-maps-android/pull/1659))
 * Support `pitch` and `distanceFromCenter` filters in symbol layers. ([1662](https://github.com/mapbox/mapbox-maps-android/pull/1662))
 * Migrate telemetry APIs from [mobile-events-android](https://github.com/mapbox/mapbox-events-android) to common SDK implementation. ([1672](https://github.com/mapbox/mapbox-maps-android/pull/1672))
+## Bug fixes 🐞
+Fix an issue when touch events didn't pass through the view annotation. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
+
+# 10.9.0
 
 ## Bug fixes 🐞
 * Fix frequent layout invalidation caused by view annotations calling `View.bringToFront()`. ([1744](https://github.com/mapbox/mapbox-maps-android/pull/1744))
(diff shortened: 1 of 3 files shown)


yunikkk:
@alexander-kulikovskii could you describe the PR and also what about the issues with slow gestures being not recognized you've mentioned?

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -296,9 +301,54 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
   }
 
+  override fun onInterceptTouchEvent(event: MotionEvent): Boolean {
+    return when (event.actionMasked) {
+      MotionEvent.ACTION_DOWN -> {
+        interceptedActionDownEvent = MotionEvent.obtain(event)
+        false
+      }
+      MotionEvent.ACTION_POINTER_DOWN -> {
+        interceptedActionPointerDownEvent = MotionEvent.obtain(event)
```
yunikkk, line 330: Since there's no difference how `interceptedActionDownEvent` and `interceptedActionDownEvent` are handled - maybe just add them to a collection? 
Also - seems case with 3+ fingers is not intercepted correctly. Zoom in-out works fine using 3 and more fingers if the gesture is started on the map, and not when it starts on view annotation. Suppose its due to the fact that we overwrite `interceptedActionPointerDownEvent` on 3 and following events

alexander-kulikovskii: > Since there's no difference how interceptedActionDownEvent and interceptedActionDownEvent are handled - maybe just add them to a collection?

What is the purpose of such an approach?

> Suppose its due to the fact that we overwrite interceptedActionPointerDownEvent on 3 and following events

Yeah, it's a bug. Also try to store `interceptedActionPointerDownEvent` when it's ONLY empty. But unsuccessfully

yunikkk: > What is the purpose of such an approach?

It will be less code, especially if we'll take into account 3-4-... multitouch events

alexander-kulikovskii:
> also what about the issues with slow gestures being not recognized you've mentioned?

it's expected behaviour for the current gestures implementation.

alexander-kulikovskii:
> > also what about the issues with slow gestures being not recognized you've mentioned?
> 
> it's expected behaviour for the current gestures implementation.

done

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Mapbox welcomes participation and contributions from everyone.
 
 # main
 
+## Bug fixes 🐞
+Fix an issue when touch events didn't pass through the view annotation. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
```
kiryldz, line 8:
```suggestion
Fix an issue when touch events didn't pass through clickable view annotations. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
```

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -296,9 +300,45 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
+      MotionEvent.ACTION_DOWN -> {
+        interceptedActionList.add(MotionEvent.obtain(event))
+        false
+      }
+      MotionEvent.ACTION_POINTER_DOWN -> {
+        interceptedActionList.add(MotionEvent.obtain(event))
+        false
+      }
+      MotionEvent.ACTION_MOVE -> {
+        interceptedActionList.any { it.hypot(event) > touchSlop }
```
kiryldz, line 325: I'm not sure this logic is correct. If I remember correctly we actually need to cache `ACTION_POINTER_DOWN` event just to explicitly pass it to gestures why threshold check here should check cached `ACTION_DOWN` with new event. Now it's quite hard for me to understand what happening - we're checking `any` element.

Could you please clarify what's the logic now? Using the list could actually accumulate events as we're polling only the first inside `onTouchEvent`.

alexander-kulikovskii: test with list size logging. this value does not exceed the number of fingers set on the screen when swiping.
suggest leave as is

kiryldz (approved):

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -296,9 +300,45 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    */
   @SuppressLint("ClickableViewAccessibility")
   override fun onTouchEvent(event: MotionEvent): Boolean {
+    interceptedActionList.firstOrNull()?.let {
+      try {
+        return mapController.onTouchEvent(it)
+      } finally {
+        it.recycle()
+        interceptedActionList.removeFirst()
```
yunikkk, line 308: Shouldn't they all be cleared?

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -296,9 +300,45 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
...
+        // In general, we don't want to intercept touch events. They should be
+        // handled by the child view.
+        false
+      }
+    }
+  }
+
+  private fun MotionEvent.hypot(event: MotionEvent): Int {
+    return hypot(
+      x = (this?.x ?: event.x) - event.x,
```
yunikkk, line 344: nit : since `this` is not nullable - can be simplified to `x = this.x - event.x,`

sdk/src/main/java/com/mapbox/maps/MapView.kt
yunikkk: @alexander-kulikovskii to me it seems that for multi touch gestures we need to compare them by pointer id, since on multi touch gestures MotionEvent contain info about all the pointers. 
Moreover it seems we might not even need a list - just keep the latest event detected by ACTION_DOWN / ACTION_POINTER_DOWN, and then iterate through it's pointers, e.g. 
```
        for (i in 0 until event.pointerCount) {
// motion event contains info about all the pointers
          println(" id ${event.getPointerId(i)} / ${event.getX(i)} - ${event.getY(i)}")
        }

```

alexander-kulikovskii: good idea! 
refactor `hypot` function, but leave list of motion events. 
it's important to pass all events not only last one.

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -296,9 +300,65 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
    */
   @SuppressLint("ClickableViewAccessibility")
   override fun onTouchEvent(event: MotionEvent): Boolean {
+    interceptedActionList.firstOrNull()?.let {
```
yunikkk, line 303: Shouldn't all events be passed and not just first one?

yunikkk: E.g., instead of the current setup - 
```
if (intereceptedEvents.notEmpty) {
  intereceptedEvents.forEach { 
    mapController.onTouchEvent(it)
  }
  interceptedActionList.clear()
}
```

alexander-kulikovskii: done

sdk/src/main/java/com/mapbox/maps/MapView.kt
```diff
@@ -39,6 +41,8 @@ open class MapView : FrameLayout, MapPluginProviderDelegate, MapControllable {
   internal var mapController: MapController
     private set
 
+  private var interceptedActionList: MutableList<MotionEvent> = mutableListOf()
```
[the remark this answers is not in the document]
alexander-kulikovskii: rename

yunikkk (approved):

yunikkk (approved):

PR #1746, opened 2022-10-13
PR #1744 port to release branch

yunikkk:
This is #1744 port to the release branch.


alexander-kulikovskii (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 Mapbox welcomes participation and contributions from everyone.
 
+# main
```
kiryldz, line 5: could remove it either now or when cutting the release but I'd remove it now 😄

kiryldz (approved):
LGTM with a nit

.circleci/config.yml
```diff
@@ -347,13 +347,13 @@ commands:
             if [[ -n "${GCLOUD_SERVICE_ACCOUNT_JSON}" ]]; then
               gcloud firebase test android models list
               ### We use following devices to run robo tests:
-              ### * 1 old API (the lower the better)
-              ### * 1 new API (the higher the better)
-              ### * 1 x86-64 emulator(Pixel2)
-              ### * API level 23 which is known to work weird sometimes
+              ### * 1 old API 21, Virtual Nexus 4
```
yunikkk, line 350: Replaced hwALE device that is not present anymore on gcloud, failed with 

```
Some device dimensions are not compatible and will be skipped:
  [OS-version 21 on hwALE-H]
Firebase Test Lab will execute your robo test on 3 device(s).
```

yunikkk: https://app.circleci.com/pipelines/github/mapbox/mapbox-maps-android/6781/workflows/50301da3-815e-4f4a-9fa1-c33a39d85529/jobs/56985

pengdev: please keep the original description where it describes why the specific device is chosen to run the tests.

Also why this was not fixed in the original PR but only in the port PR?

yunikkk: The tests still passed there - turns out unsupported device does not fail the test run but is ignored silently. And I've discovered it within this PR while retrying the flaky robo tests.  
But you're right that it's better to introduce the fix separately, will revert it here.

PR #1751, opened 2022-10-14
[viewannotation] Introduce get annotations list api MAPSAND-463

ank27:
### Summary of changes

Introduce view annotation `ViewAnnotationManager.annotations` API to access list of added view annotations.

## Pull request checklist:
    
 - [x] If this PR is a `v10.10` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.10` release branch.

### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -38,6 +38,7 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
+  private val annotationsMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
 
   // using copy on write as user could remove listener while callback is invoked
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
@@ -92,6 +93,7 @@ internal class ViewAnnotationManagerImpl(
     currentlyDrawnViewIdSet.clear()
     annotationMap.clear()
     idLookupMap.clear()
+    annotationsMap.clear()
   }
 
   override fun updateViewAnnotation(
@@ -109,6 +111,7 @@ internal class ViewAnnotationManagerImpl(
         it.measuredHeight = USER_FIXED_DIMENSION
       }
       getValue(mapboxMap.updateViewAnnotation(id, options))
+      annotationsMap[view] = options
       return true
     } ?: return false
   }
@@ -145,6 +148,11 @@ internal class ViewAnnotationManagerImpl(
   override fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode {
     return renderThread.viewAnnotationMode
   }
+
+  override val annotations: HashMap<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      this.putAll(annotationsMap)
+    }
 
   /**
    * We will have two calls of this callback:
@@ -300,6 +308,7 @@ internal class ViewAnnotationManagerImpl(
     inflatedView.addOnAttachStateChangeListener(viewAnnotation.attachStateListener)
     annotationMap[viewAnnotation.id] = viewAnnotation
     idLookupMap[inflatedView] = viewAnnotation.id
+    annotationsMap[inflatedView] = updatedOptions
     getValue(mapboxMap.addViewAnnotation(viewAnnotation.id, updatedOptions))
     return inflatedView
   }
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -39,7 +39,6 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
-  private val annotationOptionsLookupMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
 
   // using copy on write as user could remove listener while callback is invoked
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
@@ -83,7 +82,6 @@ internal class ViewAnnotationManagerImpl(
   override fun removeViewAnnotation(view: View): Boolean {
     val id = idLookupMap.remove(view) ?: return false
     val annotation = annotationMap.remove(id) ?: return false
-    annotationOptionsLookupMap.remove(view) ?: return false
     remove(id, annotation)
     return true
   }
@@ -95,7 +93,6 @@ internal class ViewAnnotationManagerImpl(
     currentlyDrawnViewIdSet.clear()
     annotationMap.clear()
     idLookupMap.clear()
-    annotationOptionsLookupMap.clear()
   }
 
   override fun updateViewAnnotation(
@@ -113,7 +110,6 @@ internal class ViewAnnotationManagerImpl(
         it.measuredHeight = USER_FIXED_DIMENSION
       }
       getValue(mapboxMap.updateViewAnnotation(id, options))
-      annotationOptionsLookupMap[view] = options
       return true
     } ?: return false
   }
@@ -153,7 +149,11 @@ internal class ViewAnnotationManagerImpl(
 
   override val annotations: Map<View, ViewAnnotationOptions>
     get() = HashMap<View, ViewAnnotationOptions>().apply {
-      putAll(annotationOptionsLookupMap)
+      idLookupMap.forEach { lookupMap ->
+        getViewAnnotationOptionsByView(lookupMap.key)?.let { viewAnnotationOptions ->
+          put(lookupMap.key, viewAnnotationOptions)
+        }
+      }
     }
 
   /**
@@ -310,7 +310,6 @@ internal class ViewAnnotationManagerImpl(
     inflatedView.addOnAttachStateChangeListener(viewAnnotation.attachStateListener)
     annotationMap[viewAnnotation.id] = viewAnnotation
     idLookupMap[inflatedView] = viewAnnotation.id
-    annotationOptionsLookupMap[inflatedView] = updatedOptions
     getValue(mapboxMap.addViewAnnotation(viewAnnotation.id, updatedOptions))
     return inflatedView
   }
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -149,7 +149,7 @@ internal class ViewAnnotationManagerImpl(
 
   override val annotations: Map<View, ViewAnnotationOptions>
     get() = HashMap<View, ViewAnnotationOptions>().apply {
-      idLookupMap.forEach { lookupMap ->
+      idLookupMap.forEach { (view, _) ->
         getViewAnnotationOptionsByView(lookupMap.key)?.let { viewAnnotationOptions ->
           put(lookupMap.key, viewAnnotationOptions)
         }
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -13,6 +13,7 @@ import com.mapbox.maps.viewannotation.ViewAnnotation.Companion.USER_FIXED_DIMENS
 import com.mapbox.maps.viewannotation.ViewAnnotationVisibility
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArraySet
+import kotlin.collections.Map
 
 internal class ViewAnnotationManagerImpl(
   mapView: MapView,
@@ -38,7 +39,7 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
-  private val annotationsMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
+  private val annotationOptionsLookupMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
 
   // using copy on write as user could remove listener while callback is invoked
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
@@ -82,6 +83,7 @@ internal class ViewAnnotationManagerImpl(
   override fun removeViewAnnotation(view: View): Boolean {
     val id = idLookupMap.remove(view) ?: return false
     val annotation = annotationMap.remove(id) ?: return false
+    annotationOptionsLookupMap.remove(view) ?: return false
     remove(id, annotation)
     return true
   }
@@ -93,7 +95,7 @@ internal class ViewAnnotationManagerImpl(
     currentlyDrawnViewIdSet.clear()
     annotationMap.clear()
     idLookupMap.clear()
-    annotationsMap.clear()
+    annotationOptionsLookupMap.clear()
   }
 
   override fun updateViewAnnotation(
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -92,6 +93,7 @@ internal class ViewAnnotationManagerImpl(
     currentlyDrawnViewIdSet.clear()
     annotationMap.clear()
     idLookupMap.clear()
+    annotationsMap.clear()
```
kiryldz, line 96: I see clearing whole map on `removeAllViewAnnotations` but can't see removing an item when `removeViewAnnotation` is called

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,6 +326,44 @@ class ViewAnnotationManagerTest {
...
+    val anotherView = mockk<View>()
+    mockView(anotherView)
+    viewAnnotationManager.addViewAnnotation(
+      anotherView,
+      viewAnnotationOptions {
+        geometry(Point.fromLngLat(90.0, 90.0))
+      }
+    )
+    every { anotherView.removeOnAttachStateChangeListener(any()) } just Runs
+    assertNotNull(viewAnnotationManager.annotations)
```
kiryldz, line 345: guess this assert is useless assuming we have check size assert below

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +149,11 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override val annotations: HashMap<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      this.putAll(annotationsMap)
```
kiryldz, line 154: `this` could be ommited

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: good catch 💎

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +150,11 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override val annotations: HashMap<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
```
alexander-kulikovskii, line 154: maybe create hashMap with `initialCapacity`?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -38,6 +38,7 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
+  private val annotationsMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
```
alexander-kulikovskii, line 41: maybe I missed smth but when do you add elements to with map?

kiryldz: it's happening on L312 with `map[key]=value` syntax

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,6 +326,77 @@ class ViewAnnotationManagerTest {
...
+    viewAnnotationManager.removeViewAnnotation(view)
+    assert(viewAnnotationManager.annotations.isEmpty())
+  }
+
+  @Test
+  fun getViewAnnotationsAfterRemoveAllAnnotations() {
+    every { mapboxMap.removeViewAnnotation(any()) } returns ExpectedFactory.createNone()
+    every { view.removeOnAttachStateChangeListener(any()) } just Runs
+
+    viewAnnotationManager.removeAllViewAnnotations()
```
kiryldz, line 375: it seems we did not even add anything beforehand - is this expected?

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,6 +326,77 @@ class ViewAnnotationManagerTest {
...
+    assert(viewAnnotationManager.annotations.isEmpty())
+  }
+
+  @Test
+  fun validateUpdateOptionsFromAnnotationsMap() {
+    every { mapboxMap.updateViewAnnotation(any(), any()) } returns ExpectedFactory.createNone()
+    val updatedOptions = viewAnnotationOptions {
```
kiryldz, line 382:
```suggestion
    val expectedOptions = viewAnnotationOptions {
```

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,6 +326,77 @@ class ViewAnnotationManagerTest {
...
+        geometry(DEFAULT_GEOMETRY)
+      }
+    )
+    // update annotation options
+    val result = viewAnnotationManager.updateViewAnnotation(view, updatedOptions)
+    assertTrue(result)
+
+    // validate annotation options from annotations map
+    val options = viewAnnotationManager.annotations[view]
+    assertEquals(options, updatedOptions)
```
kiryldz, line 397: nit but guess should be: `assertEquals(expected, actual)`

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +150,11 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override val annotations: HashMap<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      putAll(annotationsMap)
+    }
```
alexander-kulikovskii, line 156: maybe add private setter despite that modification will not update actual annotation options?

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,8 +331,88 @@ class ViewAnnotationManagerTest {
...
+
+    // validate annotation options from annotations map
+    val actualOptions = viewAnnotationManager.annotations[view]
+    assertEquals(expectedOptions, actualOptions)
+  }
+
   private companion object {
     const val FEATURE_ID = "featureId"
     val DEFAULT_GEOMETRY: Geometry = Point.fromLngLat(0.0, 0.0)
+    const val DEFAULT_WIDTH = 20
```
alexander-kulikovskii, line 415:
```suggestion
    private const val DEFAULT_WIDTH = 20
```

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
```diff
@@ -326,8 +331,88 @@ class ViewAnnotationManagerTest {
...
+      viewAnnotationOptions {
+        geometry(Point.fromLngLat(90.0, 90.0))
+      }
+    )
+    every { anotherView.removeOnAttachStateChangeListener(any()) } just Runs
+    assertEquals(2, viewAnnotationManager.annotations.size)
+  }
+
+  @Test
+  fun getViewAnnotationOptionsFromAnnotationsMap() {
```
alexander-kulikovskii, line 353: maybe add test for:
- checking adding same annotations? size should be const
- adding several times different annotations and between it check list size

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +155,17 @@ interface ViewAnnotationManager {
...
+  /**
+   * Retrieve added [View] with their [ViewAnnotationOptions]. This will return
+   * fresh copies of [View] and associated [ViewAnnotationOptions] every time.
+   *
+   * Note: Modifying [ViewAnnotationOptions] will not update actual annotation options. Use
+   * [ViewAnnotationManager.updateViewAnnotation] instead. Modifying [View] will update the actual view property.
+   *
+   * @return Map of [View] and associated [ViewAnnotationOptions].
+   */
+  val annotations: HashMap<View, ViewAnnotationOptions>
```
yunikkk, line 167: Any reasons to return specific HashMap and not just plain Map?

ank27: Yeah, Interface sounds good. 👍

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -38,6 +39,7 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
+  private val annotationsMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
```
yunikkk, line 42: There's already `private val annotationMap = ConcurrentHashMap<String, ViewAnnotation>()` at line 37, I'd suggest to use different name here)

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +155,17 @@ interface ViewAnnotationManager {
    */
   fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode
 
+  /**
+   * Retrieve added [View] with their [ViewAnnotationOptions]. This will return
+   * fresh copies of [View] and associated [ViewAnnotationOptions] every time.
```
yunikkk, line 160: IMHO wording is a bit confusing, we won't copy actual `views`, right.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Btw - do we need this separate field at all? Can we just get the views from `idLookupMap` and map them with `ViewAnnotationOptions` using `getViewAnnotationOptionsByView` ?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
ank27: yeah better wording can be `This will return Map of [View] and associated [ViewAnnotationOptions] every time.`, wdyt?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +151,11 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override val annotations: Map<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      putAll(annotationOptionsLookupMap)
+    }
```
jush, line 157: This does the same:
```suggestion
    get() = annotationOptionsLookupMap.toMap()
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -38,6 +39,7 @@ internal class ViewAnnotationManagerImpl(
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
   private val unpositionedViews = mutableSetOf<View>()
+  private val annotationOptionsLookupMap = ConcurrentHashMap<View, ViewAnnotationOptions>()
```
jush, line 42: Is it okay to keep a local copy of `ViewAnnotationOptions` instead of calling the [native method](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt#L1774) to get those?

Could there be some discrepancies between the one stored here and the one in the native side?

We could build that map when `annotations` are requested by using that native method plus the `idLookupMap` that we already have.

ank27: > Is it okay to keep a local copy of ViewAnnotationOptions instead of calling the [native method](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt#L1774) to get those?

we do update these annotationOptions in map whenever `updateViewAnnotation` is called. so there will not be any discrepancies between the ones stored vs native one.

> We could build that map when annotations are requested by using that native method plus the idLookupMap that we already have.

I guess yes, but we don't want to call native method that many times (that's the idea initially ), for eg if there are 100 viewAnnotations added and user request for annotations map, then we would call [native method](https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt#L1774) that many times every time user query annotations. I think [this](https://github.com/mapbox/mapbox-maps-android/pull/1751#discussion_r1006655113) is similar question.

jush: Indeed is the same question, sorry about that.
Thanks for explaining!

ank27: @jush updated implementation to build map from idlookup.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: updated.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +147,15 @@ internal class ViewAnnotationManagerImpl(
...
   }
 
+  override val annotations: Map<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      idLookupMap.forEach { lookupMap ->
+        getViewAnnotationOptionsByView(lookupMap.key)?.let { viewAnnotationOptions ->
+          put(lookupMap.key, viewAnnotationOptions)
+        }
+      }
+    }
```
yunikkk, line 157: Thought that you can easily use `map` kotlin function but apparently it maps to List for some reason. However, another notation is possible : 
```
  val annotations: Map<View, ViewAnnotationOptions>
    get() = idLookupMap.map { (view, _) ->
      view to getViewAnnotationOptionsByView(view)!!
    }.toMap()
```
with additional `toMap` that maps list of pairs to Map

Another q is - how to treat nullable `getViewAnnotationOptionsByView`? Seems it shouldn't happen, so probably worth to log an error or even use `!!`, wdyt?

ank27: > Thought that you can easily use map kotlin function but apparently it maps to List for some reason

Yeah i did try that earlier but didn't want to use `!!` to be safe, so went ahead with currentl impl, also imo we should not add the entry if `getViewAnnotationOptionsByView ` is null (in case).

yunikkk: Should we log an error though for the null case?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -146,6 +147,15 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override val annotations: Map<View, ViewAnnotationOptions>
+    get() = HashMap<View, ViewAnnotationOptions>().apply {
+      idLookupMap.forEach { lookupMap ->
```
yunikkk, line 152:
```suggestion
      idLookupMap.forEach { (view, _) ->
```

yunikkk (commented):
LGTM, with minor q

jush (approved):

PR #1753, opened 2022-10-17
[viewannotation] Introduce show annotations api [MAPSAND-464]

ank27:
### Summary of changes
Introduce `ViewAnnotationManager.cameraForAnnotations` that returns `CameraOptions` for provided annotations. Users can use custom animation to move camera to fit specified anntations.

If users set `ViewAnnotationOptions.visible = false ` explicitly, we will exclude those views from the calculation for `cameraOptions`. Also, only the views for which visibility is set to Visible, will be considered in the calculation.

### User impact (optional)

<video src="https://user-images.githubusercontent.com/11589497/196913569-e0d0c782-89d7-4a3d-bb51-ff4bfa3847c9.mp4" width = 500/>

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
+++ b/sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
@@ -1,12 +1,17 @@
 package com.mapbox.maps
 
+import android.graphics.Rect
 import android.os.Looper
 import android.view.*
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import android.widget.FrameLayout
 import androidx.annotation.*
 import androidx.asynclayoutinflater.view.AsyncLayoutInflater
+import androidx.core.view.isVisible
 import com.mapbox.bindgen.Expected
+import com.mapbox.geojson.Point
+import com.mapbox.maps.extension.style.layers.properties.generated.ProjectionName
+import com.mapbox.maps.extension.style.projection.generated.getProjection
 import com.mapbox.maps.viewannotation.*
 import com.mapbox.maps.viewannotation.ViewAnnotation
 import com.mapbox.maps.viewannotation.ViewAnnotation.Companion.USER_FIXED_DIMENSION
@@ -14,6 +19,7 @@ import com.mapbox.maps.viewannotation.ViewAnnotationVisibility
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArraySet
 import kotlin.collections.Map
+import kotlin.math.abs
 
 internal class ViewAnnotationManagerImpl(
   mapView: MapView,
@@ -35,6 +41,7 @@ internal class ViewAnnotationManagerImpl(
   }
 
   private val annotationMap = ConcurrentHashMap<String, ViewAnnotation>()
+
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal val idLookupMap = ConcurrentHashMap<View, String>()
   private val currentlyDrawnViewIdSet = mutableSetOf<String>()
@@ -155,6 +162,115 @@ internal class ViewAnnotationManagerImpl(
         }
       }
     }
+
+  override fun cameraForAnnotations(
+    annotations: List<View>,
+    edgeInsets: EdgeInsets?,
+    bearing: Double?,
+    pitch: Double?
+  ): CameraOptions? {
+    if (mapboxMap.style?.getProjection()?.name == ProjectionName.GLOBE || annotations.isEmpty()) {
+      return null
+    }
+    val viewAnnotationOptions = annotations.mapNotNull {
+      if (it.isVisible) return@mapNotNull getViewAnnotationOptionsByView(it) else null
+    }.filter { it.visible != false }
+
+    if (viewAnnotationOptions.isEmpty()) return null
+    val coordinates = coordinatesFromAnnotations(viewAnnotationOptions)
+    val paddings = calculateEdgeInsets(viewAnnotationOptions, edgeInsets)
+    return mapboxMap.cameraForCoordinates(
+      coordinates,
+      paddings,
+      bearing,
+      pitch
+    )
+  }
+
+  /**
+   * Function to get coordinates from list of [ViewAnnotationOptions].
+   */
+  private fun coordinatesFromAnnotations(annotationOptions: List<ViewAnnotationOptions>): List<Point> {
+    val coordinatesList = mutableListOf<Point>()
+    annotationOptions.forEach {
+      it.geometry?.let { geometry ->
+        coordinatesList.add(geometry as Point)
+      }
+    }
+    return coordinatesList
+  }
+
+  /**
+   * Calculate paddings to show viewAnnotations.
+   * Get the topMost, leftMost, rightMost, bottomMost annotation options and apply paddings accordingly.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun calculateEdgeInsets(
+    viewAnnotationOptions: List<ViewAnnotationOptions>,
+    edgeInsets: EdgeInsets? = null
+  ): EdgeInsets {
+    val filteredViewAnnotations = viewAnnotationOptions.filter { it.geometry != null }
+    val topAnnotation =
+      filteredViewAnnotations.maxByOrNull { (it.geometry!! as Point).latitude() }
+    val bottomAnnotation =
+      filteredViewAnnotations.minByOrNull { (it.geometry!! as Point).latitude() }
+    val leftAnnotation =
+      filteredViewAnnotations.minByOrNull { (it.geometry!! as Point).longitude() }
+    val rightAnnotation =
+      filteredViewAnnotations.maxByOrNull { (it.geometry!! as Point).longitude() }
+
+    return EdgeInsets(
+      (edgeInsets?.top ?: 0).toDouble()
+        .plus(abs(getViewAnnotationOptionsFrame(topAnnotation)?.top ?: 0)),
+      (edgeInsets?.left ?: 0).toDouble()
+        .plus(abs(getViewAnnotationOptionsFrame(leftAnnotation)?.left ?: 0)),
+      (edgeInsets?.bottom ?: 0).toDouble()
+        .plus(getViewAnnotationOptionsFrame(bottomAnnotation)?.bottom ?: 0),
+      (edgeInsets?.right ?: 0).toDouble()
+        .plus(getViewAnnotationOptionsFrame(rightAnnotation)?.right ?: 0)
+    )
+  }
+
+  /**
+   * Get [Rect] from [ViewAnnotationOptions]'s geometry, width and height.
+   * This function takes [ViewAnnotationOptions.geometry] as the center of rectangle and
+   * use width, height and offset values to calculate [Rect] associated.
+   *
+   * @return [Rect] associated with [ViewAnnotationOptions]
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun getViewAnnotationOptionsFrame(viewAnnotationOptions: ViewAnnotationOptions?): Rect? {
+    viewAnnotationOptions?.let { options ->
+      if (options.width != null && options.height != null) {
+        val offsetWidth = if (options.width!! > 0) (options.width!! * 0.5).toInt() else 0
+        val offsetHeight = if (options.height!! > 0) (options.height!! * 0.5).toInt() else 0
+        // create a dummy rect with center assume at 0,0 with offsetWidth and offsetHeight.
+        val rect = Rect(
+          -offsetWidth,
+          -offsetHeight,
+          offsetWidth,
+          offsetHeight
+        )
+
+        // offset rect with respect to anchor defined in viewannotation options.
+        when (options.anchor ?: ViewAnnotationAnchor.CENTER) {
+          ViewAnnotationAnchor.TOP -> rect.offset(0, offsetHeight)
+          ViewAnnotationAnchor.TOP_LEFT -> rect.offset(offsetWidth, offsetHeight)
+          ViewAnnotationAnchor.TOP_RIGHT -> rect.offset(-offsetWidth, offsetHeight)
+          ViewAnnotationAnchor.BOTTOM -> rect.offset(0, -offsetHeight)
+          ViewAnnotationAnchor.BOTTOM_LEFT -> rect.offset(offsetWidth, -offsetHeight)
+          ViewAnnotationAnchor.BOTTOM_RIGHT -> rect.offset(-offsetWidth, -offsetHeight)
+          ViewAnnotationAnchor.LEFT -> rect.offset(offsetWidth, 0)
+          ViewAnnotationAnchor.RIGHT -> rect.offset(-offsetWidth, 0)
+          else -> rect.offset(0, 0)
+        }
+        // add view annotation option's offsetX and offsetY field to offset the rect.
+        rect.offset(options.offsetX ?: 0, options.offsetY ?: 0)
+        return rect
+      }
+    }
+    return null
+  }
 
   /**
    * We will have two calls of this callback:
--- a/sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
+++ b/sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
@@ -6,6 +6,8 @@ import androidx.asynclayoutinflater.view.AsyncLayoutInflater
 import com.mapbox.geojson.Feature
 import com.mapbox.geojson.Geometry
 import com.mapbox.geojson.Point
+import com.mapbox.maps.CameraOptions
+import com.mapbox.maps.EdgeInsets
 import com.mapbox.maps.MapView
 import com.mapbox.maps.ViewAnnotationOptions
 
@@ -165,6 +167,32 @@ interface ViewAnnotationManager {
    * @return Map of [View] and associated [ViewAnnotationOptions].
    */
   val annotations: Map<View, ViewAnnotationOptions>
+
+  /**
+   * Return camera options bound to given view annotation list, padding, bearing and pitch values.
+   * Annotations with [ViewAnnotationOptions.visible] set to false will be excluded from the calculations of [CameraOptions].
+   * Annotations with only [View.VISIBLE] will be included in the calculations for [CameraOptions]
+   *
+   * Note: This API isn't supported by Globe projection and will return NULL.
+   * Calling this API immediately after adding the view is a no-op.
+   * Please refer to [OnViewAnnotationUpdatedListener] documentation for understanding the exact moment of time when
+   * view annotation is positioned.
+   *
+   * @param annotations view annotation list to be shown. Annotations should be added beforehand
+   * with [ViewAnnotationManager.addViewAnnotation] API.
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   *
+   * @return [CameraOptions] object or NULL if [annotations] list is empty.
+   *
+   */
+  fun cameraForAnnotations(
+    annotations: List<View>,
+    edgeInsets: EdgeInsets? = null,
+    bearing: Double? = null,
+    pitch: Double? = null
+  ): CameraOptions?
 
   /**
    * Static methods and variables.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone.
 # main
 ## Features ✨ and improvements 🏁
 * Introduce view annotation `ViewAnnotationManager.annotations` API to access list of added view annotations. ([1751](https://github.com/mapbox/mapbox-maps-android/pull/1751))
+* Introduce view annotation `ViewAnnotationManager.cameraForAnnotations` API to get camera options for given view annotations list. ([1753](https://github.com/mapbox/mapbox-maps-android/pull/1753))
 
 ## Bug fixes 🐞
 Fix an issue when touch events didn't pass through clickable view annotations. ([1745](https://github.com/mapbox/mapbox-maps-android/pull/1745))
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationBasicAddActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/markersandcallouts/ViewAnnotationBasicAddActivity.kt
@@ -4,12 +4,14 @@ import android.annotation.SuppressLint
 import android.os.Bundle
 import android.view.Menu
 import android.view.MenuItem
+import android.view.View
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
 import com.mapbox.geojson.Point
 import com.mapbox.maps.*
+import com.mapbox.maps.plugin.animation.flyTo
 import com.mapbox.maps.plugin.gestures.*
 import com.mapbox.maps.testapp.R
 import com.mapbox.maps.testapp.databinding.ActivityViewAnnotationShowcaseBinding
@@ -25,6 +27,7 @@ class ViewAnnotationBasicAddActivity : AppCompatActivity(), OnMapClickListener {
 
   private lateinit var mapboxMap: MapboxMap
   private lateinit var viewAnnotationManager: ViewAnnotationManager
+  private val viewAnnotationViews = mutableListOf<View>()
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
@@ -43,6 +46,17 @@ class ViewAnnotationBasicAddActivity : AppCompatActivity(), OnMapClickListener {
           }
         }
         Toast.makeText(this@ViewAnnotationBasicAddActivity, STARTUP_TEXT, Toast.LENGTH_LONG).show()
+      }
+    }
+
+    binding.fabReframe.setOnClickListener {
+      if (viewAnnotationViews.isNotEmpty()) {
+        val cameraOptions = viewAnnotationManager.cameraForAnnotations(viewAnnotationViews)
+        cameraOptions?.let {
+          mapboxMap.flyTo(it)
+        }
+      } else {
+        Toast.makeText(this@ViewAnnotationBasicAddActivity, ADD_VIEW_ANNOTATION_TEXT, Toast.LENGTH_LONG).show()
       }
     }
   }
@@ -80,10 +94,12 @@ class ViewAnnotationBasicAddActivity : AppCompatActivity(), OnMapClickListener {
         allowOverlap(true)
       }
     )
+    viewAnnotationViews.add(viewAnnotation)
     ItemCalloutViewBinding.bind(viewAnnotation).apply {
       textNativeView.text = "lat=%.2f\nlon=%.2f".format(point.latitude(), point.longitude())
       closeNativeView.setOnClickListener {
         viewAnnotationManager.removeViewAnnotation(viewAnnotation)
+        viewAnnotationViews.remove(viewAnnotation)
       }
       selectButton.setOnClickListener { b ->
         val button = b as Button
@@ -109,5 +125,6 @@ class ViewAnnotationBasicAddActivity : AppCompatActivity(), OnMapClickListener {
   private companion object {
     const val SELECTED_ADD_COEF_PX = 25
     const val STARTUP_TEXT = "Click on a map to add a view annotation."
+    const val ADD_VIEW_ANNOTATION_TEXT = "Add view annotations to re-frame map camera"
   }
 }
--- a/app/src/main/res/layout/activity_view_annotation_showcase.xml
+++ b/app/src/main/res/layout/activity_view_annotation_showcase.xml
@@ -20,4 +20,14 @@
         app:layout_constraintEnd_toEndOf="parent"
         tools:backgroundTint="@color/blue"/>
 
+    <com.google.android.material.floatingactionbutton.FloatingActionButton
+        android:id="@+id/fab_reframe"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="16dp"
+        android:src="@drawable/ic_baseline_refresh_24"
+        app:layout_constraintBottom_toTopOf="@+id/fab_style_toggle"
+        app:layout_constraintEnd_toEndOf="parent"
+        tools:backgroundTint="@color/blue"/>
+
 </androidx.constraintlayout.widget.ConstraintLayout>
--- a/sdk/api/metalava.txt
+++ b/sdk/api/metalava.txt
@@ -584,6 +584,7 @@ package com.mapbox.maps.viewannotation {
     method public android.view.View addViewAnnotation(@LayoutRes int resId, com.mapbox.maps.ViewAnnotationOptions options);
     method public void addViewAnnotation(@LayoutRes int resId, com.mapbox.maps.ViewAnnotationOptions options, androidx.asynclayoutinflater.view.AsyncLayoutInflater asyncInflater, kotlin.jvm.functions.Function1<? super android.view.View,kotlin.Unit> asyncInflateCallback);
     method public void addViewAnnotation(android.view.View view, com.mapbox.maps.ViewAnnotationOptions options);
+    method public com.mapbox.maps.CameraOptions? cameraForAnnotations(java.util.List<? extends android.view.View> annotations, com.mapbox.maps.EdgeInsets? edgeInsets = null, Double? bearing = null, Double? pitch = null);
     method public java.util.Map<android.view.View,com.mapbox.maps.ViewAnnotationOptions> getAnnotations();
     method public android.view.View? getViewAnnotationByFeatureId(String featureId);
     method public com.mapbox.maps.ViewAnnotationOptions? getViewAnnotationOptionsByFeatureId(String featureId);
--- a/sdk/api/sdk.api
+++ b/sdk/api/sdk.api
@@ -587,6 +587,7 @@ public abstract interface class com/mapbox/maps/viewannotation/ViewAnnotationMan
 	public abstract fun addViewAnnotation (ILcom/mapbox/maps/ViewAnnotationOptions;)Landroid/view/View;
 	public abstract fun addViewAnnotation (ILcom/mapbox/maps/ViewAnnotationOptions;Landroidx/asynclayoutinflater/view/AsyncLayoutInflater;Lkotlin/jvm/functions/Function1;)V
 	public abstract fun addViewAnnotation (Landroid/view/View;Lcom/mapbox/maps/ViewAnnotationOptions;)V
+	public abstract fun cameraForAnnotations (Ljava/util/List;Lcom/mapbox/maps/EdgeInsets;Ljava/lang/Double;Ljava/lang/Double;)Lcom/mapbox/maps/CameraOptions;
 	public abstract fun getAnnotations ()Ljava/util/Map;
 	public abstract fun getViewAnnotationByFeatureId (Ljava/lang/String;)Landroid/view/View;
 	public abstract fun getViewAnnotationOptionsByFeatureId (Ljava/lang/String;)Lcom/mapbox/maps/ViewAnnotationOptions;
@@ -600,6 +601,10 @@ public abstract interface class com/mapbox/maps/viewannotation/ViewAnnotationMan
 }
 
 public final class com/mapbox/maps/viewannotation/ViewAnnotationManager$Companion {
+}
+
+public final class com/mapbox/maps/viewannotation/ViewAnnotationManager$DefaultImpls {
+	public static synthetic fun cameraForAnnotations$default (Lcom/mapbox/maps/viewannotation/ViewAnnotationManager;Ljava/util/List;Lcom/mapbox/maps/EdgeInsets;Ljava/lang/Double;Ljava/lang/Double;ILjava/lang/Object;)Lcom/mapbox/maps/CameraOptions;
 }
 
 public final class com/mapbox/maps/viewannotation/ViewAnnotationOptionsKtxKt {
--- a/sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
+++ b/sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerTest.kt
@@ -1,16 +1,16 @@
 package com.mapbox.maps
 
+import android.graphics.Rect
 import android.view.View
 import android.widget.FrameLayout
 import com.mapbox.bindgen.ExpectedFactory
+import com.mapbox.bindgen.Value
 import com.mapbox.geojson.Geometry
 import com.mapbox.geojson.Point
 import com.mapbox.maps.ViewAnnotationManagerImpl.Companion.EXCEPTION_TEXT_ASSOCIATED_FEATURE_ID_ALREADY_EXISTS
+import com.mapbox.maps.extension.style.layers.properties.generated.ProjectionName
 import com.mapbox.maps.renderer.MapboxRenderThread
-import com.mapbox.maps.viewannotation.OnViewAnnotationUpdatedListener
-import com.mapbox.maps.viewannotation.ViewAnnotationManager
-import com.mapbox.maps.viewannotation.ViewAnnotationUpdateMode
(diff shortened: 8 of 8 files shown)


sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,17 @@ interface ViewAnnotationManager {
    */
   fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode
 
+  /**
+   * Sets the visible region of the map that displays all the given views.
+   * This API is not supported by Globe projection.
```
kiryldz, line 162: looking at the impl now I don't see that use-case handled; can we actually check if globe projection is enabled and throw some our Mapbox exception then?

ank27: I took the docs from `cameraForCoordinateBounds` which doesn't support Globe projection. 

> can we actually check if globe projection is enabled and throw some our Mapbox exception then?

Yeah good idea,  i will add exception when using globe.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +158,48 @@ interface ViewAnnotationManager {
...
+   *
+   * @param annotations list of all the views that needs to fit in the visible frame.
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   * @param animationDurationMs animation duration in millis, default to [DEFAULT_ANNOTATION_ANIMATION_DURATION]
+   *
+   * This API is not supported by Globe projection.
+   */
+  fun showAnnotations(
```
yunikkk, line 175: What does this method actually change? Only visible region / camera? IF so - to me the name seems a bit misleading, I'd expect it actually adds shows (adds) the annotations

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,157 @@ internal class ViewAnnotationManagerImpl(
...
+  /**
+   * Calculate paddings to show viewAnnotations.
+   * Get the topMost, leftMost, rightMost, bottomMost annotation options and apply paddings accordingly.
+   */
+  private fun calculateEdgeInset(
+    viewAnnotationOptions: List<ViewAnnotationOptions?>,
+    edgeInsets: EdgeInsets?
+  ): EdgeInsets {
+    val topAnnotation =
+      viewAnnotationOptions.maxByOrNull { Point.fromJson(it?.geometry?.toJson()!!).latitude() }
```
yunikkk, line 244: Could `!!` throw here?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,157 @@ internal class ViewAnnotationManagerImpl(
...
+      }
+    }
+    return CoordinateBounds(Point.fromLngLat(west, south), Point.fromLngLat(east, north), false)
+  }
+
+  /**
+   * Calculate paddings to show viewAnnotations.
+   * Get the topMost, leftMost, rightMost, bottomMost annotation options and apply paddings accordingly.
+   */
+  private fun calculateEdgeInset(
```
yunikkk, line 239: Would be nice to drop some tests for the calculation methods

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
ank27: it basically update camera so that all annotations provided will fit in the visible region. 

> to me the name seems a bit misleading, I'd expect it actually adds shows (adds) the annotations

The name is kept similar to [ios v6](https://docs.mapbox.com/ios/maps/api/6.4.1/Classes/MGLMapView.html#/c:objc(cs)MGLMapView(im)showAnnotations:animated:). I guess we can improve docs a bit to be clear about the api usage.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Yes, I will add tests.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,157 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override fun showAnnotations(
```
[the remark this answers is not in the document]
ank27: We need this method. `cameraForAnnotations` will provide `cameraOptions` which users can use with their custom animations. `showAnnotations` is more like a convenience method which does animations itself using viewport (for now).

> Isn't it possible to just
> 
> map.flyTo(viewAnnotationsManager.cameraForAnnotations)

That's correct. It's much simpler and easy 😃. but I felt viewport animations are much better than simple `flyTo`. 
we can discuss if we really need to use viewport here or simple `flyTo` works. ?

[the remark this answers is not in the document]
yunikkk: What confuses me is that we are kind of mixing the functions of different plugins, allowing ViewAnnotationManager not only to control view annotations but also the Camera.
At the same time if they work differently there indeed might be a reason to use Viewport and not direct camera animations. Thinking about the way to better expose the apis here...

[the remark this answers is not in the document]
yunikkk: Maybe leave method for calculation of camera parameters in ViewAnnotationPlugin, and then add an extension method for animating the viewport to the ViewportPlugin? Or to map CameraOptions to OverviewViewportStateOptions?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
kiryldz: we're using default args in those methods - do we need to mark them as `@JvmOverloads`?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,159 @@ internal class ViewAnnotationManagerImpl(
...
+      val coordinateBounds = coordinateBoundsFromAnnotations(viewAnnotationOptions)
+      val paddings = calculateEdgeInset(viewAnnotationOptions, edgeInsets)
+      val geometry = GeometryCollection.fromGeometries(
+        listOf(
+          coordinateBounds.northeast,
+          coordinateBounds.southwest
+        )
+      )
+      // create overview viewport state with geometry, padding and animation duration
+      val overviewViewportState = mapView.viewport.makeOverviewViewportState(
```
kiryldz, line 173: what of viewport is not on the classpath?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,159 @@ internal class ViewAnnotationManagerImpl(
...
+    edgeInsets: EdgeInsets?,
+    bearing: Double?,
+    pitch: Double?,
+    animationDurationMs: Long
+  ) {
+    if (annotations.isNotEmpty()) {
+      val viewAnnotationOptions = annotations.mapNotNull {
+        getViewAnnotationOptionsByView(it)
+      }
+      val coordinateBounds = coordinateBoundsFromAnnotations(viewAnnotationOptions)
```
kiryldz, line 164: what if annotation is `View.INVISIBLE` or `View.GONE`? not sure it should participate in calculations then

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -439,5 +608,8 @@ internal class ViewAnnotationManagerImpl(
     internal const val EXCEPTION_TEXT_GEOMETRY_IS_NULL = "Geometry can not be null!"
     internal const val EXCEPTION_TEXT_ASSOCIATED_FEATURE_ID_ALREADY_EXISTS =
       "View annotation with associatedFeatureId=%s already exists!"
+    internal const val MAX_LATITUDE = 90.0
+    internal const val MAX_LONGITUDE = 180.0
```
kiryldz, line 612: perhaps those constants are shared already somewhere in the project...

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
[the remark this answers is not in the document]
ank27: Yeah i guess safe option would be to use `flyTo`, since it could be possible that `viewport` might not be included in classpath (as its a different plugin) as mentioned [here](https://github.com/mapbox/mapbox-maps-android/pull/1753#discussion_r1000749223), we could although use conditional check for that.

> What confuses me is that we are kind of mixing the functions of different plugins, allowing ViewAnnotationManager not only to control view annotations but also the Camera.

even if we use `map.flyTo` we will be controling camera in viewAnnotationManager, isn't it? I feel using `viewport.transitionTo` is same as `map.flyTo`. True that viewport is a separate plugin and it might not be available in classpath.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Yeah either we can check if its available or simply use `map.flyTo` as discussed [here](https://github.com/mapbox/mapbox-maps-android/pull/1753#discussion_r1001981929)

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: I am testing that. will add condition for that too

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
kiryldz: @ank27 @maios - after some offline discussion with the team I realized that `showAnnotations` indeed looks redundant. It may have been useful in v6 / v9 where we did not have flexible camera animation system but now it looks strange that we introduce new method `showAnnotations` (naming is confusing as users could think that this utility method makes given view annotations View.VISIBLE from View.INVISIBLE) that allows to create the camera `flyTo` with one parameter - duration. What if the user want to change let's say some other `MapAnimationOptions` like `owner`, `startDelay` etc? 

My proposal is to remove this method at all and leave the only `cameraForAnnotations` returning `CameraOptions` and point user to the animation plugin documentation allowing to customize that animation in any way.

yunikkk: > it basically update camera so that all annotations provided will fit in the visible region.

right, and in Android world it's common to have `.show***` methods that actually change the views visibility - that's why it seems mesleading to me. 

I also think that methods to update the camera would be better placed in the Viewport plugin or MapboxMap itself.

ank27: > My proposal is to remove this method at all and leave the only cameraForAnnotations returning CameraOptions and point user to the animation plugin documentation allowing to customize that animation in any way.

Correct, I am too in favor of just keeping `cameraForAnnotations`, and letting users create their own animations. they can always use either `mapboxMap.flyTo` or `viewport.transitionTo` , which allows them to control and customize the animation options.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +149,117 @@ internal class ViewAnnotationManagerImpl(
...
+    var south = MAX_LATITUDE
+    var west = MAX_LONGITUDE
+    var east = -MAX_LONGITUDE
+    annotationOptions.map {
+      it.geometry?.let { geometry ->
+        val point = Point.fromJson(geometry.toJson())
+        north = maxOf(north, point.latitude())
+        south = minOf(south, point.latitude())
+        west = minOf(west, point.longitude())
+        east = maxOf(east, point.longitude())
```
jush, line 230: Does this algorithm work if I have two annotations, one with longitude = -170 and the other one with longitude 170?

I guess it would return a very large longitude span (340º) when instead I would expect it to be 20º.

What do we usually do for this cases?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
pengdev: agree with above assessment, in the view annotation, I think we just need to expose the cameraForAnnotations. We can build an example of a custom ViewportState to work with viewport plugin, or build a ViewportState separately inside the viewport plugin.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Yeah, correct. it does return large longitude. Normally, in case of camera movements we calculate the shortest path between points. Need to check if we do compensate when crossing antimeridian for camera bounds.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +149,117 @@ internal class ViewAnnotationManagerImpl(
...
 
+  override fun showAnnotations(
+    annotations: List<View>,
+    edgeInsets: EdgeInsets?,
+    bearing: Double?,
+    pitch: Double?,
+    animationDurationMs: Long
+  ) {
+    if (annotations.isNotEmpty()) {
+      val viewAnnotationOptions = annotations.mapNotNull {
```
alexander-kulikovskii, line 160: should it just `map`? as I can see `annotations: List<View>` should contain only not null views.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +149,117 @@ internal class ViewAnnotationManagerImpl(
...
+      val coordinateBounds = coordinateBoundsFromAnnotations(viewAnnotationOptions)
+      val paddings = calculateEdgeInset(viewAnnotationOptions, edgeInsets)
+      return mapboxMap.cameraForCoordinateBounds(
+        coordinateBounds,
+        paddings,
+        bearing,
+        pitch
+      )
+    }
+    return null
```
alexander-kulikovskii, line 212:
```suggestion
    return if (annotations.isNotEmpty()) {
      val viewAnnotationOptions = annotations.mapNotNull {
        getViewAnnotationOptionsByView(it)
      }
      val coordinateBounds = coordinateBoundsFromAnnotations(viewAnnotationOptions)
      val paddings = calculateEdgeInset(viewAnnotationOptions, edgeInsets)
     mapboxMap.cameraForCoordinateBounds(
        coordinateBounds,
        paddings,
        bearing,
        pitch
      )
    } else {
       null
    }
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +149,117 @@ internal class ViewAnnotationManagerImpl(
...
+   * Convenience function to get bounds occupied by list of [ViewAnnotationOptions]'s geometry.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun coordinateBoundsFromAnnotations(annotationOptions: List<ViewAnnotationOptions>): CoordinateBounds {
+    var north = -MAX_LATITUDE
+    var south = MAX_LATITUDE
+    var west = MAX_LONGITUDE
+    var east = -MAX_LONGITUDE
+    annotationOptions.map {
+      it.geometry?.let { geometry ->
```
alexander-kulikovskii, line 225:
```suggestion
    annotationOptions.forEach { annotationOption ->
      annotationOption.geometry?.let { geometry ->
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Looked a bit into that and i think instead of calculating camerabounds here, we could just use `map.cameraForCoordinates` and let gl-native handle bounds (although that method also gives longitudnal span of 340 for this case, and doesn't consider through antimeridian, but will confirm this with native too). I will update pr accordingly :)

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
```diff
@@ -0,0 +1,46 @@
+@file:JvmName("ViewAnnotationOptionsExt")
```
kiryldz, line 1: we don't want to expose it, this should be removed

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,26 @@ interface ViewAnnotationManager {
...
+   * Get the [CameraOptions] object bound to specified annotations on the MapView.
+   *
+   * Using this method with GlobeProjection might lead to incorrect results.
+   *
+   * @param annotations list of views needs to fit in the visible frame.
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   *
+   * @return [CameraOptions] object, Will be null if empty annotations provided.
```
kiryldz, line 170:
```suggestion
   * @return [CameraOptions] object or NULL if [annotations] list is empty.
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,26 @@ interface ViewAnnotationManager {
    */
   fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode
 
+  /**
+   * Get the [CameraOptions] object bound to specified annotations on the MapView.
+   *
+   * Using this method with GlobeProjection might lead to incorrect results.
```
kiryldz, line 163: how incorrect are they? I thought we will be introducing some sort of exception or we could return NULL if projection is globe

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,26 @@ interface ViewAnnotationManager {
    */
   fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode
 
+  /**
+   * Get the [CameraOptions] object bound to specified annotations on the MapView.
```
kiryldz, line 161:
```suggestion
   * Return camera options showing given view annotation list, padding, bearing and pitch values.
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,26 @@ interface ViewAnnotationManager {
...
+   *
+   * Using this method with GlobeProjection might lead to incorrect results.
+   *
+   * @param annotations list of views needs to fit in the visible frame.
```
kiryldz, line 165:
```suggestion
   * @param view annotation list to be shown.
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,26 @@ interface ViewAnnotationManager {
...
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   *
+   * @return [CameraOptions] object, Will be null if empty annotations provided.
+   *
+   */
+  fun cameraForAnnotations(
```
kiryldz, line 173: State behaviour in documentation about framing invisible / gone views? How will it behave in general, did you test it?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Get [Rect] from [ViewAnnotationOptions]'s geometry, width and height.
+ * This function takes [ViewAnnotationOptions.geometry] as the center of rectangle and
+ * use width, height and offset values to calculate [Rect] associated.
+ *
+ * @return [Rect] associated with [ViewAnnotationOptions]
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal fun ViewAnnotationOptions.frame(): Rect? {
+  if (width != null && height != null) {
```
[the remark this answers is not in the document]
kiryldz: @ank27 what about use-cases with WRAP_CONTENT and width being negative for a while - should we filter it as well? also curious if it all works as expected for the use-case described in https://github.com/mapbox/mapbox-maps-android/pull/1753/files#r1006568024

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
ank27: I have tested it and `cameraForCoordinates` doesn't provide any exceptions, so no need to add exceptions here.

> how incorrect are they? 

For eg. if we select points that are far apart, when it calculates cameraOption center, it does fall in the place where both annotations are not visible.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +146,69 @@ internal class ViewAnnotationManagerImpl(
...
+    }
+    return coordinatesList
+  }
+
+  /**
+   * Calculate paddings to show viewAnnotations.
+   * Get the topMost, leftMost, rightMost, bottomMost annotation options and apply paddings accordingly.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun calculateEdgeInset(
```
yunikkk, line 190:
```suggestion
  internal fun calculateEdgeInsets(
```

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
kiryldz: imo we should either support it fully - then remove the warning docs or do not support at all. Looking at `MapboxMap#cameraForCoordinateBounds` docs: 
```
This API isn't supported by Globe projection and will return a no-op result matching the world bounds.
```
I lean to having similar behaviour here

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +146,69 @@ internal class ViewAnnotationManagerImpl(
     return renderThread.viewAnnotationMode
   }
 
+  override fun cameraForAnnotations(
```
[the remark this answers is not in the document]
ank27: It makes sense to have it in mapboxMap, but i think mapboxMap class will be crowded with all the calculations that we do to calculate bounds and edgeInsets. although, we can use extension functions for that. 

cc @maios how does this sound from ios side?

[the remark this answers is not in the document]
maios: imo it is better to house this method in `ViewAnnotationManager`, I think `ViewAnnotationManager`is created to have a separated set of APIs relating to view annotations, and to offload some interfaces from MapboxMap so that it does not become too crowded, what do you think @mapbox/maps-ios ?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: `mapNotNull` works on transformed result and if its not null, then adds to the list. i.e. if `getViewAnnotationOptionsByView ` produces null result, then it will not be added to resulting list.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
ank27: It's needed in test.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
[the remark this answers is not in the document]
ank27: i haven't checked usecase for that. will test it and update filtering accordingly.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +148,68 @@ internal class ViewAnnotationManagerImpl(
...
+    annotations: List<View>,
```
jush, line 152: With this approach we're forcing the consumers of this API to keep track of the view annotations themselves. They need to keep a [list](https://github.com/mapbox/mapbox-maps-android/pull/1753/files#diff-84626270966cd9b838ebc2ab46c6e4fb92f9354c0f9e9f821fa7813d18f2358aR30), [add](https://github.com/mapbox/mapbox-maps-android/pull/1753/files#diff-84626270966cd9b838ebc2ab46c6e4fb92f9354c0f9e9f821fa7813d18f2358aR91) and [remove](https://github.com/mapbox/mapbox-maps-android/pull/1753/files#diff-84626270966cd9b838ebc2ab46c6e4fb92f9354c0f9e9f821fa7813d18f2358aR96).

Is there any specific reason we can't use the list of views `ViewAnnotationsManagerImpl` keeps internally?

jush: Sorry I misunderstood it. Disregard this comment.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
kiryldz: do we need it in the Java test? still don't get it...

kiryldz: `ViewAnnotationOptions.frame()` below is internal and should be accessible without specifying `@file:JvmName("ViewAnnotationOptionsExt")` I assume. Specifying it that way you do now actually add an empty entry in `sdk/api/metalava.txt` file above which we do not need

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
```diff
@@ -0,0 +1,53 @@
...
+      -offset.first.toInt(),
+      -offset.second.toInt(),
+      offset.first.toInt(),
+      offset.second.toInt()
+    )
+
+    // offset rect with respect to anchor defined in viewannotation options.
+    when (anchor ?: ViewAnnotationAnchor.CENTER) {
+      ViewAnnotationAnchor.TOP ->
+        rect.offset(0, offset.second.toInt())
```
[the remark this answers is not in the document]
ank27: lets keep it same for now.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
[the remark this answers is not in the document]
ank27: updated.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -143,6 +150,68 @@ internal class ViewAnnotationManagerImpl(
...
+    edgeInsets: EdgeInsets? = null
+  ): EdgeInsets {
+    val filteredViewAnnotations = viewAnnotationOptions.filter { it.geometry != null }
+    val topAnnotation =
+      filteredViewAnnotations.maxByOrNull { Point.fromJson(it.geometry!!.toJson()).latitude() }
+    val bottomAnnotation =
+      filteredViewAnnotations.minByOrNull { Point.fromJson(it.geometry!!.toJson()).latitude() }
+    val leftAnnotation =
+      filteredViewAnnotations.minByOrNull { Point.fromJson(it.geometry!!.toJson()).longitude() }
+    val rightAnnotation =
```
yunikkk, line 204: Do you need to compare lat/lon, or simply offsetX/offsetY should be enough?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
```diff
@@ -0,0 +1,46 @@
...
+ * This function takes [ViewAnnotationOptions.geometry] as the center of rectangle and
+ * use width, height and offset values to calculate [Rect] associated.
+ *
+ * @return [Rect] associated with [ViewAnnotationOptions]
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal fun ViewAnnotationOptions.frame(): Rect? {
+  if (width != null && height != null) {
+    val offsetWidth = if (width!! > 0) (width!! * 0.5).toInt() else 0
+    val offsetHeight = if (height!! > 0) (height!! * 0.5).toInt() else 0
```
yunikkk, line 20: So for views that are not yet measured (width / height = -1) we'll have Rect(0,0,0,0). Is that intentional? Maybe return `null` for such cases?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Ah nevermind, I've confused it with the `ViewAnnotationPositionDescriptor` coordinate, it's offset relative to the geometry point...

yunikkk: But at the same time it seems could calculate all annotations in one pass. Now the whole list of annotations is mapped to/from json 4 times.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
ank27: These are for the calculations of paddings. if width/height is -1 we will use 0 paddings for those. I have added docs for this.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Yeah, i have that in mind, will add a ticket to improve the code after beta.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationOptionsExt.kt
yunikkk: Right, so it'll return Rect with width and height == 0, right?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: Btw since you anyway expect Point - is Json needed at all? Cant you just cast geometry to point and take lat/lon ? Eg
`(options.geometry as Point).latitude()`?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -155,6 +157,30 @@ interface ViewAnnotationManager {
    */
   fun getViewAnnotationUpdateMode(): ViewAnnotationUpdateMode
 
+  /**
+   * Return camera options showing given view annotation list, padding, bearing and pitch values.
+   * Annotations with [ViewAnnotationOptions.visible] set to false will be excluded from the calculations of [CameraOptions].
+   * Annotations with only [View.VISIBLE] will be included in the calculations for [CameraOptions]
+   *
+   * Note: This API isn't supported by Globe projection and will return NULL.
+   * Calling this API immediately after adding the view is a no-op.
```
kiryldz, line 166:
```suggestion
   * Note: Calling this API immediately after adding the view is a no-op. Please refer to [OnViewAnnotationUpdatedListener] documentation for understanding the exact moment of time when view annotation is positioned.
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: good suggestion. will update that :)

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -156,6 +163,115 @@ internal class ViewAnnotationManagerImpl(
...
+  }
+
+  /**
+   * Function to get coordinates from list of [ViewAnnotationOptions].
+   */
+  private fun coordinatesFromAnnotations(annotationOptions: List<ViewAnnotationOptions>): List<Point> {
+    val coordinatesList = mutableListOf<Point>()
+    annotationOptions.forEach {
+      it.geometry?.let { geometry ->
+        coordinatesList.add(Point.fromJson(geometry.toJson()))
```
yunikkk, line 197: Btw here you can also get rid of json and just `it.geometry!! as Point`

yunikkk (approved):
LGTM, one nit

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -166,6 +168,32 @@ interface ViewAnnotationManager {
    */
   val annotations: Map<View, ViewAnnotationOptions>
 
+  /**
+   * Return camera options showing given view annotation list, padding, bearing and pitch values.
```
yunikkk, line 172: I'd reworded a bit, sounds a bit confusing

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -156,6 +163,115 @@ internal class ViewAnnotationManagerImpl(
...
+    val bottomAnnotation =
+      filteredViewAnnotations.minByOrNull { (it.geometry!! as Point).latitude() }
+    val leftAnnotation =
+      filteredViewAnnotations.minByOrNull { (it.geometry!! as Point).longitude() }
+    val rightAnnotation =
+      filteredViewAnnotations.maxByOrNull { (it.geometry!! as Point).longitude() }
+
+    return EdgeInsets(
+      (edgeInsets?.top ?: 0).toDouble()
+        .plus(abs(getViewAnnotationOptionsFrame(topAnnotation)?.top ?: 0)),
```
yunikkk, line 224: What will happen if topmost annotation is small (so getViewAnnotationOptionsFrame will also return some small value), but another one, that is below - is much bigger. Will edge insets calculated correctly then? Won't this bigger annotation be hidden?

yunikkk (commented):
Couple questions

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -166,6 +168,32 @@ interface ViewAnnotationManager {
...
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   *
+   * @return [CameraOptions] object or NULL if [annotations] list is empty.
+   *
+   */
+  fun cameraForAnnotations(
+    annotations: List<View>,
+    edgeInsets: EdgeInsets? = null,
```
alexander-kulikovskii, line 192: try to set
`EdgeInsets(-1000000.0, ...` or `100000` and it's break sample. not crash, just blank screen

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -166,6 +168,32 @@ interface ViewAnnotationManager {
...
+   * @param annotations view annotation list to be shown. Annotations should be added beforehand
+   * with [ViewAnnotationManager.addViewAnnotation] API.
+   * @param edgeInsets paddings to apply.
+   * @param bearing camera bearing to apply.
+   * @param pitch camera pitch to apply.
+   *
+   * @return [CameraOptions] object or NULL if [annotations] list is empty.
+   *
+   */
+  fun cameraForAnnotations(
```
alexander-kulikovskii, line 190: is it expected behavior when using only one annotation that zoom changed?

ank27: Yes its expected. since we are using `cameraForCoodinates`, it will update cameraOptions which has zoom and center associated with that.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
ank27: I haven't seen any validation on `EdgeInsets`. if users provide really large edgeInsets, imo its their responsibility. I can however check how does `cameraForCoordinate` or `coordinateBoundsForCamera` works in case of these values.

ank27: gl-native log error when it couldn't calculate correct zoom, although it still provides wrong cameraOptions. so its upto users to provide correct edgeinsets or they will have errors in log.

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
```diff
@@ -166,6 +168,32 @@ interface ViewAnnotationManager {
...
```
kiryldz, line 190: @ank27 should it have `@JvmOverloads` as parameters by default are used>

kiryldz (approved):
LGTM with one question

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
yunikkk: > If another one (let's say annotation2 whose geometry is below annotation1) is bigger and its top is bigger than annotation1, then annotation2 will become topmost from the calculation

@ank27 not sure I see how this is implemented then. First we lookup a view annotation with max latitude of the bound geometry, view size is not taken into account, e.g.
```
val topAnnotation =
      filteredViewAnnotations.maxByOrNull { (it.geometry!! as Point).latitude() }
```

Then we use this topAnnotation to calculate view bounds and then edge inset
```
(edgeInsets?.top ?: 0).toDouble()
        .plus(abs(getViewAnnotationOptionsFrame(topAnnotation)?.top ?: 0))
```

Am I missing something here?

sdk/src/main/java/com/mapbox/maps/viewannotation/ViewAnnotationManager.kt
ank27: as discussed offline, kotlin doesn't support `@JvmOverloads` in interfaces.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Yeah i see your point here. i will handle this case in https://mapbox.atlassian.net/browse/MAPSAND-577.

yunikkk: 👍

yunikkk (approved):

alexander-kulikovskii (approved):

PR #1783, opened 2022-10-31
Test geojson to value mapping

yunikkk:
### Summary of changes

Added direct mapping for geojson -> Value mapping to save some cpu cycles. 

In general seems faster 2 times and more, depending on the size of the geojson.
However, `Expression.fromRaw` seem compatibly fast, so the question is - do we want to introduce these optimizations or just add some words suggesting using `fromRaw`, maybe on background thread for large expressions @mapbox/maps-android ? 

## Pull request checklist:
    

app/src/main/java/com/mapbox/maps/testapp/examples/DistanceExpressionActivity.kt
```diff
@@ -40,20 +43,59 @@ class DistanceExpressionActivity : AppCompatActivity() {
...
+          fillLayer(CIRCLE_ID2, CIRCLE_ID2) {
+            fillOpacity(0.5)
+            fillColor("#00b2d0")
+          },
+          below = POI_LABEL
+        )
       }
     ) {
-      val symbolLayer = it.getLayer("poi-label") as SymbolLayer
+      val symbolLayer = it.getLayer(POI_LABEL) as SymbolLayer
```
yunikkk, line 68: For sake of testing

extension-style/src/main/java/com/mapbox/maps/extension/style/expressions/generated/Expression.kt
```diff
@@ -2301,6 +2305,139 @@ class Expression : Value {
     fun cos(block: ExpressionBuilder.() -> Unit): Expression =
       ExpressionBuilder("cos").apply(block).build()
 
+    fun BoundingBox.toValue(): Value {
```
yunikkk, line 2308: Previously geojson was [serialized](https://github.com/mapbox/mapbox-java/blob/a86a6148b87ff3e4a9bd1096eaed5715b80db5e1/services-geojson/src/main/java/com/mapbox/geojson/BaseGeometryTypeAdapter.java) to json, then parsed back from string.

app/src/main/java/com/mapbox/maps/testapp/examples/DistanceExpressionActivity.kt
```diff
@@ -40,20 +43,59 @@ class DistanceExpressionActivity : AppCompatActivity() {
...
+      for (i in 0..iter) {
+        Expression.distance(lineString)
+      }
+      println("Distance time : ${System.nanoTime() - startNano}")
+
+      startNano = System.nanoTime()
+      for (i in 0..iter) {
+        Expression.fromRaw(LINESTRING_JSON)
+      }
+      println("From raw : ${System.nanoTime() - startNano}")
```
jush, line 91: Nit: ICYMI you can use Kotlin [measureNanoTime](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.system/measure-nano-time.html)
```suggestion
val timeInNanos = measureNanoTime {
      repeat(iter) {
        Expression.fromRaw(LINESTRING_JSON)
      }
}
println("From raw : ${timeInNanos}")
```

pengdev:
> In general seems faster 2 times and more, depending on the size of the geojson.
> However, Expression.fromRaw seem compatibly fast, so the question is - do we want to introduce these optimizations or just add some words suggesting using fromRaw, maybe on background thread for large expressions

@yunikkk I think user might prefer to use strongly typed GeoJson class, would it be possible to move this Utility to Value class itself? so it can be reused in other places as well? e.g. `Value.fromGeometry()`,`Value.fromFeature()`,`Value.fromFeatureCollection()`?
And inside the distance(geometry), we just use the fast conversion under the hood?

yunikkk:
Closing as discussed in favor of using `Expression.fromRaw` on the client end on the BG thread

PR #1591, opened 2022-08-12
Fix MapboxMap extension plugin functions throwing exceptions

kiryldz:
### Summary of changes

When using extension functions for camera and gesture plugins (e.g. `MapboxMap.flyTo` and others) we previously were relying on weak references. That could lead to runtime crash:
```kotlin
checkNotNull(cameraAnimationsPlugin?.get()) {
  "Camera plugin is not added as the part of MapInitOptions for given MapView."
}
```
not only when camera plugin was excluded (as was per design) but also let's say when `MapView` was already destroyed and extension function was accessed.

Fixed this by removing weak references and improving warning messages.

## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [ ] Optimize code for java consumption (`@JvmOverloads`, `@file:JvmName`, etc).
 - [ ] Add example if relevant.
 - [ ] Document any changes to public APIs.
 - [ ] Run `make update-api` to update generated api files, if there's public API changes, otherwise the `verify-api-*` CI steps might fail.
 - [x] Update [CHANGELOG.md](../CHANGELOG.md) or use the label 'skip changelog', otherwise `check changelog` CI step will fail.
 - [x] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` firstly and then port to `v10.[version]` release branch.

Fixes: < Link to related issues that will be fixed by this pull request, if they exist >

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-android/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1797, opened 2022-11-03
Trigger repaint after `BitmapWidget` is updated.

pengdev:
### Summary of changes

This PR adds `Widget.triggerRepaint` internal method, to be able to trigger immediate repaint of the widget.
Also updated the BitmapWidget to trigger repaint after bitmap, position or rotation is changed.

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+## Features ✨ and improvements 🏁
+* Add `Widget.triggerRepaint` protected API and updated the `BitmapWidget` to trigger repaint after widget is updated.
 
 # 10.10.0-beta.1 November 03, 2022
 ## Features ✨ and improvements 🏁
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 # main
 ## Features ✨ and improvements 🏁
-* Add `Widget.triggerRepaint` protected API and updated the `BitmapWidget` to trigger repaint after widget is updated.
+* Add `Widget.triggerRepaint` protected API and updated the `BitmapWidget` to trigger repaint after widget is updated. ([1797](https://github.com/mapbox/mapbox-maps-android/pull/1797))
 
 # 10.10.0-beta.1 November 03, 2022
 ## Features ✨ and improvements 🏁
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,10 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
-## Features ✨ and improvements 🏁
-* Add `Widget.triggerRepaint` protected API and updated the `BitmapWidget` to trigger repaint after widget is updated. ([1797](https://github.com/mapbox/mapbox-maps-android/pull/1797))
+
+# 10.10.0-rc.1
+## Bug fixes 🐞
+* Trigger repaint after `BitmapWidget` is updated. ([1797](https://github.com/mapbox/mapbox-maps-android/pull/1797))
 
 # 10.10.0-beta.1 November 03, 2022
 ## Features ✨ and improvements 🏁
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -219,12 +219,16 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
 
   override fun addWidget(widget: Widget) {
     val weakThis = WeakReference(this)
-    renderer.renderThread.addWidget(widget.also { it.setRepaintTrigger { weakThis.get()?.renderer?.scheduleRepaint() } })
+    widget.setRepaintTrigger {
+      weakThis.get()?.renderer?.scheduleRepaint()
+    }
+    renderer.renderThread.addWidget(widget)
     renderer.scheduleRepaint()
   }
 
   override fun removeWidget(widget: Widget): Boolean {
     val wasRemoved = renderer.renderThread.removeWidget(widget)
+    widget.setRepaintTrigger(null)
     if (wasRemoved) {
       renderer.scheduleRepaint()
     }
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -37,7 +37,6 @@ import com.mapbox.maps.renderer.MapboxRenderer
 import com.mapbox.maps.renderer.OnFpsChangedListener
 import com.mapbox.maps.renderer.RendererSetupErrorListener
 import com.mapbox.maps.renderer.widget.Widget
-import java.lang.ref.WeakReference
 
 internal class MapController : MapPluginProviderDelegate, MapControllable {
 
@@ -218,9 +217,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun addWidget(widget: Widget) {
-    val weakThis = WeakReference(this)
-    widget.setRepaintTrigger {
-      weakThis.get()?.renderer?.scheduleRepaint()
+    widget.setTriggerRepaintAction {
+      renderer.scheduleRepaint()
     }
     renderer.renderThread.addWidget(widget)
     renderer.scheduleRepaint()
@@ -229,7 +227,6 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   override fun removeWidget(widget: Widget): Boolean {
     val wasRemoved = renderer.renderThread.removeWidget(widget)
     if (wasRemoved) {
-      widget.setRepaintTrigger(null)
       renderer.scheduleRepaint()
     }
     return wasRemoved
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -37,6 +37,7 @@ import com.mapbox.maps.renderer.MapboxRenderer
 import com.mapbox.maps.renderer.OnFpsChangedListener
 import com.mapbox.maps.renderer.RendererSetupErrorListener
 import com.mapbox.maps.renderer.widget.Widget
+import java.lang.ref.WeakReference
 
 internal class MapController : MapPluginProviderDelegate, MapControllable {
 
@@ -217,7 +218,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun addWidget(widget: Widget) {
-    renderer.renderThread.addWidget(widget)
+    val weakThis = WeakReference(this)
+    renderer.renderThread.addWidget(widget.also { it.setRepaintTrigger { weakThis.get()?.renderer?.scheduleRepaint() } })
     renderer.scheduleRepaint()
   }
 
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -228,8 +228,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
 
   override fun removeWidget(widget: Widget): Boolean {
     val wasRemoved = renderer.renderThread.removeWidget(widget)
-    widget.setRepaintTrigger(null)
     if (wasRemoved) {
+      widget.setRepaintTrigger(null)
       renderer.scheduleRepaint()
     }
     return wasRemoved
--- a/sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
@@ -8,6 +8,15 @@ import com.mapbox.maps.MapboxExperimental
 @MapboxExperimental
 abstract class Widget internal constructor() {
   internal abstract val renderer: WidgetRenderer
+  private var repaintTriggerWeakReference: (() -> Unit)? = null
+
+  /**
+   * Triggers a repaint of the widget, as well as the map.
+   * It will do nothing if the widget has not been added to the map.
+   */
+  protected fun triggerRepaint() {
+    repaintTriggerWeakReference?.invoke()
+  }
 
   /**
    * Update the widget to the new position.
@@ -40,4 +49,9 @@ abstract class Widget internal constructor() {
    * Get absolute rotation of widget in degrees.
    */
   abstract fun getRotation(): Float
+
+  @JvmSynthetic
+  internal fun setRepaintTrigger(trigger: () -> Unit) {
+    repaintTriggerWeakReference = trigger
+  }
 }
--- a/sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
+++ b/sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
@@ -8,14 +8,15 @@ import com.mapbox.maps.MapboxExperimental
 @MapboxExperimental
 abstract class Widget internal constructor() {
   internal abstract val renderer: WidgetRenderer
-  private var repaintTrigger: (() -> Unit)? = null
+  private var triggerRepaintAction: (() -> Unit)? = null
 
   /**
    * Triggers a repaint of the widget, as well as the map.
    * It will do nothing if the widget has not been added to the map.
    */
-  protected fun triggerRepaint() {
-    repaintTrigger?.invoke()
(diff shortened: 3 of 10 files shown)


sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -217,7 +218,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun addWidget(widget: Widget) {
-    renderer.renderThread.addWidget(widget)
+    val weakThis = WeakReference(this)
```
jush, line 221: I'm not sure I understand why we need `WeakReference`.
Isn't it enough to remove it when `removeWidget` is called?

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
```diff
@@ -40,4 +49,9 @@ abstract class Widget internal constructor() {
    * Get absolute rotation of widget in degrees.
    */
   abstract fun getRotation(): Float
+
+  @JvmSynthetic
+  internal fun setRepaintTrigger(trigger: () -> Unit) {
```
jush, line 54: if we rename to `invalidate` this maybe could be called `addOnInvalidateListener`.

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
```diff
@@ -8,6 +8,15 @@ import com.mapbox.maps.MapboxExperimental
 @MapboxExperimental
 abstract class Widget internal constructor() {
   internal abstract val renderer: WidgetRenderer
+  private var repaintTriggerWeakReference: (() -> Unit)? = null
```
jush, line 11:
```suggestion
  private var repaintTrigger: (() -> Unit)? = null
```

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
```diff
@@ -8,6 +8,15 @@ import com.mapbox.maps.MapboxExperimental
...
+
+  /**
+   * Triggers a repaint of the widget, as well as the map.
+   * It will do nothing if the widget has not been added to the map.
+   */
+  protected fun triggerRepaint() {
```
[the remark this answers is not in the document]
pengdev: It not only invalidate the widgets, but will actually trigger a repaint of the map, including the widget, I'd keep the `triggerRepaint` as it behaves the same gl-native's `triggerRepaint` API.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -217,7 +218,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+    renderer.renderThread.addWidget(widget.also { it.setRepaintTrigger { weakThis.get()?.renderer?.scheduleRepaint() } })
```
kiryldz, line 222: looks pretty crazy imo 😄 
guess here Java style may look better:
```
widget.setRepaintTrigger {

}
renderer.renderThread.addWidget(widget)
```

jush (approved):

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: User might not call the `removeWidget`, and if the MapView is destroyed(e.g. the fragment that contains the mapview is detached from the view) while user keep the widget instance in their app, there could potentially be a memory leak.

But agree we should clean up the repaintTrigger when widget is removed.

pengdev: Reset the repaint trigger to null when the widget is removed from the map.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: refactored

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
kiryldz: we did not use `Trigger` anywhere else and I'm not sure I did see such naming before...
perhaps `doOnRepaint`?

sdk/src/main/java/com/mapbox/maps/MapController.kt
jush: Discussed offline. `removeWidget` might not be called under all circumstances.

Still, we should add removing the trigger in case the user retains the widget.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -217,12 +218,17 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
   }
 
   override fun addWidget(widget: Widget) {
+    val weakThis = WeakReference(this)
+    widget.setRepaintTrigger {
+      weakThis.get()?.renderer?.scheduleRepaint()
```
ank27, line 223: do we need `scheduleRepaint` again at L226? since we are doing `scheduleRepaint()` when adding widget here?

Also maybe change `weakThis` to `mapControllerWeakReference` to understand better.

sdk/src/main/java/com/mapbox/maps/MapController.kt
yunikkk: Could you elaborate please? It seems it's a clear user mistake to keep widgets / (or view annotations) etc created by the MapView

yunikkk: And if user does not keep Widgets explicitly - they'll be GC'ed together with the MapView itself, shouldn't be a problem it seems.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -217,12 +218,17 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
...
+    }
     renderer.renderThread.addWidget(widget)
     renderer.scheduleRepaint()
   }
 
   override fun removeWidget(widget: Widget): Boolean {
     val wasRemoved = renderer.renderThread.removeWidget(widget)
+    widget.setRepaintTrigger(null)
```
jush, line 231: Should this be inside the `if (wasRemoved)`?

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: > It seems it's a clear user mistake to keep widgets / (or view annotations) etc created by the MapView

Actually it's not, the Widget is usually created by user, for example CompassWidget should be created and the instance should be kept by the user, to be able to update the bearing etc later on.

> And if user does not keep Widgets explicitly - they'll be GC'ed together with the MapView itself, shouldn't be a problem it seems.

Not really, user can accidentally keep a reference of the widget even after the MapView is destroyed. Also [this](https://stackoverflow.com/questions/42271208/kotlin-safe-lambdas-no-memory-leak) would be an interesting read.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: > do we need scheduleRepaint again at L226?

yes, we should do repaint when widget is added to the map. The `setRepaintTrigger` is only useful when user modify the widget itself, but not when the constructed widget is added to the map.

> Also maybe change weakThis to mapControllerWeakReference to understand better.

It feels weakThis is meaningful enough as we are creating the WeakReference with `this`.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: fixed

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
pengdev: renamed to `doOnRepaint`

sdk/src/main/java/com/mapbox/maps/MapController.kt
yunikkk: > Actually it's not, the Widget is usually created by user, for example CompassWidget should be created and the instance should be kept by the user, to be able to update the bearing etc later on.

I agree - however how is it different from the View Annotation, for example? User might keep it to update some properties - obviously it will hold the parent MapView as well. It's absolutely the same case IMO.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+## Features ✨ and improvements 🏁
+* Add `Widget.triggerRepaint` protected API and updated the `BitmapWidget` to trigger repaint after widget is updated. ([1797](https://github.com/mapbox/mapbox-maps-android/pull/1797))
```
kiryldz, line 7: @pengdev I think technically we could still land this in RC, wdyt? see no major risks, this API is nice addition

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
```diff
@@ -8,6 +8,15 @@ import com.mapbox.maps.MapboxExperimental
 @MapboxExperimental
 abstract class Widget internal constructor() {
   internal abstract val renderer: WidgetRenderer
+  private var repaintTrigger: (() -> Unit)? = null
```
kiryldz, line 11:
```suggestion
  private var onRepaintAction: (() -> Unit)? = null
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: > however how is it different from the View Annotation, for example? User might keep it to update some properties - obviously it will hold the parent MapView as well. It's absolutely the same case IMO.

My argument here is that it's not a user mistake to keep the widgets that they constructed and update it at a later time, actually rather it's an expected usage pattern. I'm not sure I follow the comparison with ViewAnnotation. In this widget's particular case, we are keeping an lamda(potentially keeps a strong MapController) that invokes a method on MapController.

CHANGELOG.md
pengdev: There's one potential public API addition, but it's only visible for classes that extends Widget(where the constructor is internal). I think the public API surface is minimal, I'd agree to include in RC. (given the widget API is still experimental)

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
```diff
@@ -40,4 +49,9 @@ abstract class Widget internal constructor() {
    * Get absolute rotation of widget in degrees.
    */
   abstract fun getRotation(): Float
+
+  @JvmSynthetic
+  internal fun setRepaintTrigger(doOnRepaint: (() -> Unit)?) {
```
kiryldz, line 54: after looking a bit deeper and understanding what's happening here I guess it should be more like:
```
private var onTriggerRepaintAction: (() -> Unit)? = null

internal fun onTriggerRepaintAction(action: (() -> Unit)?) {
  onTriggerRepaintAction = action
}
```

cause now protected function is named `triggerRepaint` while here it's `setRepaintTrigger`, I didn't even connect them in the beginning...

pengdev: imho the suggested name is a bit more confusing for me, but as these are internal functions, I'd say we can leave it for now, unless someone else has strong opinion on the naming? cc @jush @yunikkk @ank27 @paulrenn67

CHANGELOG.md
kiryldz: if we decide to do so - we could already create v10.10.0-rc.1 section in changelog then as `#main` indicates it will be available in next minor release

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: @pengdev my 5 cents here are:
1. When using weak ref like now - that indeed potentially saves us from memory leak when map controller is destroyed but widgets (owned by user) are not. However if user will call `widget.triggerRepaint` after controller is destroyed - `weakThis.get()` will be null and this is no-op operation anyway. If same widget will be used with another map controller - in order for that to work widget will need to be re-added to the new map controller.
2. If not using weak ref and use code as now - I could believe we could have the memory leak.

I would actually remove weak ref and explicitly call `widget.setRepaintTrigger(null)` on all widgets that were _not_ removed by the user explicitly in `MapController.onDestroy`. In that case we retain current logic (no leaking) and don't mess with weak refs (we have removed almost all of them from codebase and I do not want to add more of them).

yunikkk: > My argument here is that it's not a user mistake to keep the widgets that they constructed and update it at a later time, actually rather it's an expected usage pattern.

Right, but what is expected if user adds widget to the mapView (that owns mapController), then keep it after the mapView/container is destroyed? The parallel with view annotation is that if user creates view annotation, then keeps it while mapView is destroyed - it'll also hold the whole mapView, and this is expected, right?

pengdev: > don't mess with weak refs (we have removed almost all of them from codebase and I do not want to add more of them).

Maybe I'm out of the loop, any reasons we want to avoid the usage of weak refs? just want have a deeper understanding for future reference.

I also agree to clean up all the widgets on MapController.destroy(), and probably that's the best solution here, actually will refactor it that way.

> is expected if user adds widget to the mapView (that owns mapController), then keep it after the mapView/container is destroyed?

It's not expected/preferred, but will be what many users do regardless, in any case, we should avoid a memory leak if possible.

yunikkk: > Maybe I'm out of the loop, any reasons we want to avoid the usage of weak refs? just want have a deeper understanding for future reference.

Problem with weakrefs is that they tend to hide the issues with the calling or our code, and then leads to subtle bugs, iirc we've faced them with style/maps/locationengine several times. 

And if we treat widget/child view (view annotation or compass or logo plugin) of the map as a _child_ of the map (this is my understanding of both) its imo it's fine to keep the reference to the parent, if child is stored. It's same case with regular Android classes, if you keep reference to the view - the hosting fragment/activity will also be kept.

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
yunikkk: While it's indeed internal API and probably does not matter that much I'd probably renamed it to use `triggerRepaint` consistently, maybe 
```
private var triggerRepaintAction: (() -> Unit)? = null internal fun setTriggerRepaintAction(action: (() -> Unit)?) { triggerRepaintAction = action }

  protected fun triggerRepaint() {
    triggerRepaintAction?.invoke()
  }
```

sdk/src/main/java/com/mapbox/maps/MapController.kt
kiryldz: > Maybe I'm out of the loop, any reasons we want to avoid the usage of weak refs? just want have a deeper understanding for future reference.

I think the most obvious reason (at least that's my IMO) is that that puts our SDK out of control. GC could be called literally at any moment of time by the system and it makes it pretty hard to reproduce / fix any issues connected with that. We had several examples when rare undetermined errors were reported by customers, the latest one I found is https://github.com/mapbox/mapbox-maps-android/pull/1591.

CHANGELOG.md
yunikkk: Btw - do we need to mention the internal API? Maybe just say that `update the widget when a property changes` ?

sdk/src/main/java/com/mapbox/maps/renderer/widget/Widget.kt
pengdev: refactored as suggested

CHANGELOG.md
pengdev: As you can see there's changes in the api file, so definitely there's API changes, however given the fact that user can not actually use the new APIs(only extendable for internal classes), maybe I can just make it internal, so there wouldn't be additions in the API file.

Updated, please take another look.

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: Thanks @yunikkk @yunikkk for the thoughts, I'd agree to avoid usage of weak reference if possible, refactored the widgets renderer to do proper clean up during destroy, please take another look.

kiryldz (approved):
💎

jush (approved):

ank27 (approved):

PR #1801, opened 2022-11-04
Set events as queued by default

brunoabinader:
### Summary of changes

This is a one-liner change that changes the events priority to queued by default. Queued events are batched in groups of up to 180 events and sent as a single gzip-compressed HTTP request, thus reducing network bandwidth consumption.

Interaction and feedback events are not time-critical, so there is no need for marking these as immediate priority.

No added tests as this logic implementation is part of Common SDK.

### User impact (optional)

Because we're marking Maps SDK interaction and feedback events as queued, there will be a reduction in terms of HTTP requests generated from the SDK.

## Pull request checklist:
    
### DIFF
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -161,6 +161,8 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     renderer.onStop()
     pluginRegistry.onStop()
+    // flush the queued events before destroy to avoid lost telemetry events
+    MapProvider.flushPendingEvents(mapInitOptions.resourceOptions.accessToken)
   }
 
   override fun onDestroy() {
@@ -168,8 +170,6 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       return
     }
     lifecycleState = LifecycleState.STATE_DESTROYED
-    // flush the queued events before destroy to avoid lost telemetry events
-    MapProvider.flushPendingEvents(mapInitOptions.resourceOptions.accessToken)
     pluginRegistry.onDestroy()
     nativeObserver.onDestroy()
     renderer.onDestroy()
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -169,11 +169,7 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
     }
     lifecycleState = LifecycleState.STATE_DESTROYED
     // flush the queued events before destroy to avoid lost telemetry events
-    MapProvider.getEventService(mapInitOptions.resourceOptions.accessToken).flush { expected ->
-      expected.error?.let { error ->
-        logE(TAG, "EventService flush error: $error")
-      }
-    }
+    MapProvider.flushPendingEvents(mapInitOptions.resourceOptions.accessToken)
     pluginRegistry.onDestroy()
     nativeObserver.onDestroy()
     renderer.onDestroy()
--- a/sdk/src/main/java/com/mapbox/maps/MapController.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapController.kt
@@ -168,6 +168,12 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       return
     }
     lifecycleState = LifecycleState.STATE_DESTROYED
+    // flush the queued events before destroy to avoid lost telemetry events
+    MapProvider.getEventService(mapInitOptions.resourceOptions.accessToken).flush { expected ->
+      expected.error?.let { error ->
+        logE(TAG, "EventService flush error: $error")
+      }
+    }
     pluginRegistry.onDestroy()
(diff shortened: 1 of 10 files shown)


pengdev (approved):
LGTM with a nit

1ec5:
> Because we're marking Maps SDK interaction and feedback events as queued, there will be a reduction in terms of HTTP requests generated from the SDK.

Since there is user impact, perhaps there should be a changelog entry and this PR shouldn’t be labeled https://github.com/mapbox/mapbox-maps-android/labels/skip%20changelog.

pengdev (changes requested):
Per discussion offline, I think we need to call `EventsService.flush()` during the MapController's onDestroy to avoid losing events.

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -168,6 +168,12 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       return
     }
     lifecycleState = LifecycleState.STATE_DESTROYED
+    // flush the queued events before destroy to avoid lost events
```
kiryldz, line 171: can't say I fully understand what is 'lost event' here; if I understand correctly `flush` here clears the queue but how bad is it if we have that 'lost event'? could it be a memory leak?

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -50,6 +53,12 @@ internal object MapProvider {
     return mapTelemetry
   }
 
+  fun getEventService(accessToken: String): EventsService {
```
yunikkk, line 56: nit : getEventsService

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -50,6 +53,12 @@ internal object MapProvider {
...
+    val eventsServiceOptions =
+      EventsServerOptions(accessToken, BuildConfig.MAPBOX_EVENTS_USER_AGENT, null)
```
yunikkk, line 58: If the token is passed from outside - maybe pass useragent as well?

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -50,6 +53,12 @@ internal object MapProvider {
...
+    return EventsService.getOrCreate(eventsServiceOptions)
```
jush, line 59: What happens if the options are different than the ones inside the `MapTelemetryImpl` constructor?

sdk/src/main/java/com/mapbox/maps/MapController.kt
```diff
@@ -168,6 +168,12 @@ internal class MapController : MapPluginProviderDelegate, MapControllable {
       return
     }
     lifecycleState = LifecycleState.STATE_DESTROYED
+    // flush the queued events before destroy to avoid lost telemetry events
+    MapProvider.getEventService(mapInitOptions.resourceOptions.accessToken).flush { expected ->
```
jush, line 172: Nit: Would it be better instead of getting the event service have a `MapProvider.flushPendingEvents`?

sdk/src/main/java/com/mapbox/maps/MapController.kt
pengdev: some of the telemetry events will be lost, which is not desirable, but no memory leak 😅

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
pengdev: If we expose both, I guess no need to wrap to this object class 😅 The `MAPBOX_EVENTS_USER_AGENT` is not supposed to change with each build, the access token can be runtime-configured on the other hand.

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
jush: I can answer myself. It seems a new service will be created.

Therefore, I'm a bit concerned that we would need to maintain this `options` and the `MapTelemetryImpl` one in sync.

pengdev: then I assume we will create a new instance of `EventsService`, which is not correct. We need to make sure to use the same configuration from MapTelemetryImpl.

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
yunikkk: As far as I understood this class - it just abstracts creation of objects, that eases testing =) But anyway - not critical, up to you to decide

jush (approved):

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
pengdev: refactored to flushEvents as suggested by @jush

yunikkk (approved):

pengdev (approved):

jush (approved):

yunikkk (approved):

PR #1833, opened 2022-11-14
[widget] Fix BitmapWidget.setTranslation API to offset relative to its original position.

pengdev:
### Summary of changes

Fix BitmapWidget.setTranslation API to offset relative to its original position.

### User impact (optional)

With v10.10.0-beta.1 release, we [refactored](https://github.com/mapbox/mapbox-maps-android/pull/1782) the Bitmap position API. We accidentally introduced a behavioural change: setTranslation was changed to be relative to its current position, however the behaviour in v10.9.0 is to be relative to its original position.

This PR reverts this behavioural change so it's compatible to previous implementation.

## Pull request checklist:
    

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidget.kt
```diff
@@ -12,7 +12,7 @@ import com.mapbox.maps.MapboxExperimental
 @MapboxExperimental
 open class BitmapWidget @JvmOverloads constructor(
   bitmap: Bitmap,
-  position: WidgetPosition = WidgetPosition {
+  private val position: WidgetPosition = WidgetPosition {
```
jush, line 15: Nit: even though it's private,  maybe rename it to `originalPosition` to avoid any confusion with `getPosition()`.

jush (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Mapbox welcomes participation and contributions from everyone.
 # 10.10.0-rc.1
 ## Bug fixes 🐞
 * Trigger repaint after `BitmapWidget` is updated. ([1797](https://github.com/mapbox/mapbox-maps-android/pull/1797))
+* Fix BitmapWidget.setTranslation API to offset relative to its original position. ([1833](https://github.com/mapbox/mapbox-maps-android/pull/1833))
```
kiryldz, line 10: just a point to consider - when we will be cutting the final v10.10.0 version - that bug fix relative to v10.9.0 changelog might be confusing (as there's no bug if we compare v10.9.0 and v10.0.0). Mayber add a note that this fix is relevant only comparing with beta?

pengdev: yeah, I was thinking to remove this change log entry in stable release note.

sdk/src/main/java/com/mapbox/maps/renderer/widget/BitmapWidget.kt
pengdev: sounds good, do you think we should expose it as public API?

jush: No, I don't think we need to expose it for now.
Let's do so when there's a valid use-case. Moreover, the user can keep a copy of it around.

yunikkk (approved):

kiryldz (approved):

kmadsen (approved):
`10.10.0-beta.1-MAPSAND-608-bitmap-widget-changes-in-10-10-break-alignment-for-10-9-upgrades-df1760069-SNAPSHOT`

I tested this snapshot on a downstream upgrade branch https://github.com/mapbox/mapbox-navigation-android/pull/6588

And the issue is not happening 👍 .

Also the changes LGTM, downstream relies on an original position and then a translation from that position.

CHANGELOG.md
pengdev:
```suggestion
* Fix a regression from v10.10.0-beta.1 release, the BitmapWidget.setTranslation API should offset relative to its original position. ([1833](https://github.com/mapbox/mapbox-maps-android/pull/1833))
```

PR #1842, opened 2022-11-15
[camera] Fix immediate camera animation on API level 23 or below - MAPSAND-614

pengdev:
### Summary of changes

This PR bypasses the immediate camera animation(where duration and start delay of the animation is 0) on the animator, and emit a single animator update instead of trying to start the ValueAnimator directly.

More context on the issue with animators on API 23 or below:
> Devices with API <= 23 animating with duration = 0 will send initial value from Looper with some small delay as a result.
> Hence multiple immediate animations sent close to each other may cancel previous ones.

### User impact (optional)

This change particularly affects the camera animation and viewport plugin on API level 23 or below.
For example viewport plugin's `FollowPuckViewportState` behaviour will be fixed as shown below:

| Before | After |
| ----- | ----- |
| <video src="https://user-images.githubusercontent.com/2764714/202478245-d6b1ebc4-cbae-4a44-9cbf-90f81b2008fe.webm" width = 250/> | <video src="https://user-images.githubusercontent.com/2764714/202478303-2d5a7f57-6f75-479a-aa69-0063d620a086.webm" width = 250/> |

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+* Fix immediate camera animation on API level 23. ([1842](https://github.com/mapbox/mapbox-maps-android/pull/1842))
 
 # 10.10.0-rc.1
 ## Bug fixes 🐞
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
-* Fix immediate camera animation on API level 23. ([1842](https://github.com/mapbox/mapbox-maps-android/pull/1842))
+* Fix immediate camera animation on API level 23 or below. ([1842](https://github.com/mapbox/mapbox-maps-android/pull/1842))
 
 # 10.10.0-rc.1
 ## Bug fixes 🐞
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
@@ -9,7 +9,6 @@ import com.mapbox.maps.plugin.locationcomponent.LocationConsumer
 import com.mapbox.maps.plugin.locationcomponent.LocationProvider
 import com.mapbox.maps.plugin.locationcomponent.location
 import com.mapbox.maps.plugin.viewport.ViewportPlugin
-import com.mapbox.maps.plugin.viewport.state.FollowPuckViewportState
 import com.mapbox.maps.plugin.viewport.transition.ViewportTransition
 import com.mapbox.maps.plugin.viewport.viewport
 import com.mapbox.maps.testapp.BaseMapTest
@@ -31,7 +30,6 @@ class ViewportPluginTest : BaseMapTest() {
 
   private val handler = Handler(Looper.getMainLooper())
   private lateinit var viewportPlugin: ViewportPlugin
-  private lateinit var followPuckViewportState: FollowPuckViewportState
   private lateinit var immediateViewportTransition: ViewportTransition
 
   private val locationProvider = object : LocationProvider {
@@ -51,7 +49,6 @@ class ViewportPluginTest : BaseMapTest() {
     rule.scenario.onActivity { activity ->
       activity.runOnUiThread {
         viewportPlugin = mapView.viewport
-        followPuckViewportState = viewportPlugin.makeFollowPuckViewportState()
         immediateViewportTransition = viewportPlugin.makeImmediateViewportTransition()
         mapView.location.apply {
           enabled = true
@@ -86,13 +83,15 @@ class ViewportPluginTest : BaseMapTest() {
         )
       }
       // transition to the followPuckViewportState with default transition
-      viewportPlugin.transitionTo(followPuckViewportState) {
+      viewportPlugin.transitionTo(viewportPlugin.makeFollowPuckViewportState()) {
         latch.countDown()
       }
     }
 
     // Wait for 5 seconds since the default transition time is 3.5 seconds
-    latch.await(5000, TimeUnit.SECONDS)
+    if (!latch.await(5, TimeUnit.SECONDS)) {
+      throw TimeoutException()
+    }
     handler.post {
       val cameraState = mapView.getMapboxMap().cameraState
       cameraState.center.assertEquals(TEST_POINT)
@@ -120,12 +119,12 @@ class ViewportPluginTest : BaseMapTest() {
         )
       }
       // immediately transition to the followPuckViewportState
-      viewportPlugin.transitionTo(followPuckViewportState, immediateViewportTransition) {
+      viewportPlugin.transitionTo(viewportPlugin.makeFollowPuckViewportState(), immediateViewportTransition) {
         latch.countDown()
       }
     }
 
-    if (!latch.await(2, TimeUnit.SECONDS)) {
+    if (!latch.await(200, TimeUnit.MILLISECONDS)) {
       throw TimeoutException()
     }
 
@@ -142,19 +141,19 @@ class ViewportPluginTest : BaseMapTest() {
     handler.post {
       // immediate update location puck to test location.
       locationProvider.locationConsumers.forEach {
-        it.onLocationUpdated(
-          TEST_POINT,
-          options = { duration = 0 }
-        )
-      }
-      locationProvider.locationConsumers.forEach {
         it.onBearingUpdated(
           TEST_BEARING,
           options = { duration = 0 }
         )
       }
+      locationProvider.locationConsumers.forEach {
+        it.onLocationUpdated(
+          TEST_POINT,
+          options = { duration = 0 }
+        )
+      }
       // immediately transition to the followPuckViewportState
-      viewportPlugin.transitionTo(followPuckViewportState, immediateViewportTransition) {
+      viewportPlugin.transitionTo(viewportPlugin.makeFollowPuckViewportState(), immediateViewportTransition) {
         // now we are in the follow puck viewport state
         val cameraState = mapView.getMapboxMap().cameraState
         cameraState.center.assertEquals(TEST_POINT)
@@ -182,8 +181,8 @@ class ViewportPluginTest : BaseMapTest() {
     // validate the camera is at the moved location
     handler.post {
       val cameraState = mapView.getMapboxMap().cameraState
-      assertEquals(TEST_BEARING + 90.0, cameraState.bearing, EPS)
       cameraState.center.assertEquals(TEST_POINT_MOVED)
+      assertEquals(TEST_BEARING + 90.0, cameraState.bearing, EPS)
     }
   }
 
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
@@ -91,7 +91,8 @@ class ViewportPluginTest : BaseMapTest() {
       }
     }
 
-    latch.await(10, TimeUnit.SECONDS)
+    // Wait for 5 seconds since the default transition time is 3.5 seconds
+    latch.await(5000, TimeUnit.SECONDS)
     handler.post {
       val cameraState = mapView.getMapboxMap().cameraState
       cameraState.center.assertEquals(TEST_POINT)
@@ -124,7 +125,7 @@ class ViewportPluginTest : BaseMapTest() {
       }
     }
 
-    if (!latch.await(10, TimeUnit.SECONDS)) {
+    if (!latch.await(2, TimeUnit.SECONDS)) {
       throw TimeoutException()
     }
 
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
@@ -139,44 +139,51 @@ class ViewportPluginTest : BaseMapTest() {
   fun testFollowPuckViewportState() {
     val latch = CountDownLatch(1)
     handler.post {
+      assertEquals(0.0, mapView.getMapboxMap().cameraState.bearing, EPS)
+      mapView.getMapboxMap().cameraState.center.assertEquals(NULL_ISLAND)
       // immediate update location puck to test location.
-      locationProvider.locationConsumers.forEach {
-        it.onBearingUpdated(
-          TEST_BEARING,
-          options = { duration = 0 }
-        )
-      }
       locationProvider.locationConsumers.forEach {
         it.onLocationUpdated(
           TEST_POINT,
           options = { duration = 0 }
         )
       }
+      locationProvider.locationConsumers.forEach {
+        it.onBearingUpdated(
+          TEST_BEARING,
+          options = { duration = 0 }
+        )
+      }
       // immediately transition to the followPuckViewportState
       viewportPlugin.transitionTo(viewportPlugin.makeFollowPuckViewportState(), immediateViewportTransition) {
-        // now we are in the follow puck viewport state
-        val cameraState = mapView.getMapboxMap().cameraState
-        cameraState.center.assertEquals(TEST_POINT)
-        assertEquals(TEST_BEARING, cameraState.bearing, EPS)
-        // emit new bearing and location updates, location component plugin should be driving the animation.
-        // and viewport plugin should do animation with 0 duration on each animated location puck position
-        locationProvider.locationConsumers.forEach {
-          it.onBearingUpdated(
-            TEST_BEARING + 90.0,
-            options = { duration = 1000 }
-          )
-        }
-        locationProvider.locationConsumers.forEach {
-          it.onLocationUpdated(
-            TEST_POINT_MOVED,
-            options = { duration = 1000 }
-          )
-        }
+        latch.countDown()
+      }
+    }
+
+    if (!latch.await(200, TimeUnit.MILLISECONDS)) {
+      throw TimeoutException()
+    }
+
+    val latch2 = CountDownLatch(1)
+    handler.post {
+      // emit new bearing and location updates, location component plugin should be driving the animation.
+      // and viewport plugin should do animation with 0 duration on each animated location puck position
+      locationProvider.locationConsumers.forEach {
+        it.onBearingUpdated(
+          TEST_BEARING + 90.0,
+          options = { duration = 1000 }
+        )
+      }
+      locationProvider.locationConsumers.forEach {
+        it.onLocationUpdated(
+          TEST_POINT_MOVED,
+          options = { duration = 1000 }
+        )
       }
     }
 
     // The location update will be animated with 1 second duration, we wait for 2 seconds for the animation to finish
-    latch.await(2, TimeUnit.SECONDS)
+    latch2.await(2, TimeUnit.SECONDS)
 
     // validate the camera is at the moved location
     handler.post {
--- a/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
+++ b/app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
@@ -1,6 +1,5 @@
 package com.mapbox.maps.testapp.viewport
 
-import android.os.Build
 import android.os.Handler
 import android.os.Looper
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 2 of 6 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -81,6 +81,21 @@ abstract class CameraAnimator<out T> (
    */
   final override fun start() {
     postOnAnimatorThread {
+      if (duration == 0L && startDelay == 0L) {
```
yunikkk, line 84: Should we limit it to run on API <= 23 only ?

pengdev: refactored the code a bit, please take another look.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -93,6 +97,55 @@ abstract class CameraAnimator<out T> (
...
+   *
+   * @return true if the animation is handled, false otherwise
+   */
+  private fun handleImmediateAnimationOnAPI23(): Boolean {
+    // Devices with API <= 23 animating with duration = 0 will send initial value from Looper with some small delay as a result.
+    // Hence multiple immediate animations sent close to each other may cancel previous ones.
+    // For these cases, we bypass the ValueAnimator and emit the user registered AnimatorListeners and AnimatorUpdateListeners immediately.
+    if (Build.VERSION.SDK_INT <= 23) {
+      if (duration == 0L && startDelay == 0L) {
+        val tmpListeners = listeners.toList()
```
yunikkk, line 131: Why is `toList` needed btw? To avoid ConcurrentModificationException? I guess it's still possible, if list is being changed from the different thread, `toList` also iterates the list. 
We probable need to make listeners CopyOnWriteArraySet to avoid any problems

pengdev: sounds good!

pengdev: I've changed our declared variables to CopyOnWriteArraySet, but kept the tmpListeners since it's from `ValueAnimator.getListeners()`, and it's handled in the same way in ValueAnimator itself: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/animation/ValueAnimator.java#1329

yunikkk: Indeed... Might be a problem, though will happen only with 23 API and animations with 0 lenght, probably won't happen that often 😆

pengdev: > avoid ConcurrentModificationException?

I encountered ConcurrentModificationException on my first try to run this fix 🙈

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
```diff
@@ -273,7 +273,7 @@ class CameraAnimationsListenersTest {
     bearingAnimator.start()
     Shadows.shadowOf(Looper.getMainLooper()).idle()
 
-    Assert.assertEquals(5, valuesList.size)
+    Assert.assertEquals(3, valuesList.size)
```
pengdev, line 276: The number of triggered events has been reduced here, because I've changed the `userUpdateListeners` from mutable list to `CopyOnWriteArraySet`.
As we are adding the same user update listeners twice, I believe the correct behaviour is to ignore the second addlistener, as it will result in duplicated update events. I'd consider this maybe a bug fix instead, wdyt? cc @kiryldz @yunikkk

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -93,6 +98,54 @@ abstract class CameraAnimator<out T> (
...
```
yunikkk, line 132: @pengdev confirming that it does not throw ConcurrentModification anymore and we dont need to synchronize somehow?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -93,6 +98,54 @@ abstract class CameraAnimator<out T> (
...
+   * @return animatedValue The value most recently calculated by this <code>ValueAnimator</code> for
+   * the single property being animated. If there are several properties being animated
+   * (specified by several PropertyValuesHolder objects in the constructor), this function
+   * returns the animated value for the first of those objects.
+   */
+  override fun getAnimatedValue(): Any {
+    // For immediate animations on API 23, as we introduced the bypass logic in handleImmediateAnimationOnAPI23(),
+    // the returned ValueAnimator.getAnimatedValue() will be null, in this case, we should return the
+    // last configured target value, so we immediately jump to the target value.
+    return super.getAnimatedValue() ?: targets.last() as Any
```
yunikkk, line 117: nit - maybe modify returned value only for API 23 also?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: yes, confirmed no issues after make a clone of the listeners before use them.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -93,6 +98,54 @@ abstract class CameraAnimator<out T> (
...
+   * Handle immediate animation(when duration and startDelay of the animation is 0) on devices running
+   * API 23 or below, by sending updates to animator listeners directly without triggering [ValueAnimator]'s start().
+   *
+   * @return true if the animation is handled, false otherwise
+   */
+  private fun handleImmediateAnimationOnAPI23(): Boolean {
+    // Devices with API <= 23 animating with duration = 0 will send initial value from Looper with some small delay as a result.
+    // Hence multiple immediate animations sent close to each other may cancel previous ones.
+    // For these cases, we bypass the ValueAnimator and emit the user registered AnimatorListeners and AnimatorUpdateListeners immediately.
+    if (Build.VERSION.SDK_INT <= 23) {
```
alexander-kulikovskii, line 135: should we add some instrumentation tests?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: 🤔 maybe, to limit the scope of this PR even more

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -93,6 +98,54 @@ abstract class CameraAnimator<out T> (
...
+    return super.getAnimatedValue() ?: targets.last() as Any
+  }
+
+  /**
+   * Handle immediate animation(when duration and startDelay of the animation is 0) on devices running
+   * API 23 or below, by sending updates to animator listeners directly without triggering [ValueAnimator]'s start().
+   *
+   * @return true if the animation is handled, false otherwise
+   */
+  private fun handleImmediateAnimationOnAPI23(): Boolean {
```
alexander-kulikovskii, line 131:
```suggestion
  private fun handleImmediateAnimationOnAPI23OrBelow(): Boolean {
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+* Fix immediate camera animation on API level 23. ([1842](https://github.com/mapbox/mapbox-maps-android/pull/1842))
```
alexander-kulikovskii, line 6:
```suggestion
* Fix immediate camera animation on API level 23 or below. ([1842](https://github.com/mapbox/mapbox-maps-android/pull/1842))
```

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: Fixed.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: we do have instrumentation tests on API level 23, these are used in both gestures and viewport.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: technically our minimum supported API level is 23 🙈

pengdev: right, I thought our minimum supported API is 23, but it's 21 indeed, will fix

alexander-kulikovskii: https://github.com/mapbox/mapbox-maps-android/blob/main/buildSrc/src/main/kotlin/Project.kt#L2
okay :wink:

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
yunikkk: Yeah I agree it should be correct to deliver the animated value once to the same listener. However I cant say I understand the test fully. Why is it 3 now since? Does it send start and end value for the animation with duration 0?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
alexander-kulikovskii: and after this fix nothing changed in the tests. so how do they work before? :see_no_evil:

pengdev: we have implemented workarounds in gestures, which solves the issue, maybe I will add more test to the viewport.

kiryldz:
@pengdev can you try running tests locally on API = 23 device? we ignore them now as per https://github.com/mapbox/mapbox-maps-android/blob/651553fd9b573e2ef2f3c7da037971327730495f/sdk/src/androidTest/java/com/mapbox/maps/CameraAnimationsPluginTest.kt#L1027

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
pengdev: to be honest I've no idea why it was 5 before, maybe @kiryldz have more context on it?

pengdev:
> can you try running tests locally on API = 23 device? we ignore them now as per

@kiryldz I tried it, still `testEaseToStartDelay` and `testEaseToStartDelayCancelled` are failing, others are passing. Also I've tested the result is the same in the main branch. 

Given the scope of this PR only helps immediate animation with both duration and startDelay is set to 0. Probably we should add separate tests this use case. I will try to improve the tests as well.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
pengdev: Brought back ignored tests in CameraAnimationsPluginTest, and added followpuck viewport state test, where under the hood, it uses immediate camera updates.

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
```diff
@@ -70,52 +68,134 @@ class ViewportPluginTest : BaseMapTest() {
...
+      }
+      // immediately transition to the followPuckViewportState
+      viewportPlugin.transitionTo(followPuckViewportState, immediateViewportTransition) {
+        latch.countDown()
+      }
     }
-    val latch = CountDownLatch(1)
-    latch.await(200, TimeUnit.MILLISECONDS)
+
+    if (!latch.await(10, TimeUnit.SECONDS)) {
```
alexander-kulikovskii, line 127: maybe if smth goes wrong we shouldn't wait here so long?
2 seconds and throw exception

pengdev: Reduced the timeout

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
```diff
@@ -70,52 +68,135 @@ class ViewportPluginTest : BaseMapTest() {
...
+      }
+      // transition to the followPuckViewportState with default transition
+      viewportPlugin.transitionTo(followPuckViewportState) {
+        latch.countDown()
+      }
     }
 
-    val latch = CountDownLatch(1)
-    latch.await(4000, TimeUnit.MILLISECONDS)
+    // Wait for 5 seconds since the default transition time is 3.5 seconds
```
yunikkk, line 94: Is it still 3.5 seconds for the immediate transition?

pengdev: No, for immediate transition we use 2 seconds timeout as below, and this is just worst case timeout, we do count down the latch in the transition finished callback, so it should Finnish pretty fast.

pengdev: actually updated to 200ms timeout, as I don't see this operation should take that much time, if it does, then probably there's something wrong.

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
```diff
@@ -70,52 +68,135 @@ class ViewportPluginTest : BaseMapTest() {
...
+    latch.await(5000, TimeUnit.SECONDS)
```
jush, line 95: Isn't this now waiting 5000 seconds?!?
```suggestion
    latch.await(5, TimeUnit.SECONDS)
```

jush, line 95: Should we do anything with the `latch.await` result?

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
```diff
@@ -70,52 +68,135 @@ class ViewportPluginTest : BaseMapTest() {
...
+      viewportPlugin.transitionTo(followPuckViewportState, immediateViewportTransition) {
+        latch.countDown()
+      }
     }
-    val latch = CountDownLatch(1)
-    latch.await(200, TimeUnit.MILLISECONDS)
+
+    if (!latch.await(200, TimeUnit.MILLISECONDS)) {
+      throw TimeoutException()
+    }
```
jush, line 130: Why not assert?
```suggestion
    Assert.assertTrue(latch.await(200, TimeUnit.MILLISECONDS))
```

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
pengdev: good catch 🙈

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
pengdev: I think `TimeoutException` might be more clear message in this case, and easier to figure out some thing might be flaky when looking at the report.

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
pengdev: 👍 added check

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
alexander-kulikovskii: or you can create extension for `count down latch`
`throwExceptionIfNotAwait(timeMs:Long)`

but it would be local function. we need extra module for such functions

.circleci/config.yml
```diff
@@ -356,7 +356,7 @@ commands:
                   --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
                   --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
                   --device model=blueline,version=28,locale=en,orientation=portrait \
-                  --timeout 45m --use-orchestrator --no-record-video --num-flaky-test-attempts 1 \
```
yunikkk, line 359: Why was the --no-record-video introduced at all, do you have the context? I haven't added it in Flutter config and it's really convenient to see what's going on with the test actually

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
yunikkk: It's still 5 seconds delay here, seems 200 ms was only changed in a differet test

yunikkk (approved):

.circleci/config.yml
pengdev: to save cost I assume 😄

app/src/androidTest/java/com/mapbox/maps/testapp/viewport/ViewportPluginTest.kt
pengdev: yeah, here it should be 5 second delay in this test, as the transition itself takes 3.5 seconds, we need to wait for that transition to finish.

ank27 (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
```diff
@@ -42,8 +44,8 @@ abstract class CameraAnimator<out T> (
   private var internalUpdateListener: AnimatorUpdateListener? = null
   private var internalListener: AnimatorListener? = null
 
-  private val userUpdateListeners = mutableListOf<AnimatorUpdateListener?>()
-  private val userListeners = mutableListOf<AnimatorListener?>()
+  private val userUpdateListeners = CopyOnWriteArraySet<AnimatorUpdateListener?>()
+  private val userListeners = CopyOnWriteArraySet<AnimatorListener?>()
```
kiryldz, line 48: why are those copy on write sets? first issue I see - why is it a set, imo adding duplicates should be allowed; another thing - we is it copy on write? we indeed need to operate on local copy of `listeners` below but are those needed to be copy on write?

pengdev: I've also started the conv in above comment [here](https://github.com/mapbox/mapbox-maps-android/pull/1842#discussion_r1023831004)

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsListenersTest.kt
kiryldz: why is adding the same listener considered to be a bug?

pengdev: Adding same listener twice effective means for each camera update, user will be notified twice, is that expected though? What's the use case for this behaviour?

pengdev: And the listeners behaviour I think so far is aligned across the codebase, do you see issues with that approach?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/animator/CameraAnimator.kt
kiryldz: discussed offline, seems reasonable

PR #1858, opened 2022-11-21
[testapp] fix example app build for playstore publish

ank27:
### Summary of changes
- Fixes playstore error to target api 31 (see screenshot).
- Update `tagsize` in gradle script to fix automatic deploy to playstore

Screenshot 2022-11-17 at 4 37 13 PM

## Pull request checklist:
    
### DIFF
--- a/buildSrc/src/main/kotlin/Project.kt
+++ b/buildSrc/src/main/kotlin/Project.kt
@@ -12,6 +12,11 @@ object AndroidVersions {
     const val targetSdkVersion = 30
     const val compileSdkVersion = 31
   }
+  object ExampleApp {
+    const val minSdkVersion = 23
+    const val targetSdkVersion = 31
+    const val compileSdkVersion = 31
+  }
 }
 
 object Plugins {
--- a/gradle/script-git-version.gradle
+++ b/gradle/script-git-version.gradle
@@ -11,7 +11,8 @@ import org.ajoberstar.grgit.Grgit
 
 ext {
     git = Grgit.open(currentDir: projectDir)
-    def tagSize = git.tag.list().findAll { (it.name =~ "android-v*" || it.name =~ "v1\\d.*") }.size()
+    // added extra 1 to fix playstore error regarding versioncode
+    def tagSize = git.tag.list().findAll { (it.name =~ "android-v*" || it.name =~ "v1\\d.*") }.size() + 1
     gitVersionCode = tagSize + 1
     gitVersionName = "0.1.$tagSize".toString()
 }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -12,7 +12,7 @@ apply {
 val buildFromSource: String by project
 
 android {
-  compileSdk = AndroidVersions.compileSdkVersion
+  compileSdk = AndroidVersions.ExampleApp.compileSdkVersion
   signingConfigs {
     create("release") {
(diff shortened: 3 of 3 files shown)


buildSrc/src/main/kotlin/Project.kt
```diff
@@ -12,6 +12,11 @@ object AndroidVersions {
     const val targetSdkVersion = 30
     const val compileSdkVersion = 31
   }
+  object ExampleApp {
```
yunikkk, line 15: Should we migrate everywhere to target 31 if Google Play now enforces it?

ank27: for sdk we can keep as it is, since its upto developers who uses maps-sdk to use targetSdk as 31 for app to deploy. but we can definitely update the downstream app's target version to 31. I will update those.

buildSrc/src/main/kotlin/Project.kt
```diff
@@ -12,6 +12,11 @@ object AndroidVersions {
...
+    const val minSdkVersion = 23
+    const val targetSdkVersion = 31
+    const val compileSdkVersion = 31
```
alexander-kulikovskii, line 18: I guess nice to tests with the latest version.
```suggestion
    const val targetSdkVersion = 33
    const val compileSdkVersion = 33
```

gradle/script-git-version.gradle
```diff
@@ -11,7 +11,7 @@ import org.ajoberstar.grgit.Grgit
 
 ext {
     git = Grgit.open(currentDir: projectDir)
-    def tagSize = git.tag.list().findAll { (it.name =~ "android-v*" || it.name =~ "v1\\d.*") }.size()
+    def tagSize = git.tag.list().findAll { (it.name =~ "android-v*" || it.name =~ "v1\\d.*") }.size() + 1
```
alexander-kulikovskii, line 14: why did you add `+1`?

yunikkk:
@ank27 since Play Store no longer accepts apps with targetSdk < 31 - should be migrate all of our apps? Would be helpful to users to see them with the correct and fresh target.

buildSrc/src/main/kotlin/Project.kt
yunikkk: Not sure I understand - why do we want to keep outdated target for the SDK? I assume we might need to update some parts of the SDK as well to correctly handle and adopt all of the requirements of the newer target. 
https://developer.android.com/google/play/requirements/target-sdk from the list of requirements it seems that we need to make sure `location permission` and `intent filters` sections are correctly implemented in the SDK, otherwise users might have problems I assume.

ank27: > I assume we might need to update some parts of the SDK as well to correctly handle and adopt all of the requirements of the newer target.

yeah, I have this concern only and upgrading target of sdk is not in scope of this pr. 
created [ticket](https://mapbox.atlassian.net/browse/MAPSAND-651) to track this

gradle/script-git-version.gradle
ank27: we provided an extra build to rave team manually. this was added to remove the error regarding versionCode should be greater than older versionCode in playstore

alexander-kulikovskii: ah. got it
maybe you should add extra comment here

alexander-kulikovskii (approved):
LGTM with small question

PR #1861, opened 2022-11-23
[viewannotation] optimize cameraForAnnotations api

ank27:
## Pull request checklist:
    

sdk/src/test/java/com/mapbox/maps/shadows/ShadowCoordinateBounds.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * To avoid calling native method of CoordinateBounds, this shadow class
+ * will be used for the Robolectric unit tests.
+ */
+@Implements(CoordinateBounds.class)
+public class ShadowCoordinateBounds {
+    @Implementation
+    public static double north() {
+        return 0.0;
```
kiryldz, line 15: should we actually shadow some valid coordinate bounds? not zero ones

kiryldz: and test some actual use-case

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,94 @@ internal class ViewAnnotationManagerImpl(
...
+    var west: Pair<ViewAnnotationOptions, Rect?>? = null
+    var south: Pair<ViewAnnotationOptions, Rect?>? = null
+
+    while (!isCorrectBound) {
+      val zoom = cameraOptionForCoordinates.zoom
+      isCorrectBound = true
+      viewAnnotationOptions.forEach { options ->
+        val frame = getViewAnnotationOptionsFrame(options)
+        val annotationBounds = calculateCoordinateBoundForAnnotation(options, zoom)
+        if (north == null || calculateCoordinateBoundForAnnotation(north!!.first, zoom).north() < annotationBounds.north()) {
```
kiryldz, line 218: maybe get rid of `!!` similar to how you do it in L275? we try to avoid `!!` in SDK code

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -25,9 +25,9 @@ internal class ViewAnnotationManagerImpl(
   mapView: MapView,
   private val viewAnnotationsLayout: FrameLayout = FrameLayout(mapView.context),
 ) : ViewAnnotationManager, ViewAnnotationPositionsUpdateListener {
-
   private val mapboxMap: MapboxMap = mapView.getMapboxMap()
   private val renderThread = mapView.mapController.renderer.renderThread
+  private var pixelRatio = mapView.resources.displayMetrics.density
```
alexander-kulikovskii, line 30:
```suggestion
  private val pixelRatio = mapView.resources.displayMetrics.density
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,94 @@ internal class ViewAnnotationManagerImpl(
...
+          isCorrectBound = false
+        }
+        if (west == null || calculateCoordinateBoundForAnnotation(west!!.first, zoom).west() > annotationBounds.west()) {
+          west = Pair(options, frame)
+          isCorrectBound = false
+        }
+      }
+
+      val coordinateBoundsForCamera = CoordinateBounds(
+        Point.fromLngLat((west!!.first.geometry as Point).longitude(), (south!!.first.geometry as Point).latitude()),
```
alexander-kulikovskii, line 237:
```suggestion
        Point.fromLngLat(west.longitude(), south.latitude()),
```

```
fun Pair<ViewAnnotationOptions, Rect?>?.longitude() = first.geometry as Point.longitude()
```

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,94 @@ internal class ViewAnnotationManagerImpl(
...
+   * Function to calculate coordinatebound associated with annotation option.
+   * this uses [MapboxMap.projectedMetersForCoordinate] and [MapboxMap.coordinateForProjectedMeters]
+   * function to get correct coordinate for northEast and southWest point of annotations.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun calculateCoordinateBoundForAnnotation(
+    viewAnnotationOptions: ViewAnnotationOptions,
+    zoom: Double?
+  ): CoordinateBounds {
+    val frame = getViewAnnotationOptionsFrame(viewAnnotationOptions)
```
alexander-kulikovskii, line 268: can we create some default frame is can't get with `viewAnnotationOptions`
it will simplify lines 274-278

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,94 @@ internal class ViewAnnotationManagerImpl(
...
     )
+
+    var isCorrectBound = false
+    // keep north, east, west and southmost Annotation and its frame to adjust paddings for zoom.
+    var north: Pair<ViewAnnotationOptions, Rect?>? = null
+    var east: Pair<ViewAnnotationOptions, Rect?>? = null
+    var west: Pair<ViewAnnotationOptions, Rect?>? = null
+    var south: Pair<ViewAnnotationOptions, Rect?>? = null
+
+    while (!isCorrectBound) {
```
alexander-kulikovskii, line 212: how many iteration can be here in worst case? can it be forever loop?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: I think here its fine, because we are already checking for null in the first option of `or` logic. if its null, then it won't execute next option of `or` logic. 
although i have added another check below for nullability. (that code shouldn't trigger in normal case because we have already checking for all the nullability conditions before, but good to include extra check).

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: Can't be a forever loop, the worst case iteration is 2 loops. 
Initially, we are getting innerBounds `cameraOption` before while loop, then the next iteration is checking if it needs extra paddings. All other nullability checks have already been tested.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,102 @@ internal class ViewAnnotationManagerImpl(
...
+      val zoom = cameraOptionForCoordinates.zoom
+      isCorrectBound = true
+      viewAnnotationOptions.forEach { options ->
+        val frame = getViewAnnotationOptionsFrame(options)
+        val annotationBounds = calculateCoordinateBoundForAnnotation(options, zoom)
```
yunikkk, line 217: Nit : you can pass calculated `frame` to `calculateCoordinateBoundForAnnotation` - it's calculated there second time

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,102 @@ internal class ViewAnnotationManagerImpl(
...
+   * function to get correct coordinate for northEast and southWest point of annotations.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun calculateCoordinateBoundForAnnotation(
+    viewAnnotationOptions: ViewAnnotationOptions,
+    zoom: Double?
+  ): CoordinateBounds {
+    val frame = getViewAnnotationOptionsFrame(viewAnnotationOptions) ?: Rect(0, 0, 0, 0)
+    val metersPerPixelAtLatitude =
+      if (zoom == null) mapboxMap.getMetersPerPixelAtLatitude((viewAnnotationOptions.geometry as Point).latitude()) else
```
yunikkk, line 278: Can it actually be null? What if it's called with the `null` value - will it use current zoom level of the map hence get incorrect `MetersPerPixelAtLatitude` value?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,102 @@ internal class ViewAnnotationManagerImpl(
...
+      val coordinateBoundsForCamera = CoordinateBounds(
+        Point.fromLngLat(west!!.coordinate().longitude(), south!!.coordinate().latitude()),
+        Point.fromLngLat(east!!.coordinate().longitude(), north!!.coordinate().latitude())
+      )
+
+      val paddings = EdgeInsets(
+        (edgeInsets?.top ?: 0).toDouble().plus(abs((north!!.second?.top ?: 0.0).toDouble())),
+        (edgeInsets?.left ?: 0).toDouble().plus(abs((west!!.second?.left ?: 0.0).toDouble())),
+        (edgeInsets?.bottom ?: 0).toDouble().plus(abs((south!!.second?.bottom ?: 0.0).toDouble())),
+        (edgeInsets?.right ?: 0).toDouble().plus(abs((east!!.second?.right ?: 0.0).toDouble()))
```
yunikkk, line 251: Reads hardly, maybe 
```suggestion
        (edgeInsets?.right ?: 0).toDouble() + abs((east!!.second?.right ?: 0.0).toDouble())
```
?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,102 @@ internal class ViewAnnotationManagerImpl(
...
+        coordinateBoundsForCamera,
+        paddings,
+        bearing,
+        pitch
+      )
+    }
+    return cameraOptionForCoordinates
+  }
+
+  private fun Pair<ViewAnnotationOptions, Rect?>.coordinate() = (first.geometry as Point)
```
yunikkk, line 264: Imho would be more useful to extend `ViewAnnotationOptions.` - there are more places with the cast `as Point`

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: in our implementation, it can't be `null` (we have checks before we calculate cameraOptions). but this function can be used independently to get the `coordinateBounds` by viewAnnotation. If the zoom is not provided it will use current map zoom to get the coordinatebound. So it's upto the calling function how they want to use it.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
evil159: We discovered a case where this can loop forever on iOS - with certain combination of camera state/map size/view annotations there might be one pair(view annotation options & rect) to be e.g. southern-most for a particular zoom level, and for another zoom level the different pair can become southern-most.

Potentially this code is also susceptible to the same scenario, worthwhile to double check. @maios is leading the investigation of this on iOS.

yunikkk: I'd probably removed the loop, extracted the code calculating `cameraOptionForCoordinates` and explicitly called it twice to avoid any risks.

ank27: Yeah, make sense. although I couldn't reproduce the issue on android with the same setup. I will try to loop it with a counter and that would be our maximum tries to adjust the bounds. 
I can see that this might not provide correct bounds for all cases, but we have discussed this with gl-native to improve this in next releases.

ank27: added [ticket](https://mapbox.atlassian.net/browse/MAPSAND-649) to track this issue.

yunikkk: @ank27 at the same time why do we need a loop for two iterations? What about extracting to function and calling twice? Code will be more clear and not additional variables involved

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,105 @@ internal class ViewAnnotationManagerImpl(
...
+        pitch
+      )
+    }
+    return cameraOptionForCoordinates
+  }
+
+  private fun ViewAnnotationOptions.coordinate() = geometry as Point
+
+  /**
+   * Function to calculate coordinatebound associated with annotation option.
```
yunikkk, line 270:
```suggestion
   * Function to calculate [CoordinateBounds] associated with annotation option.
```

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+## Bug fixes 🐞
+* Fix a bug where `ViewAnnotationManager.annotations` API doesn't return correct camera bounds. ([1861](https://github.com/mapbox/mapbox-maps-android/pull/1861))
```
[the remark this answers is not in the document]
ank27: yes. I can add that to v10.10 changelog when creating port pr .

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: In case we need to have more iteration in the future to have better bounds. but that's debatable. 
btw why can't we use loop though?

yunikkk: @ank27 I was thinking that the loop is not needed if we know that we have fixed and limited number (== 2) of iterations to calculate the bounds. Now it seems that it's not - and I'm actually ok with having the loop with counter which can be controlled. The question is do we want it to be 2 or some larger value to be almost certain that the bounds will be correctly calculated.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,105 @@ internal class ViewAnnotationManagerImpl(
...
     )
+
+    var isCorrectBound = false
+    // keep north, east, west and southmost Annotation and its frame to adjust paddings for zoom.
+    var north: Pair<ViewAnnotationOptions, Rect?>? = null
+    var east: Pair<ViewAnnotationOptions, Rect?>? = null
+    var west: Pair<ViewAnnotationOptions, Rect?>? = null
+    var south: Pair<ViewAnnotationOptions, Rect?>? = null
+
+    // we run the loop twice to adjust bounds correctly to fit all the annotations.
```
yunikkk, line 212: Nit - reword to hint that there is still no guarantee that bounds will fit all the annotations?

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,105 @@ internal class ViewAnnotationManagerImpl(
...
+    var boundsCounter = 1
+    while (!isCorrectBound && boundsCounter <= MAX_ADJUST_BOUNDS_COUNTER) {
+      val zoom = cameraOptionForCoordinates.zoom
+      boundsCounter++
+      isCorrectBound = true
+      viewAnnotationOptions.forEach { options ->
+        val frame = getViewAnnotationOptionsFrame(options) ?: Rect(0, 0, 0, 0)
```
yunikkk, line 219: nit : make `getViewAnnotationOptionsFrame` return nun-nullable value? Options are non-null

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
```diff
@@ -178,13 +195,105 @@ internal class ViewAnnotationManagerImpl(
...
+    // keep north, east, west and southmost Annotation and its frame to adjust paddings for zoom.
+    var north: Pair<ViewAnnotationOptions, Rect?>? = null
+    var east: Pair<ViewAnnotationOptions, Rect?>? = null
+    var west: Pair<ViewAnnotationOptions, Rect?>? = null
+    var south: Pair<ViewAnnotationOptions, Rect?>? = null
+
+    // we run the loop twice to adjust bounds correctly to fit all the annotations.
+    var boundsCounter = 1
+    while (!isCorrectBound && boundsCounter <= MAX_ADJUST_BOUNDS_COUNTER) {
+      val zoom = cameraOptionForCoordinates.zoom
```
yunikkk, line 215: nit : declare val out of loop above

yunikkk (approved):

kiryldz:
@ank27 can you please rebase it? seems to be pretty outdated

CHANGELOG.md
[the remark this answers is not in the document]
kiryldz: @ank27 issue with your approach is that in that case changlelog will be correct in `v10.10` branch but it will be incorrect in `main` branch and this record might be duplicated in `v10.11-beta.1` then. 
I would advise to rebase now and add it to the correct section.

sdk/src/test/java/com/mapbox/maps/ViewAnnotationManagerAddTest.kt
```diff
@@ -54,7 +55,8 @@ class ViewAnnotationManagerAddTest(
     every { viewAnnotationsLayout.addView(any()) } just Runs
     every { viewAnnotationsLayout.removeView(any()) } just Runs
     every { viewAnnotationsLayout.context } returns mockk()
-
+    val displayMetrics = DisplayMetrics().also { it.density = 1f }
```
kiryldz, line 58: nit-picking:
```suggestion
    val displayMetrics = DisplayMetrics().apply { density = 1f }
```

CHANGELOG.md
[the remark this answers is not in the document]
ank27: sure, will do :)

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: ohh wait, it's used inside while loop only, so I will keep it here.

sdk/src/main/java/com/mapbox/maps/ViewAnnotationManagerImpl.kt
ank27: 2 iteration i found good for the cases i have tested, but i can't be sure of all edge-cases (for eg. pitched map and with bearing). so i am not 100% sure about fixed number. 

> The question is do we want it to be 2 or some larger value to be almost certain that the bounds will be correctly calculated.

cut a [ticket](https://mapbox.atlassian.net/browse/MAPSAND-649) to look more into finding correct calculations for all cases.

PR #1876, opened 2022-11-29
Improve performance by setting geojson data directly

kiryldz:
## Pull request checklist:
    
### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
@@ -1,29 +1,28 @@
 package com.mapbox.maps.testapp.examples
 
+import android.graphics.BitmapFactory
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
-import androidx.core.view.postDelayed
 import com.mapbox.bindgen.Value
 import com.mapbox.common.SettingsServiceFactory
 import com.mapbox.common.SettingsServiceStorageType
 import com.mapbox.geojson.Feature
-import com.mapbox.geojson.LineString
 import com.mapbox.geojson.Point
 import com.mapbox.maps.CameraOptions
 import com.mapbox.maps.MapView
 import com.mapbox.maps.Style
-import com.mapbox.maps.extension.observable.model.SourceDataType
+import com.mapbox.maps.extension.style.image.image
 import com.mapbox.maps.extension.style.layers.addLayer
 import com.mapbox.maps.extension.style.layers.generated.lineLayer
+import com.mapbox.maps.extension.style.layers.generated.symbolLayer
+import com.mapbox.maps.extension.style.layers.properties.generated.IconAnchor
 import com.mapbox.maps.extension.style.sources.addSource
 import com.mapbox.maps.extension.style.sources.generated.geoJsonSource
 import com.mapbox.maps.extension.style.style
-import com.mapbox.maps.logE
 import com.mapbox.maps.plugin.animation.MapAnimationOptions
 import com.mapbox.maps.plugin.animation.flyTo
-import com.mapbox.maps.plugin.delegates.listeners.OnSourceDataLoadedListener
+import com.mapbox.maps.testapp.R
 import com.mapbox.maps.testapp.examples.annotation.AnnotationUtils
-import java.util.*
 
 /**
  * Example of passing large geojson to verify it does not block UI thread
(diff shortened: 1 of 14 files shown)


sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -963,6 +963,22 @@ class Style internal constructor(
     return styleManager.setStyleSourceProperties(sourceId, properties)
   }
 
+  override fun setStyleSourceData(sourceId: String, data: Value): Expected<String, None> {
+    // could be called from any thread
```
paulrenn67, line 967: There is [AnyThreaf](https://developer.android.com/reference/kotlin/androidx/annotation/AnyThread) annotation, should we consider marking our APIs accordingly?

kiryldz: yeah, perhaps - but when doing so it makes sense guess to update all other methods with `@UiThread` which is obviously not in scope of this PR. Will give it a thought, thanks 😄

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -618,30 +699,53 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
   /**
    * Static variables and methods.
    */
   companion object {
+    private const val TAG = "GeoJsonSource"
+
+    private fun directSetterEnabled(): Boolean {
+      val settingValue = SettingsServiceFactory
+        .getInstance(SettingsServiceStorageType.NON_PERSISTENT)
+        .get("geojson_allow_direct_setter")
```
alexander-kulikovskii, line 755: if it possible it will be nice to save to some field and don't recalculate it each time

kiryldz: fixed

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -607,9 +694,13 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
      * @param value the feature collection
      */
     fun featureCollection(value: FeatureCollection) = apply {
-      rawGeoJson = value
-      val propertyValue = PropertyValue("data", "null")
-      properties[propertyValue.propertyName] = propertyValue
+      geoJson = value
+      if (directSetterEnabled) {
```
alexander-kulikovskii, line 698: code from 698-703 copied 3 times:
`feature`, `featureCollection`, `geometry`

alexander-kulikovskii (approved):

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -35,25 +40,83 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
         }
       }
     }
   }
 
+  private fun setGeoJSON(geoJson: GeoJson) {
+    workerHandler.removeCallbacksAndMessages(null)
+    workerHandler.post {
+      if (delegate != null) {
+        delegate?.setStyleSourceData(sourceId, toGeoJsonData(geoJson))
```
yunikkk, line 74: Is it really ok to post on worker thread? What if map is destroyed when it's launched?

kiryldz: discussed offline, should be ok

extension-style/api/extension-style.api
```diff
@@ -3667,7 +3667,8 @@ public final class com/mapbox/maps/extension/style/sources/generated/Encoding :
 public final class com/mapbox/maps/extension/style/sources/generated/GeoJsonSource : com/mapbox/maps/extension/style/sources/Source {
 	public static final field Companion Lcom/mapbox/maps/extension/style/sources/generated/GeoJsonSource$Companion;
 	public fun <init> (Lcom/mapbox/maps/extension/style/sources/generated/GeoJsonSource$Builder;)V
-	public synthetic fun <init> (Lcom/mapbox/maps/extension/style/sources/generated/GeoJsonSource$Builder;Lcom/mapbox/geojson/GeoJson;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (Lcom/mapbox/maps/extension/style/sources/generated/GeoJsonSource$Builder;Lcom/mapbox/geojson/GeoJson;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
```
kiryldz, line 3670: I'm not sure why we have this line updated as I added parameter to private constructor

extension-style/api/extension-style.api
```diff
@@ -3724,6 +3725,7 @@ public final class com/mapbox/maps/extension/style/sources/generated/GeoJsonSour
 }
 
 public final class com/mapbox/maps/extension/style/sources/generated/GeoJsonSource$Companion {
+	public final fun directSetterEnabled ()Z
```
kiryldz, line 3728: need to make this static method public to be visible for unit tests in annotation plugin

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -618,30 +692,61 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+    private const val TAG = "GeoJsonSource"
+
+    /**
+     * Check if runtime property geojson_allow_direct_setter is enabled.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun directSetterEnabled(): Boolean {
+      val settingValue = SettingsServiceFactory
+        .getInstance(SettingsServiceStorageType.NON_PERSISTENT)
+        .get("geojson_allow_direct_setter")
```
jush, line 746: Should string `"geojson_allow_direct_setter"` be a constant so users can use it directly instead of hardcoded string?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -35,25 +40,83 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+            setProperty(property, throwRuntimeException = false)
+          }
         }
       }
     }
   }
 
+  private fun setGeoJSON(geoJson: GeoJson) {
+    workerHandler.removeCallbacksAndMessages(null)
+    workerHandler.post {
```
jush, line 72: Do we really need to jump to the worker thread or can it be run from any thread?(even main?)

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: That's not really something we want to expose (that's the reason for marking this PR as skip changelog) so exposing constant imo is not needed here.

kiryldz: As per plan using direct setter will be introduced fully in v10.11-beta.1

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
kiryldz: that's the purpose of the fix - calling native method is thread-safe and it does still perform some basic marshalling (e.g. in our example it takes around 100 ms) - so we're performing this on worker thread in order not to block the main thread.

jush: Thanks for the explanation!

app/src/main/java/com/mapbox/maps/testapp/examples/LargeGeojsonPerformanceActivity.kt
```diff
@@ -34,6 +37,10 @@ class LargeGeojsonPerformanceActivity : AppCompatActivity() {
     val mapView = MapView(this)
     setContentView(mapView)
 
+    SettingsServiceFactory
```
yunikkk, line 40: Remove from the public demo?

extension-style/api/extension-style.api
yunikkk: Just discussed offline - turns out it's generated by Kotlin to allow to the private constructor to be called from Builder, which otherwise wouldn't be possible.
Similar question discussed : https://stackoverflow.com/questions/54613891/why-private-constructor-of-sealed-class-can-be-called-in-sub-class

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -35,25 +40,83 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
+      if (delegate != null) {
+        delegate?.setStyleSourceData(sourceId, toGeoJsonData(geoJson))
+        initGeoJSON = null
```
yunikkk, line 75: IMO it's confusing that class var is reset here, while it's not used and geoJson is passed as an argument. I'd move the = null to the place where initGeoJSON is passed

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -618,30 +692,61 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
      * @param value the geometry
      */
     fun geometry(value: Geometry) = apply {
-      rawGeoJson = value
-      val propertyValue = PropertyValue("data", "null")
-      properties[propertyValue.propertyName] = propertyValue
+      setGeoJson(value)
+    }
+
+    private fun setGeoJson(geoJson: GeoJson) {
```
yunikkk, line 698:
```suggestion
    private fun geoJson(geoJson: GeoJson) {
```
since it's builder

sdk/src/main/java/com/mapbox/maps/Style.kt
```diff
@@ -379,6 +381,31 @@ class Style internal constructor(
     return styleManager.setStyleLightProperty(id, light)
   }
 
+  /**
```
yunikkk, line 384: Reminder to remove

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -618,30 +692,61 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
...
 
     /**
      * Build the GeoJsonSource.
      *
      * @return the GeoJsonSource
      */
-    fun build() = GeoJsonSource(this, rawGeoJson)
+    fun build(): GeoJsonSource {
+      if (directSetterEnabled) {
+        // we need empty initial data when creating the geojson (as we don't set it in constructor) -
```
yunikkk, line 724:
```suggestion
        // set default data to allow empty data source.
```
the same comment is added to the builder

yunikkk (approved):

ank27 (approved):

buildSrc/src/main/kotlin/Project.kt
```diff
@@ -108,7 +108,7 @@ object Versions {
   const val mapboxGestures = "0.8.0"
   const val mapboxJavaServices = "5.4.1"
   const val mapboxBase = "0.8.0"
-  const val mapboxGlNative = "10.10.0-rc.1"
+  const val mapboxGlNative = "10.10.0-rc.1-dd749b21a7-SNAPSHOT"
```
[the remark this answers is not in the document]
kiryldz: it's OK to leave before merge but we just need to be careful when rebasing to `v10.10` branch. Thanks for the reminder!

OdNairy (approved):

PR #1878, opened 2022-11-30
[MAPSAND-630] Include gl-native license

ank27:
### Summary of changes
This pr include gl-native license to `License.md` and move license validation checks to internal repo.

## Pull request checklist:
    

LICENSE.md
```diff
@@ -400,3 +400,1477 @@ License: [The Apache Software License, Version 2.0](http://www.apache.org/licens
 
 ===========================================================================
 
+### [MapboxCoreMaps](https://www.mapbox.com/) by Mapbox
```
pengdev, line 403: the format seems a bit different from the above, if the license is just Apache or BSD or Mapbox ToS, shall we just add the url as above, instead of pasting the whole license?

.circleci/config.yml
```diff
@@ -461,9 +461,6 @@ jobs:
                 command: |
                   echo "PR Target branch: $PR_TARGET_BRANCH"
                   ./gradlew sdk:dependencyInsight --configuration releaseRuntimeClasspath --dependency com.mapbox.maps | grep -vzq 'FAILED'
-            - run:
```
alexander-kulikovskii, line 464: so, right now internal `build from source` should be required?
or maybe add extra upstream job that triggers only license check?

ank27: I don't think its a `required` job, we will keep it sync, and just validate license file on everytime we . 
I guess no need to add extra license check, it will be triggered everytime we create a commit / merge to main (Similar to codegen-verify).

alexander-kulikovskii: yeah, but if somethings change in public PR, you just run license check as a part of `build from source` and it's not required job. Potentially you can merge it

ank27: I thought about that to include in `build-from-source`, but as the name suggest it builds everything from source, and it doesn't look correct if we verify licenses in that job. 

> and it's not required job. Potentially you can merge it

Yes, potentially we should be able to merge imo. If there is some issue with the license file or let's say license-plugin, it shouldn't stop our development flow.

kiryldz (approved):

kiryldz:
@ank27 should this be rebased on `v10.10` or can we wait until `v10.11-beta.1`

ank27:
> @ank27 should this be rebased on `v10.10` or can we wait until `v10.11-beta.1`

I guess we can include it in v10.10

.circleci/config.yml
pengdev: I'd trigger a gl-native bot to validate the license on every PR, in case we accidentally introduce a new dependency while developing the public sdk.

ank27: > accidentally introduce a new dependency

we don't generate new licenses from c++ code, we just include those from gl-native. so i don't think we can accidentally introduce new dependency from our side. 
one case i can think is if certain licenses are included in `internal` of gl-native but not present in release branch. I think gl-native should keep the internal licenses upto date with these inclusion.

pengdev: I mean the dependency we introduce from Android SDK, not related to gl-native.

pengdev (approved):

PR #1909, opened 2022-12-12
Skip camera updates in Animation if it's already applied

kiryldz:
### Summary of changes
Add camera events pre-processing to:
 - do not run camera animation logic for animators where it makes no sense (start = end = current camera) - which could happen pretty often in Nav SDK / viewport use-case operating with low-level animators
 - do not call native `setCamera` if desired camera is already applied
 - fix an issue for `CameraAnimatorChangeListener<T>` listeners not reporting the change if camera has been changed not directly through the plugin (e.g. viewport size changed, direct `setCamera` call happened)

All that should:
 - reduce number of native `setCamera` calls
 - eliminate duplicates in all the plugins that implement `MapCameraPlugin`
 - eliminate useless invocations of `CameraAnimationsLifecycleListener` when animators were running that actually did not truly change the camera

Additionally changed `debugMode` to use INFO level as DEBUG one seems to not show in release.

## Pull request checklist:
    

yunikkk:
LGTM, let's wait for test to pass or fail.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -237,10 +288,38 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+    }
+    for (i in 0 until targets.size - 1) {
+      if (targets[i] != targets[i + 1]) {
+        return false
+      }
+    }
+    return when (type) {
+      CameraAnimatorType.CENTER -> Objects.equals(currentCameraState?.center, targets[0])
+      CameraAnimatorType.ZOOM -> Objects.equals(currentCameraState?.zoom, targets[0])
+      CameraAnimatorType.ANCHOR -> false
```
pengdev, line 316: could you add some comments why we returns false for ANCHOR here(skip anchor) but in `private fun skipNativeSetCamera(cameraOptions: CameraOptions): Boolean` we returned true if ANCHOR exists?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -152,11 +154,22 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+      notifyListeners(coreCameraState)
+    }
+  }
+
   /**
    * Provides all map delegate instances.
    */
   override fun onDelegateProvider(delegateProvider: MapDelegateProvider) {
     mapDelegateProvider = delegateProvider
+    mapDelegateProvider.mapListenerDelegate.addOnCameraChangeListener(cameraChangeListener)
```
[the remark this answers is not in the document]
yunikkk: Indeed, changing it as you've suggested.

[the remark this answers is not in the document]
kiryldz: That technically changes public API, not sure we want this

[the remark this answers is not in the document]
yunikkk: Should be fine since we only modify CameraAnimationsPluginImpl and only add new methods, not remove them.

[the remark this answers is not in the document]
yunikkk: @pengdev I've updated PR, PTAL

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -181,17 +206,55 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+      return false
+    }
+    if (cameraOptions.center != null && cameraOptions.center != center) {
+      return false
+    }
+    if (cameraOptions.padding != null && cameraOptions.padding != padding) {
+      return false
+    }
+    return true
+  }
```
pengdev, line 240: please add unit tests to cover these logic

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
pengdev: shall we provide the context in the inline comment?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -163,6 +163,31 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+    // Insets array order : insets.left, insets.top, insets.right, insets.bottom
+    this.padding = EdgeInsets(
+      /* top = */
+      padding[1],
+      /* left = */
+      padding[0],
+      /* bottom = */
+      padding[3],
+      /* right = */
+      padding[2],
```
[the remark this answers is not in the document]
yunikkk: It's how they're placed in the array, https://github.com/mapbox/mapbox-maps-android/blob/main/sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt#L160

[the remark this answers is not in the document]
pengdev: yes, I mean why we have the array in that order 🙈 , and probably it's easier to reuse the EdgeInsects class instead of the array there, as it's more clear what the fields are.

pengdev:
Overall LGTM with nits, could we compare the number of camera updates before/after this change in the same scenario?

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImplTest.kt
```diff
@@ -1006,31 +1018,30 @@ class CameraAnimationsPluginImplTest {
...
     cameraAnimationsPluginImpl.easeTo(cameraState.toCameraOptions(), mapAnimationOptions { duration(DURATION) })
     shadowOf(getMainLooper()).idle()
-    verify(exactly = 0) { logD(TAG, any()) }
+    verify(exactly = 0) { logI(TAG, any()) }
   }
 
   @Test
   fun catchExceptionOnSetCameraTest() {
-    executeSetCameraTest(Exception("Invalid camera options"))
+//    executeSetCameraTest(Exception("Invalid camera options"))
```
pengdev, line 1026: why do we comment out this test? if not needed, maybe use `@ignore` annotation with some context would be better.

yunikkk: Reverted, goot catch

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: Will do

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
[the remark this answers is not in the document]
yunikkk: Well it's part of the public interface so will be problematic to replace array with the Insets. However, the only children of MapCameraPlugin are CompassViewPlugin and this new CameraAnimationsPluginImpl. CompassViewPlugin does not make use of paddings at all, it just uses bearing, https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt#L180, so I'll change the order of this array elements to correspond to EdgeInsets.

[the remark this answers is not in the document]
yunikkk: Updated

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: Added

yunikkk:
> Overall LGTM with nits, could we compare the number of camera updates before/after this change in the same scenario?

For example, before the optimization FollowPuckViewportState triggers 5 updates of camera per each animation tick - for each of the available properties of CameraOptions. With the fix it's reduced to 1 update.

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
```diff
@@ -157,7 +157,7 @@ internal class MapPluginRegistry(
         cameraState.pitch,
         cameraState.bearing,
         cameraState.padding.let { insets ->
-          arrayOf(insets.left, insets.top, insets.right, insets.bottom)
+          arrayOf(insets.top, insets.left, insets.bottom, insets.right)
```
pengdev, line 160: We should update the documentation to be specific as well, this might affect other codebase as well, e.g weather cc @paulrenn67 
consider the potential impact to users who implement their own MapCameraPlugin, I'd vote for keeping the original order.

pengdev: Maybe we should just keep the original order and add specific documentation on the meaning of each element in the list?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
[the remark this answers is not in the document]
pengdev: Added more suggestions in https://github.com/mapbox/mapbox-maps-android/pull/1909#discussion_r1050563928

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
pengdev: emm. strange, I remember 
```
    if (cameraOptions.anchor != null) {
      return false
    }
```
was returning true before, never mind.

sdk/src/main/java/com/mapbox/maps/plugin/MapPluginRegistry.kt
yunikkk: Yes, this is indeed the behavioural change. Agree it's better to keep the original order, will add the docs and revert the change.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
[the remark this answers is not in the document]
yunikkk: Reverted, thanks for pushing

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -163,6 +163,42 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
+   * Could be invoked from any thread when map starts rendering.
+   *
+   * @param lat latitude
+   * @param lon longitude
+   * @param zoom zoom
+   * @param pitch pitch in degrees
+   * @param bearing bearing in degrees
+   * @param padding padding ordered as [left, top, right, bottom]
+   */
+  override fun onCameraMove(
```
paulrenn67, line 177: Question: since the comment indicates this could be called on any thread, does it mean we can be called from multiple threads concurrently? If so, then the internal state could be corrupted since we are saving values to `this`

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -237,17 +308,35 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
...
 
-  internal fun notifyListeners(cameraState: CameraState) {
-    cameraState.also {
-      bearing = it.bearing
-      center = it.center
-      padding = it.padding
-      pitch = it.pitch
-      zoom = it.zoom
+  private fun skipRedundantAnimator(targets: Array<out Any?>, type: CameraAnimatorType): Boolean {
+    if (targets.size <= 1) {
```
jush, line 325: Wouldn't it be faster to check if `cameraAnimator.targets.size` is 0 at the start of fun `updateAnimatorValues` instead? we would avoid reaching this point

jush: Moreover, we wouldn't need the below `return when (type) {...}`, would we?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
```diff
@@ -286,6 +375,13 @@ class CameraAnimationsPluginImpl : CameraAnimationsPlugin {
             if (startingAnimator.canceled) {
               return
             }
+            // Prepare animator values
+            // Some animators might not have initial values and should be skipped from internal update
+            val animatorStartRequired = updateAnimatorValues(startingAnimator)
+            if (!animatorStartRequired) {
+              startingAnimator.skipped = true
```
jush, line 382: With this change the `lifecycleListeners` wont' be notified about `onAnimatorStarting` and neither of `onAnimatorEnding` (see [L436](https://github.com/mapbox/mapbox-maps-android/pull/1909/files#diff-4afad8e84850da48f89b60130a335fe9d19ce152334ab24201f1eacd792ee4f4R436)).

Are we okay with this behaviour change?

jush: I am getting a bit lost with the code so not sure what `internal update` means here... So maybe it's okay to skip the notifications I mentioned above... 😟❓

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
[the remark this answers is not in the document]
jush: FWIW, it seems to follow the same order as in https://developer.android.com/reference/android/graphics/Rect#Rect(int,%20int,%20int,%20int)

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: Indeed. The doc about this states that it might happen when `setCamera` is called directly after `mapView.onStart`, https://github.com/mapbox/mapbox-maps-android/issues/308. 
Not sure it's still possible since mapView was pushed to work only with UiThread since. Trying locally shows main thread is used even with the described calls sequence.

yunikkk: Suppose gl-native somehow has changed the implementation of observer, I'm seeing that `MapEvents.CAMERA_CHANGED` event is delivered on the MainThread even setting camera directly after `onStart.` 
@alexshalamov @pozdnyakov do you folks remember if events are _always_ delivered on the main thread? We could drop the unneeded thread switches then.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: Commentary is not updated from the intermediate PR state so it's confusing. I'll reword

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: I'd simplified code a bit however, moved check for the empty target values to the very top of the method, PTAL

jush (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
jush: Thanks for explaining!

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraAnimationsPluginImpl.kt
yunikkk: Talked offline with @pozdnyakov - indeed camera events originating in Observer API will arrive on Main thread. Checked it as well, works as expected.
Removed the comment about AnyThread here and in CompassViewPlugin as well.

PR #1912, opened 2022-12-14
[tests] patch flaky tests

tobrun:
- rely on map loaded events instead of map idle
- remove handler concept from tests and rely on dedicated event callbacks
- remove integration tests that depend on animation system, should use unit tests instead*
- increase timeout to 10 seconds when it's about loading a style

`*` =  the integration level animation tests we have in the SDK are more testing the Android Animator framework vs testing our own business logic. We should be investing into unit tests and not into instrumentation tests, the async behavior makes them flaky.


sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
```diff
@@ -391,19 +384,16 @@ class MapIntegrationTest {
...
-          mapView.postDelayed(
-            {
-              mapboxMap.queryRenderedFeatures(
-                RenderedQueryGeometry(ScreenCoordinate(mapView.width / 2.0, mapView.height / 2.0)),
-                RenderedQueryOptions(listOf("layer"), null)
-              ) { result ->
-                if (result.value?.size == 1) {
-                  countDownLatch.countDown()
-                }
+          mapboxMap.addOnMapLoadedListener {
```
alexander-kulikovskii, line 387: maybe reuse code inside listener?

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
```diff
@@ -372,7 +325,7 @@ class OfflineTest {
         // verify no map loading errors occurred
         Assert.assertEquals(0, mapLoadingErrorCount)
       } else {
-        throw TimeoutException()
+        logE(TAG, "Timeout occurred, silently ignoring due to flaky nature of this test")
```
alexander-kulikovskii, line 328: should we add todo?
maybe track somehow?

sdk/src/androidTest/java/com/mapbox/maps/MapIntegrationTest.kt
tobrun: test removed

sdk/src/androidTest/java/com/mapbox/maps/OfflineTest.kt
tobrun: test removed

tobrun:
Even with small patches, the tests remained unstable. I'm bluntly removing them as I don't actually believe they are adding value. The overhead and costs that come from them don't match the scope of what is being tested.

yunikkk (approved):

PR #1916, opened 2022-12-15
[ci] optimize device configuration for firebase testlab

tobrun:
- run PRs only against a single device on firebase testlab on a PR basis
- run against multiple devices on a weekly basis (cron job)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -384,8 +384,8 @@ commands:
         description: app target
         type: string
         default: "app"
-      nightly:
-        description: nightly build or normal
+      multiple_devices:
+        description: multiple_devices build or normal
         type: boolean
         default: false
     steps:
@@ -402,7 +402,7 @@ commands:
                   --timeout 45m --use-orchestrator --no-record-video --num-flaky-test-attempts 1 \
                   --num-uniform-shards 3
               else
-                if << parameters.nightly >> ; then
+                if << parameters.multiple_devices >> ; then
                   gcloud beta firebase test android run --type instrumentation \
                     --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
                     --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
@@ -425,8 +425,8 @@ commands:
 
   run-firebase-robo:
     parameters:
-      nightly:
-        description: nightly build or normal
+      multiple_devices:
+        description: multiple_devices build or normal
         type: boolean
         default: false
     steps:
@@ -435,7 +435,7 @@ commands:
           no_output_timeout: 10m
           command: |
             if [[ -n "${GCLOUD_SERVICE_ACCOUNT_JSON}" ]]; then
-              if << parameters.nightly >> ; then              
+              if << parameters.multiple_devices >> ; then
                 gcloud firebase test android run --type robo \
                   --app app/build/outputs/apk/release/app-release.apk \
                   --device model=Nexus5,version=21,locale=en,orientation=portrait \
@@ -448,7 +448,7 @@ commands:
                 gcloud firebase test android run --type robo \
                   --app app/build/outputs/apk/release/app-release.apk \
                   --device model=redfin,version=30,locale=en,orientation=portrait \
-                  --timeout 90s            
+                  --timeout 90s
               fi
             fi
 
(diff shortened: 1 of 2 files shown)


tobrun:
While the code is ready, there is some strange issue that names have -1 appended because they live in 2 workflows.

image

This doesn't scale well because we would need to update our required jobs:

image

.circleci/config.yml
```diff
@@ -363,36 +380,60 @@ commands:
...
-                --num-uniform-shards 4
+                if << parameters.nightly >> ; then
+                  gcloud beta firebase test android run --type instrumentation \
+                    --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
+                    --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
+                    --device model=harpia,version=23,locale=en,orientation=portrait \
+                    --device model=blueline,version=28,locale=en,orientation=portrait \ 
+                    --device model=redfin,version=30,locale=en,orientation=portrait \
+                    --device model=oriole,version=33,locale=en,orientation=portrait \
+                    --device model=HWCOR,version=27,locale=en,orientation=portrait \
```
alexander-kulikovskii, line 391: nit:
sort by API level

.circleci/config.yml
```diff
@@ -123,6 +112,30 @@ workflows:
...
+          filters:
+            branches:
+              only:
+                - main
+    jobs:
+      - check-gradle-dependencies
+      - run-robo-test:
+          nightly: true
+          requires:
+            - build-modules-and-instrumentation-tests
```
alexander-kulikovskii, line 129: nit:
maybe make it required from `run-sdk-instrumentation-test`?
to spend less resources in case of failure

.circleci/config.yml
```diff
@@ -123,6 +112,30 @@ workflows:
     jobs:
       - run-sdk-instrumentation-test
 
+  weekly:
+    triggers:
+      - schedule:
```
alexander-kulikovskii, line 117: what about reporting? should it slack notification?

alexander-kulikovskii:
one more question. 
what about manual run? after bumping gl native or common it would be nice to check on several API levels

.circleci/config.yml
tobrun: :+1: , will reverse it though, first run robo after run instrumentation tests

tobrun:
Following @alexander-kulikovskii review comments, it's now also possible to run the weekly CI integration add-hoc (for example when bumping any of the internal dependencies). 

```bash
python3 scripts/ci-weekly-trigger --branch foobar 
```

.circleci/config.yml
```diff
@@ -363,36 +402,54 @@ commands:
...
+                    --device model=redfin,version=30,locale=en,orientation=portrait \
+                    --timeout 45m --use-orchestrator --no-record-video --num-flaky-test-attempts 1 \
+                    --num-uniform-shards 4
+                fi
               fi
             fi
 
   run-firebase-robo:
+    parameters:
+      nightly:
```
jush, line 428: Nit: maybe a better name would be `multiple-devices` (or `single-device`). It doesn't really matter if it's nightly or not.

jush (approved):
LGTM, just minor comment on naming.

.circleci/config.yml
```diff
@@ -363,36 +402,54 @@ commands:
...
+                    --device model=blueline,version=28,locale=en,orientation=portrait \
+                    --device model=redfin,version=30,locale=en,orientation=portrait \
+                    --device model=oriole,version=33,locale=en,orientation=portrait \
+                    --timeout 45m --use-orchestrator --no-record-video --num-flaky-test-attempts 1 \
+                    --num-uniform-shards 4
+                else
+                  gcloud beta firebase test android run --type instrumentation \
+                    --app << parameters.app_target >>/build/outputs/apk/debug/<< parameters.app_target >>-debug.apk \
+                    --test << parameters.module_target >>/build/outputs/apk/androidTest/debug/<< parameters.module_target >>-debug-androidTest.apk \
+                    --device model=redfin,version=30,locale=en,orientation=portrait \
```
yunikkk, line 419: Thinking that it might be useful to launch PR tests on the latest available SDK version, wdyt?

yunikkk (approved):

.circleci/config.yml
tobrun: The redfin device is currently the "default" device which means it will have the highest availability and I'm hoping it's also the most stable. We will check the latest available Android version on a weekly basis and can use this system manually when we bump upstream dependencies

yunikkk: My idea was that typically the issues happen with the newest or oldest devices, using latest would allow to catch them as soon as the buggy code gets merged. But I guess we still should be fine with the weekly tests.

alexander-kulikovskii (approved):

PR #1936, opened 2023-01-11
MAPSAND-684 Run onAppUserTurnstileEvent in thread to speed up MapView initialization

paulrenn67:
### Summary of changes

Measurement of map initialization seems to show calls to onAppUserTurnstileEvent during map construction taking around 20 milliseconds on average on the main thread.

Putting the call to onAppUserTurnstileEvent in a thread shortens this to around ~1 millisecond, so any delays that might occur during the event sending (e.g. initial report said the startup took over 100 ms) will not affect the main thread. The event is not critical to use of the MapView, it simply enqueues the turnstile event to be sent to the backend.

There were no apparent side effects of sending this event from a coroutine, so should be safe to take into use.

User impact (optional)
Should result is slightly faster MapView initialization.

## Pull request checklist:
    
### DIFF
--- a/sdk/build.gradle.kts
+++ b/sdk/build.gradle.kts
@@ -73,6 +73,7 @@ dependencies {
   api(project(":plugin-viewport"))
   compileOnly(Dependencies.asyncInflater)
   implementation(Dependencies.kotlin)
+  implementation(Dependencies.coroutines)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
   testImplementation(Dependencies.junit)
--- a/sdk/build.gradle.kts
+++ b/sdk/build.gradle.kts
@@ -73,7 +73,6 @@ dependencies {
   api(project(":plugin-viewport"))
   compileOnly(Dependencies.asyncInflater)
   implementation(Dependencies.kotlin)
-  implementation(Dependencies.coroutines)
   implementation(Dependencies.androidxCoreKtx)
   implementation(Dependencies.androidxAnnotations)
   testImplementation(Dependencies.junit)
--- a/sdk/src/main/java/com/mapbox/maps/MapProvider.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapProvider.kt
@@ -11,6 +11,8 @@ import com.mapbox.maps.base.BuildConfig
 import com.mapbox.maps.module.MapTelemetry
 import com.mapbox.maps.plugin.MapDelegateProviderImpl
 import com.mapbox.maps.plugin.MapPluginRegistry
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 
 internal object MapProvider {
 
@@ -50,7 +52,9 @@ internal object MapProvider {
         paramsProvider(context, accessToken, MapboxModuleType.MapTelemetry)
       }
     }
-    mapTelemetry.onAppUserTurnstileEvent()
+    GlobalScope.launch {
+      mapTelemetry.onAppUserTurnstileEvent()
+    }
     return mapTelemetry
   }
 
--- a/sdk/src/main/java/com/mapbox/maps/MapProvider.kt
+++ b/sdk/src/main/java/com/mapbox/maps/MapProvider.kt
@@ -11,8 +11,8 @@ import com.mapbox.maps.base.BuildConfig
 import com.mapbox.maps.module.MapTelemetry
 import com.mapbox.maps.plugin.MapDelegateProviderImpl
 import com.mapbox.maps.plugin.MapPluginRegistry
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
+import kotlin.concurrent.thread
+
 
 internal object MapProvider {
 
@@ -52,7 +52,8 @@ internal object MapProvider {
         paramsProvider(context, accessToken, MapboxModuleType.MapTelemetry)
       }
     }
-    GlobalScope.launch {
+    // Report turnstile event in background thread.
+    thread {
       mapTelemetry.onAppUserTurnstileEvent()
     }
     return mapTelemetry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+* Improvement for MapView initialization time. ([1936](https://github.com/mapbox/mapbox-maps-android/pull/1936))
 
 # 10.11.0-beta.1 January 11, 2023
 ## Features ✨ and improvements 🏁


sdk/build.gradle.kts
```diff
@@ -73,6 +73,7 @@ dependencies {
   api(project(":plugin-viewport"))
   compileOnly(Dependencies.asyncInflater)
   implementation(Dependencies.kotlin)
+  implementation(Dependencies.coroutines)
```
alexander-kulikovskii, line 76: do you have data on how much this increases our sdk size?

paulrenn67: Good question, I will try to measure it.  But I guess we should not be afraid of using language features.

yunikkk: What about method count? I'm also in doubt whether we need to add dependency on coroutines for one simple call

alexander-kulikovskii: btw, it's base module. If user use our sdk only with several module it might increase binary size.

>At least the debug builds are seem to be almost exactly the same size, with or without coroutines
I guess because we already have `coroutines` in some module

paulrenn67: In order to avoid adding coroutine dependency, we could use something like this instead:

```
ExecutorService executor = Executors.newSingleThreadExecutor();
Handler handler = new Handler(Looper.getMainLooper());

executor.execute(new Runnable() {
    @Override
    public void run() {
        ....
    }
}

Comment (paulrenn67) [reply sdk/build.gradle.kts:76]: @yunikkk @alexander-kulikovskii would you be happier with `executor`?

Comment (paulrenn67) [reply sdk/build.gradle.kts:76]: or even
```
new Thread(new Runnable() {
    @Override
    public void run() {
        ....
    }
}

jush: I don't think `executor` is needed. You can [post](https://developer.android.com/reference/android/os/Handler#post(java.lang.Runnable)) to the `handler` directly.

jush: IIRC, your coroutine solution was just enqueuing the _code_ to be run in the main thread later on. No need for a bg thread in the Handler solution either if you don't want to explicitly run in a bg thread.

paulrenn67: No, we don't want/need it on the main thread, we are just sending an event for turnstile so the user can be billed, nothing to do in the UI.

jush: Ok. Then, you could use the kotlin extension function https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.concurrent/thread.html:
```Kotlin
thread {
    mapTelemetry.onAppUserTurnstileEvent()
}
```

Sidenote: is there already any other bg thread we could re-use instead of creating a new one?

jush (approved):

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -50,7 +52,10 @@ internal object MapProvider {
         paramsProvider(context, accessToken, MapboxModuleType.MapTelemetry)
       }
     }
-    mapTelemetry.onAppUserTurnstileEvent()
+    // Report turnstile event in background thread.
+    thread {
+      mapTelemetry.onAppUserTurnstileEvent()
```
alexander-kulikovskii, line 57: just curious. is it possible, that `mapTelemetry` might be not initialized before this thread is started? Maybe thread execute `onAppUserTurnstileEvent` before line 51

sdk/build.gradle.kts
yunikkk: @paulrenn67 I'm happy with either thread/executor. 
What I would also suggest is to file a ticket in common to dispatch this call to the BG instead of the client (then our other libs using it will also benefit from the speedup)

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
paulrenn67: `MapboxModuleProvider.createModule` is synchronous so `mapTelemetry` should be valid when we hit the `mapTelemetry.onAppUserTurnstileEvent() `call.

alexander-kulikovskii: but what if compiler decide to change order of calls and block `thread { }` will runs first?

alexander-kulikovskii: and from documentation `thread { }` block has a property `start` 

`if true, the thread is immediately started` 

so it might be a little chance that thread will start with non initialized function

yunikkk: > but what if compiler decide to change order of calls and block thread { } will runs first?

I assume newly started thread is guaranteed to see the latest values of all the variables that were set by the thread that have started it, e.g. from [docs](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility) - `A call to start on a thread happens-before any action in the started thread.`

sdk/build.gradle.kts
jush: btw, Kotlin coroutines seems to add around 100kB post-shrinking. (Seen in [AndroidX API Guidelines](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:docs/api_guidelines/dependencies.md;bpv=0))

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
paulrenn67: Have checked with http interceptor, and event was successfully sent in both foreground and background threads.
Not sure @alexander-kulikovskii why the compiler would change the order of calls? Do you mean the thread executes before `mapTelemetry` is created? Can you give a code sample?

yunikkk: @paulrenn67 in general compiler might apply various optimizations and reorder instructions (e.g. [docs](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4-A)), but this should no be a problem here since [docs](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility) also say that `A call to start on a thread happens-before any action in the started thread.`

sdk/src/main/java/com/mapbox/maps/MapProvider.kt
```diff
@@ -50,7 +52,10 @@ internal object MapProvider {
         paramsProvider(context, accessToken, MapboxModuleType.MapTelemetry)
       }
     }
-    mapTelemetry.onAppUserTurnstileEvent()
+    // Report turnstile event in background thread.
+    thread {
```
yunikkk, line 56: Will it create a new thread on every map instantiation?

yunikkk: I'd probably cached created thread together with `mapTelemetry`

pengdev: @paulrenn67 can we validate the message is sent correctly? Most of our native APIs requires to be ran on main thread only. 

cc @jush do you know the thready safety for the `EventsService.sendTurnstileEvent` API?

paulrenn67: @jush I checked with Tanja and also checked with added http logger, seemed to work alright.
@yunikkk I don't think it is worth storing the thread because most apps only contain a single `MapView`, and once the event has been sent, we don't need the thread anymore.

jush: Unfortunately, I'm not familiar enough with the native side to say for sure but it seems to enqueue it immediately to a worker...

Which makes me wonder why this call is expensive in the first place...

paulrenn67: Perhaps it is the lazy initialisation of `mapTelemetry = MapboxModuleProvider.createModule` that costs time, the first time through.  In which case we could maybe use the thread technique only there.

jush: looking at the Jira ticket Peng mentions that it's due to JSON conversion

paulrenn67: JSON is very small in this case as far as I have seen

PR #1949, opened 2023-01-17
[MAPSAND-679] wrap longitude to world bound for flyTo animation.

ank27:
### Summary of changes
Current `flyTo()` animation doesn't wrap longitude values to world bounds. which results in invalid tiles in native. This pr fixes it by wraping longitude values as per native implementation.

## Pull request checklist:
    
### DIFF
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
@@ -3,6 +3,7 @@ package com.mapbox.maps.plugin.animation
 import androidx.annotation.VisibleForTesting
 import com.mapbox.geojson.Point
 import com.mapbox.maps.*
+import com.mapbox.maps.plugin.animation.CameraTransform.ge
 import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
 import com.mapbox.maps.plugin.delegates.MapTransformDelegate
 import kotlin.math.*
@@ -65,23 +66,33 @@ internal object CameraTransform {
   fun calculateScaleBy(amount: Double, currentZoom: Double) = log2(amount) + currentZoom
 
   /**
-   * Extension function to wrap longitude to one world copy.
+   * Extension function to wrap coordinate to one world copy.
    */
   fun Point.wrapCoordinate(): Point {
-    val lng = wrap(this.longitude(), -LONGITUDE_MAX.toDouble(), LONGITUDE_MAX.toDouble())
-    return Point.fromLngLat(lng, this.latitude())
+    val lng = wrap(
+      value = this.longitude(),
+      min = -LONGITUDE_MAX.toDouble(),
+      max = LONGITUDE_MAX.toDouble()
+    )
+    return if (lng.isNaN()) Point.fromLngLat(this.longitude(), this.latitude())
+    else Point.fromLngLat(lng, this.latitude())
   }
 
   // Wrap value to the given range (including min, excluding max).
   @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
   internal fun wrap(value: Double, min: Double, max: Double): Double {
-    if (value >= min && value < max) {
-      return value
-    } else if (value == max) {
+    if (value.eq(max)) {
       return min
+    } else if (value.ge(min) && value.less(max)) {
+      return value
     }
     val delta = max - min
     val wrapped = min + ((value - min) % delta)
     return if (value < min) wrapped + delta else wrapped
   }
+
+  private const val PRECISION = 1e-6
+  private fun Double.eq(other: Double) = abs(this - other) < PRECISION
+  private fun Double.ge(other: Double) = (this - other) > PRECISION || this.eq(other)
+  private fun Double.less(other: Double) = (this - other) < -PRECISION
 }
--- a/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
+++ b/plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
@@ -1,5 +1,6 @@
 package com.mapbox.maps.plugin.animation
 
+import androidx.annotation.VisibleForTesting
 import com.mapbox.geojson.Point
 import com.mapbox.maps.*
 import com.mapbox.maps.plugin.delegates.MapCameraManagerDelegate
@@ -62,4 +63,25 @@ internal object CameraTransform {
   }
 
   fun calculateScaleBy(amount: Double, currentZoom: Double) = log2(amount) + currentZoom
+
+  /**
+   * Extension function to wrap longitude to one world copy.
+   */
+  fun Point.wrapCoordinate(): Point {
+    val lng = wrap(this.longitude(), -LONGITUDE_MAX.toDouble(), LONGITUDE_MAX.toDouble())
+    return Point.fromLngLat(lng, this.latitude())
+  }
+
+  // Wrap value to the given range (including min, excluding max).
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun wrap(value: Double, min: Double, max: Double): Double {
+    if (value >= min && value < max) {
+      return value
+    } else if (value == max) {
+      return min
+    }
+    val delta = max - min
+    val wrapped = min + ((value - min) % delta)
+    return if (value < min) wrapped + delta else wrapped
+  }
 }
--- a/plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactoryTest.kt
+++ b/plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactoryTest.kt
@@ -222,18 +222,22 @@ class CameraAnimatorsFactoryTest {
 }
 
 @RunWith(Parameterized::class)
-class WrapCoordinateTest(private val actual: Double, private val expected: Double) {
+class WrapCoordinateTest(
+  private val actual: Triple<Double, Double, Double>,
+  private val expected: Double
+) {
 
   private companion object {
     @JvmStatic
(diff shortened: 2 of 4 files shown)


plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,25 @@ internal object CameraTransform {
   }
 
   fun calculateScaleBy(amount: Double, currentZoom: Double) = log2(amount) + currentZoom
+
+  /**
+   * Extension function to wrap longitude to one world copy.
+   */
+  fun Point.wrapCoordinate(): Point {
```
alexander-kulikovskii, line 70:
```suggestion
  fun Point.wrapLongitude(): Point {
```

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,25 @@ internal object CameraTransform {
...
+    val lng = wrap(this.longitude(), -LONGITUDE_MAX.toDouble(), LONGITUDE_MAX.toDouble())
```
alexander-kulikovskii, line 71:
```suggestion
    val lng = wrap(value = this.longitude(), min = -LONGITUDE_MAX.toDouble(), max = LONGITUDE_MAX.toDouble())
```

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,25 @@ internal object CameraTransform {
...
+    return Point.fromLngLat(lng, this.latitude())
+  }
+
+  // Wrap value to the given range (including min, excluding max).
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun wrap(value: Double, min: Double, max: Double): Double {
+    if (value >= min && value < max) {
```
alexander-kulikovskii, line 78: should we compare all values with some eps? 
`1e-6` for example

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactoryTest.kt
```diff
@@ -218,4 +219,29 @@ class CameraAnimatorsFactoryTest {
...
+}
+
+@RunWith(Parameterized::class)
+class WrapCoordinateTest(private val actual: Double, private val expected: Double) {
+
+  private companion object {
+    @JvmStatic
+    @Parameterized.Parameters
+    fun data() = listOf(
+      arrayOf(CameraTransform.wrap(-0.911, -LONGITUDE_MAX, LONGITUDE_MAX), -0.911),
```
alexander-kulikovskii, line 231: nit:

you can simplify call, store only data and in test function call `CameraTransform.wrap`
for spliting actual and expected can use `Triple(-0.911, -LONGITUDE_MAX, LONGITUDE_MAX)`, `-0.911`

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,25 @@ internal object CameraTransform {
...
+      return value
+    } else if (value == max) {
```
paulrenn67, line 80: We should not use equals for Double equality testing. Maybe we need to convert to `BigDecimal` or some other solution.

ank27: good point. i will update that to use some precision. that will be better than `BigDecimal` i think.

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
ank27: yes adding that :)

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,34 @@ internal object CameraTransform {
...
+      max = LONGITUDE_MAX.toDouble()
+    )
+    return if (lng.isNaN()) Point.fromLngLat(this.longitude(), this.latitude())
+    else Point.fromLngLat(lng, this.latitude())
+  }
+
+  // Wrap value to the given range (including min, excluding max).
+  @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+  internal fun wrap(value: Double, min: Double, max: Double): Double {
+    if (value.ge(min) && value < max) {
```
alexander-kulikovskii, line 83:
```suggestion
    if (value.ge(min) && value.less(max)) {
```

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,34 @@ internal object CameraTransform {
...
+      return value
+    } else if (value.eq(max)) {
+      return min
+    }
+    val delta = max - min
+    val wrapped = min + ((value - min) % delta)
+    return if (value < min) wrapped + delta else wrapped
```
alexander-kulikovskii, line 90:
```suggestion
    return if (value.less(min)) wrapped + delta else wrapped
```

plugin-animation/src/test/java/com/mapbox/maps/plugin/animation/CameraAnimatorsFactoryTest.kt
```diff
@@ -21,11 +21,13 @@ import org.junit.Assert
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.LooperMode
 import org.robolectric.shadows.ShadowLog
 import org.robolectric.shadows.ShadowLooper.shadowMainLooper
 import java.time.Duration
+import kotlin.math.ln
```
alexander-kulikovskii, line 30: rm?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +63,35 @@ internal object CameraTransform {
...
+  }
+
+  private const val PRECISION = 1e-6
+  private fun Double.eq(other: Double) = abs(this - other) < PRECISION
+  private fun Double.ge(other: Double) = this > other || this.eq(other)
```
alexander-kulikovskii, line 95: should it be also with eps?

```suggestion
  private fun Double.ge(other: Double) = this + PRECISION > other || this.eq(other)
```

or 

```suggestion
  private fun Double.ge(other: Double) = (this - other) > PRECISION || this.eq(other)
```

same question for `less` function

alexander-kulikovskii (approved):

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +64,35 @@ internal object CameraTransform {
...
+      return value
+    }
+    val delta = max - min
+    val wrapped = min + ((value - min) % delta)
+    return if (value < min) wrapped + delta else wrapped
+  }
+
+  private const val PRECISION = 1e-6
+  private fun Double.eq(other: Double) = abs(this - other) < PRECISION
+  private fun Double.ge(other: Double) = (this - other) > PRECISION || this.eq(other)
```
yunikkk, line 96: Should it be plain `this >= other`? Precision is solely for equality comparisons isn't it?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +64,35 @@ internal object CameraTransform {
...
+  private fun Double.less(other: Double) = (this - other) < PRECISION
```
yunikkk, line 97: Also here - wouldn't it fail for `2.0.less(3.0)` ?

ank27: no, it doesn't fail for that.

yunikkk: Discussed in slack

pengdev (approved):
LGTM, could you also attach videos to showcase the flickering of the symbol is resolved?

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
```diff
@@ -62,4 +64,35 @@ internal object CameraTransform {
...
+  /**
+   * Extension function to wrap coordinate to one world copy.
+   */
+  fun Point.wrapCoordinate(): Point {
+    val lng = wrap(
+      value = this.longitude(),
+      min = -LONGITUDE_MAX.toDouble(),
+      max = LONGITUDE_MAX.toDouble()
+    )
+    return if (lng.isNaN()) Point.fromLngLat(this.longitude(), this.latitude())
```
yunikkk, line 77: Wonder in which case it can be NaN. Should `wrap(` just always return correct value?

yunikkk (approved):
LGTM, one Q

plugin-animation/src/main/java/com/mapbox/maps/plugin/animation/CameraTransform.kt
ank27: `wrap` could return NaN if arguments `min` and `max` are same, although that's not the case in our code. It's just an extra check so `wrapCoordinate` doesn't return NaN lat/lng.

PR #1975, opened 2023-01-27
MAPSAND-683 Make Source constructors private

paulrenn67:
### Summary of changes

Source constructor visibility changed to private, users should use Builder class instead.

### User impact (optional)

Usage of e.g. GeoJsonSource(builder) should be changed to GeoJsonSource.Builder().build()

## Pull request checklist:
    
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+* Source constructor visibility changed to private, users should use Builder class instead. ([1975](https://github.com/mapbox/mapbox-maps-android/pull/1975))
 
 
 # 10.11.0-rc.1 January 26, 2023
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -33,20 +33,13 @@ import com.mapbox.maps.logW
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource : Source {
-  private constructor(builder: Builder) : super(builder.sourceId) {
-    sourceProperties.putAll(builder.properties)
-    volatileSourceProperties.putAll(builder.volatileProperties)
-    initGeoJson = builder.geoJson
-    initData = builder.data
-  }
+class GeoJsonSource private constructor(builder: Builder): Source(builder.sourceId) { 
+  private var initGeoJson: GeoJson? = builder.geoJson
+  private var initData: String? = builder.data
 
   private val workerHandler by lazy {
     Handler(workerThread.looper)
   }
-
-  private var initGeoJson: GeoJson? = null
-  private var initData: String? = null
 
   private fun setGeoJson(geoJson: GeoJson) {
     workerHandler.removeCallbacksAndMessages(null)
@@ -81,6 +74,11 @@ class GeoJsonSource : Source {
       setData(it)
       initData = null
     }
+  }
+
+  init {
+    sourceProperties.putAll(builder.properties)
+    volatileSourceProperties.putAll(builder.volatileProperties)
   }
 
   /**
@@ -403,12 +401,6 @@ class GeoJsonSource : Source {
     // Properties that only settable after the source is added to the style.
     internal val volatileProperties = HashMap<String, PropertyValue<*>>()
 
-    init {
-      // set default data to allow empty data source.
-      val propertyValue = PropertyValue("data", TypeUtils.wrapToValue(""))
-      properties[propertyValue.propertyName] = propertyValue
-    }
-
     /**
      * A URL to a GeoJSON file, or inline GeoJSON.
      */
@@ -645,6 +637,9 @@ class GeoJsonSource : Source {
      * @return the GeoJsonSource
      */
     fun build(): GeoJsonSource {
+      // set default data to allow empty data source.
+      val propertyValue = PropertyValue("data", TypeUtils.wrapToValue(""))
+      properties[propertyValue.propertyName] = propertyValue
       return GeoJsonSource(this)
     }
   }
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -34,13 +34,14 @@ import com.mapbox.maps.logW
  *
  */
 class GeoJsonSource : Source {
-  @Deprecated("Use builder instead", level = DeprecationLevel.WARNING)
+  @Deprecated("Use builder instead", level = DeprecationLevel.ERROR)
   constructor(builder: Builder) : super(builder.sourceId) {
-      sourceProperties.putAll(builder.properties)
-      volatileSourceProperties.putAll(builder.volatileProperties)
+    sourceProperties.putAll(builder.properties)
+    volatileSourceProperties.putAll(builder.volatileProperties)
     initGeoJson = builder.geoJson
     initData = builder.data
   }
+
   private var initGeoJson: GeoJson?
   private var initData: String?
 
@@ -638,7 +639,7 @@ class GeoJsonSource : Source {
      *
      * @return the GeoJsonSource
      */
-    @SuppressWarnings("deprecation")
+    @Suppress("DEPRECATION_ERROR")
     fun build(): GeoJsonSource {
       // set default data to allow empty data source.
       val propertyValue = PropertyValue("data", TypeUtils.wrapToValue(""))
--- a/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
+++ b/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
@@ -4,7 +4,6 @@ package com.mapbox.maps.extension.style.sources.generated
 
 import android.os.Handler
 import android.os.HandlerThread
-import android.os.Looper
 import android.os.Process.THREAD_PRIORITY_DEFAULT
 import androidx.annotation.VisibleForTesting
 import com.mapbox.bindgen.Value
@@ -34,22 +33,20 @@ import com.mapbox.maps.logW
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+class GeoJsonSource : Source {
+  private constructor(builder: Builder) : super(builder.sourceId) {
+    sourceProperties.putAll(builder.properties)
+    volatileSourceProperties.putAll(builder.volatileProperties)
+    initGeoJson = builder.geoJson
+    initData = builder.data
+  }
+
   private val workerHandler by lazy {
     Handler(workerThread.looper)
   }
 
   private var initGeoJson: GeoJson? = null
   private var initData: String? = null
-
-  private constructor(
-    builder: Builder,
-    geoJson: GeoJson?,
-    data: String?,
-  ) : this(builder) {
-    this.initGeoJson = geoJson
-    this.initData = data
-  }
 
   private fun setGeoJson(geoJson: GeoJson) {
     workerHandler.removeCallbacksAndMessages(null)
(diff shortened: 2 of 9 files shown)


ank27:
make sure to add tests and changelog :)

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -34,23 +33,14 @@ import com.mapbox.maps.logW
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
+class GeoJsonSource private constructor(builder: Builder) : Source(builder.sourceId) {
```
yunikkk, line 36: I'm having second thoughts about other Sources than GeoJsonSource. While for this one it's definitely to call `GeoJsonSource(builder)` on the user end other Sources just return `Source(builder)` in their `Builder.build()` definition, e.g. https://github.com/mapbox/mapbox-maps-android/blob/f6bd9832e16405742af74faacfb7f817a9ada03c/extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/ImageSource.kt#L140. 
I can imagine there might be a small change someone is using those constructors. To reduce risk of unneeded runtime crashes could we maybe only make private constructor for the GeoJsonSource? And just deprecate rest of Sources?

paulrenn67: I think all the Souce classes should have a consistent API

yunikkk: I see your point about consistency. At the same time it still concerns me that we introduce potential crash for the users who might use sources with perfectly working constructors of Image/Raster/RasterDem/Vector sources. While for GeoJsonSource this constructor is buggy, other sources are fine.

yunikkk: Actually I've just got another idea - use Deprecated annotation with level ERROR - this way it'll remain binary but not  source compatible, thus forcing to use another one @paulrenn67

ank27: I think it make sense to deprecate and use another one, so we don't break user's code. 
Or we can treat it as a breaking change?

paulrenn67: Hmmm ok, we can go with the Deprecation approach for now, and remove in v11

paulrenn67: @yunikkk @ank27 could you please review again, constructors now made `Deprecated`. Had to use `DeprecationLevel.WARNING` though, otherwise our own `Builder.build `gets compile error.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 # main
+* Source constructor visibility changed to private, users should use Builder class instead. ([1975](https://github.com/mapbox/mapbox-maps-android/pull/1975))
```
ank27, line 6:
```suggestion
* Deprecate `Source` constructor invocation. Use `Builder` class instead. ([1975](https://github.com/mapbox/mapbox-maps-android/pull/1975))
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -34,21 +33,19 @@ import com.mapbox.maps.logW
  * @see [The online documentation](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson)
  *
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
-  private val workerHandler by lazy {
-    Handler(workerThread.looper)
+class GeoJsonSource : Source {
+  @Deprecated("Use builder instead", level = DeprecationLevel.WARNING)
```
ank27, line 37:
```suggestion
  @Deprecated("This method is deprecated. Use Builder instead", level = DeprecationLevel.WARNING)
```

ank27:
Looks good in general. can you check the generated code, looks like there are some indentation issues

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -649,11 +638,12 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
      *
      * @return the GeoJsonSource
      */
+    @SuppressWarnings("deprecation")
```
ank27, line 641: No need to `SuppressWarning` imo.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -34,21 +33,19 @@ import com.mapbox.maps.logW
...
  */
-class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
-  private val workerHandler by lazy {
-    Handler(workerThread.looper)
+class GeoJsonSource : Source {
+  @Deprecated("Use builder instead", level = DeprecationLevel.WARNING)
+  constructor(builder: Builder) : super(builder.sourceId) {
+      sourceProperties.putAll(builder.properties)
+      volatileSourceProperties.putAll(builder.volatileProperties)
+    initGeoJson = builder.geoJson
```
alexander-kulikovskii, line 41: smth with tabs/spaces

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
yunikkk, line 37: I'd suggest to use level Error to force users to use Builder. It'll still be source-incompatible change, breaking compilation but not runtime behaviour.

yunikkk, line 37: E.g.

```suggestion
  @Deprecated("Use builder instead", level = DeprecationLevel.ERRROR)
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
pengdev: I think we might overcomplicated things here, I'd consider it as a bug fix, using constructor with Builder inside it is clearly a misuse to me, and user should fix their code if they were using it that way. wdyt @yunikkk @ank27 @paulrenn67 ?

yunikkk: @pengdev it's still better to break only source compatibility in my opinion. Otherwise it's binary breaking and if maps minor version is updated - they'll get a crash. 
At the same time clearly wrong behaviour would be to create GeoJsonSource using constructor with builder, other sources may be perfectly fine built using it.

yunikkk: > Had to use DeprecationLevel.WARNING though, otherwise our own Builder.build gets compile error.

@paulrenn67 could you check with  `@Suppress("DEPRECATION_ERROR")` ? Seems it can ignore even the error

paulrenn67: Ah maybe DEPRECATION_ERROR will do the trick.

pengdev: @paulrenn67 I don't think it's worth the hassle, as we are going to remove the private constructor in v11 (and we don't recommend to use constructor in v10 from the beginning).
Especially these classes are generated and is hard to maintain separately for GeoJsonSource from other sources.
I'm fine with deprecation if you think it's necessary, please avoid complex logic as we are going to v11 soon.

paulrenn67: @pengdev the issue is that @yunikkk raised the question of whether we should change anything at all.  So either we implement the ticket to make constructors private, or we can just discard this PR and mark the ticket as "won't do".  I think "won't do" makes more sense now if we are going to make bigger changes in V11.

yunikkk: > @yunikkk raised the question of whether we should change anything at all.

My preferred solution is to use Deprecated with the Error level, this way we 
- forbid compiling if wrong constructor is used, forcing users to change 
- avoid crashes if maps version was bumped implicitly
The only problem could be that compiler won't allow ourselves to use constructor but if that's hacked with `@Suppress("DEPRECATION_ERROR")` I think there's no more flaws here?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/ImageSource.kt
```diff
@@ -17,12 +17,6 @@ import java.util.*
  *
  */
 class ImageSource(builder: Builder) : Source(builder.sourceId) {
-
-  init {
-    sourceProperties.putAll(builder.properties)
-    volatileSourceProperties.putAll(builder.volatileProperties)
-  }
```
yunikkk, line 24: I guess that wasn't intentional?

paulrenn67: Hmmm true, that should be retained for Sources that are _not_ GeoJson. Thanks for noticing.

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
paulrenn67: Constructor changed to 
```
class GeoJsonSource : Source {
  @Deprecated("Use builder instead", level = DeprecationLevel.ERROR)
  constructor(builder: Builder) : super(builder.sourceId) {
```

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
paulrenn67: Without `@Suppress("DEPRECATION_ERROR") `we get `compiler error Using 'constructor GeoJsonSource(GeoJsonSource.Builder)' is an error. Use builder instead`

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/RasterDemSource.kt
```diff
@@ -157,7 +156,7 @@ class RasterDemSource(builder: Builder) : Source(builder.sourceId) {
      */
     get() {
       getPropertyValue<String?>("encoding")?.let {
-        return Encoding.valueOf(it.toUpperCase(Locale.US).replace('-', '_'))
```
pengdev, line 160: is this change intended?

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/GeoJsonSource.kt
```diff
@@ -86,11 +84,6 @@ class GeoJsonSource(builder: Builder) : Source(builder.sourceId) {
     }
   }
 
-  init {
-    sourceProperties.putAll(builder.properties)
-    volatileSourceProperties.putAll(builder.volatileProperties)
-  }
-
```
pengdev, line 93: any reason we removed this init block?

pengdev: ok, found it was moved to the constructor,

extension-style/src/main/java/com/mapbox/maps/extension/style/sources/generated/RasterDemSource.kt
paulrenn67: `toUpperCase` is deprecated in favour of `uppercase`

pengdev (approved):
LGTM % remaining open comments.

pengdev:
Closing as we did it in the internal repo.

PR #1977, opened 2023-01-27
Initial project setup for compose extension - MAPSAND-741

pengdev:
### Summary of changes

This PR creates the initial project setup for the Jetpack Compose extension.

## Pull request checklist:
    

jush (approved):

compose-app/src/main/java/com/mapbox/maps/compose/MainActivity.kt
```diff
@@ -3,57 +3,53 @@ package com.mapbox.maps.compose
...
 import com.mapbox.maps.Style
+import com.mapbox.maps.dsl.cameraOptions
+import com.mapbox.maps.extension.compose.MapboxMap
 
-private const val LATITUDE = 60.239
-private const val LONGITUDE = 25.004
-
-class MainActivity : ComponentActivity() {
+/**
+ * Example to showcase usage of MapView
```
yunikkk, line 19:
```suggestion
 * Example to showcase usage of MapboxMap
```
since it's called MapboxMap now?

extension-compose/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog for Mapbox Maps Android Auto Extension SDK
```
yunikkk, line 1:
```suggestion
# Changelog for Mapbox Maps Compose Extension
```

buildSrc/src/main/kotlin/Project.kt
```diff
@@ -97,7 +97,7 @@ object Dependencies {
   const val composeUi = "androidx.compose.ui:ui:${Versions.compose}"
   const val composeMaterial = "androidx.compose.material:material:${Versions.compose}"
   const val composeUiToolingPreview = "androidx.compose.ui:ui-tooling-preview:${Versions.compose}"
-
+  const val composeUiTooling = "androidx.compose.ui:ui-tooling:${Versions.compose}"
```
ank27, line 100: probably better to split compose dependencies to another object inside dependencies. so we can access them like `Dependencies.Componse.**`

compose-app/build.gradle.kts
```diff
@@ -33,6 +33,10 @@ android {
     kotlinCompilerExtensionVersion = Versions.compose
   }
 
+  kotlinOptions {
+    freeCompilerArgs += "-Xexplicit-api=strict"
```
ank27, line 37: what does this flag do ?

ank27 (approved):
one nit. else looks good :)

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Modifier to be applied to the Mapbox map.
+   */
+  modifier: Modifier = Modifier,
+  /**
+   * Defines the initialisation configurations for a [MapboxMap].
+   * It can only be set once and not mutable after the initialisation. Mutating the [MapInitOptions]
+   * during recomposition will result in a [IllegalStateException].
+   */
+  mapInitOptions: MapInitOptions = MapInitOptions(LocalContext.current.applicationContext),
```
yunikkk, line 27: Wonder if there's some way to express that the `MapInitOptions` is immutable and could not be altered and lead to recomposition (@Immutable annotation? Haven't digged into the semantics but naming is promising)

compose-app/build.gradle.kts
pengdev: it enforces us to explicitly specify the access modifier and returned type, so it's easier to identify unintentional exposed APIs.

buildSrc/src/main/kotlin/Project.kt
pengdev: Ideally we will switch to [Bill Of Materials](https://developer.android.com/jetpack/compose/bom/bom) and version catelogs after we bump Kotlin version and AGP version later. For now let's keep it simple :)

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: good point, we should take this discovery as a follow up task separately.

jush: No, `Immutable` is used for sth else:
> Indicates a type where the value of any properties will never change after the object is constructed, and all methods are referentially transparent

(https://medium.com/androiddevelopers/jetpack-compose-stability-explained-79c10db270c8)

extension-compose/build.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+    freeCompilerArgs += "-Xexplicit-api=strict"
+  }
+}
+
+dependencies {
+  dependencies {
+    compileOnly(project(":sdk"))
+    api(Dependencies.googleMaterialDesign)
+    api(Dependencies.composeUi)
+    api(Dependencies.composeMaterial)
```
jush, line 42: Do we really need in `extension-compose` the material deps (`composeMaterial` and `googleMaterialDesign`).

.circleci/config.yml
```diff
@@ -830,6 +843,25 @@ jobs:
                 variant: "Release"
       - run: exit 0
 
+  build-compose-app:
```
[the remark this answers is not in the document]
pengdev: currently there's nothing much to test, we will add more instrumentation/robo tests as we grow the feature.

extension-compose/build.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+  kotlinOptions {
+    freeCompilerArgs += "-Xexplicit-api=strict"
+  }
+}
+
+dependencies {
+  dependencies {
+    compileOnly(project(":sdk"))
+    api(Dependencies.googleMaterialDesign)
+    api(Dependencies.composeUi)
```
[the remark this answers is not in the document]
pengdev: updated and removed unnecessary dependencies.

extension-compose/build.gradle.kts
pengdev: just did a clean up of dependencies, we for now only implement `composeUi` and `composeMaterial`, these are required to build the extension.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * The content of the map.
+   */
+  content: @Composable () -> Unit
```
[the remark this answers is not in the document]
pengdev: We will add annotation to limit the access, e.g. `ComposableTargetMarker` however this annotation is only available  after Compose Version 1.2.0-alpha04.
We will address these questions as follow ups, this is just an initial project setup with minimum code.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
yunikkk: Hmm @jush doesn't `a type where the value of any properties will never change after the object is constructed` means properties could not be changed == could not trigger recomposition?

jush: I understood that you wanted to apply the `@Immutable` to the `MapInitOptions` class.
If I understand it right, in this case the user could still pass a different `MapInitOptions` which would trigger a new pass for the `MapboxMap`.

yunikkk: Ah, probably you are right. Maybe there's some similar way to express this kind of limitations, e.g. limit part of the fields while allowing other to change, imo worth to look into.

jush: Another definition can be found in @JorgeCastilloPrz `Jetpack Compose internals` book:
> This annotation is applied over a class as a strict promise for the compiler about all the publicly accessible class properties and fields remaining unchanged after creation. Note that this is a stronger promise than the language val keyword, since val only ensures that the property can’t be reassigned via setter, but it can point to a mutable data structure for example, making our data mutable even if it only has val properties. That would break the expectations of the Compose runtime. In other words, this annotation is needed by Compose essentially because the Kotlin language does not provide a mechanism (a keyword or something) to ensure when some data structure is immutable.

jush: TBH, I'm not sure it makes sense to restrict it... In my opinion it shouldn't be much different than if we change the drawable on an [Icon](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/Icon.kt;l=55?q=composable%20icon) composable call. 

It might be expensive but that's it, we should support that scenario as well as we can.

yunikkk: Well I'd say we should very clearly indicate that updating those properties will lead to the whole map recreation (which is slow or very slow for complex styles). 

Another thing is - part of the fields should not mean the whole map reload, e.g. `mapOptions`, `styleUri` or `cameraOptions` should be just quick changes without the MapView recreation.

PR #1983, opened 2023-01-31
[compose] Add MapboxMap compose entry point, expose map controller through MapEffect.

pengdev:
### Summary of changes

* Create entry point composable function MapboxMap, take MapInitOptions as parameter.
* Add map click listener hooks as the input to the MapboxMap composable function
* Add MapEffect class, pass through map controller.
* Create a small test application in the compose project.

## Pull request checklist:
    
### DIFF
--- a/extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
+++ b/extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
@@ -1,0 +1,100 @@
+package com.mapbox.maps.extension.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.currentComposer
+import com.mapbox.maps.MapView
+import com.mapbox.maps.MapboxExperimental
+import com.mapbox.maps.extension.compose.internal.MapApplier
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when MapEffect is recomposed with a different key1.
+ *
+ * The coroutine will be cancelled when the MapEffect leaves the composition.
+ *
+ * This function should not be used to (re-)launch ongoing tasks in response to callback events by
+ * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
+ * to obtain a CoroutineScope that may be used to launch ongoing jobs scoped to the composition in
+ * response to event callbacks.
+ */
+@Composable
+@MapboxMapComposable
+@MapboxExperimental
+public fun MapEffect(key1: Any?, block: suspend CoroutineScope.(MapView) -> Unit) {
+  val map = (currentComposer.applier as MapApplier).mapView
+  LaunchedEffect(key1 = key1) {
+    block(map)
+  }
+}
+
+/**
+ * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when MapEffect is recomposed with a different key1
+ * or key2.
+ *
+ * The coroutine will be cancelled when the MapEffect leaves the composition.
+ *
+ * This function should not be used to (re-)launch ongoing tasks in response to callback events by
+ * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
+ * to obtain a CoroutineScope that may be used to launch ongoing jobs scoped to the composition in
+ * response to event callbacks.
+ */
+@Composable
+@MapboxMapComposable
+@MapboxExperimental
+public fun MapEffect(key1: Any?, key2: Any?, block: suspend CoroutineScope.(MapView) -> Unit) {
+  val map = (currentComposer.applier as MapApplier).mapView
+  LaunchedEffect(key1 = key1, key2 = key2) {
+    block(map)
+  }
+}
+
+/**
+ * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when MapEffect is recomposed with a different key1,
+ * key2 or key3.
+ *
+ * The coroutine will be cancelled when the MapEffect leaves the composition.
+ *
+ * This function should not be used to (re-)launch ongoing tasks in response to callback events by
+ * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
+ * to obtain a CoroutineScope that may be used to launch ongoing jobs scoped to the composition in
+ * response to event callbacks.
+ */
+@Composable
+@MapboxMapComposable
+@MapboxExperimental
+public fun MapEffect(
+  key1: Any?,
+  key2: Any?,
+  key3: Any?,
+  block: suspend CoroutineScope.(MapView) -> Unit
+) {
+  val map = (currentComposer.applier as MapApplier).mapView
+  LaunchedEffect(key1 = key1, key2 = key2, key3 = key3) {
+    block(map)
+  }
+}
+
+/**
+ * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when MapEffect is recomposed with any different keys.
+ *
+ * The coroutine will be cancelled when the MapEffect leaves the composition.
+ *
+ * This function should not be used to (re-)launch ongoing tasks in response to callback events by
+ * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
+ * to obtain a CoroutineScope that may be used to launch ongoing jobs scoped to the composition in
+ * response to event callbacks.
+ */
+@Composable
+@MapboxMapComposable
+@MapboxExperimental
+public fun MapEffect(vararg keys: Any?, block: suspend CoroutineScope.(MapView) -> Unit) {
+  val map = (currentComposer.applier as MapApplier).mapView
+  LaunchedEffect(keys = keys) {
+    block(map)
+  }
+}
--- a/extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
+++ b/extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
@@ -9,10 +9,10 @@ import com.mapbox.maps.extension.compose.internal.MapApplier
 import kotlinx.coroutines.CoroutineScope
 
 /**
- * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
- * The coroutine will be cancelled and re-launched when MapEffect is recomposed with a different key1.
+ * When MapEffect enters the composition it will launch [block] into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when [MapEffect] is recomposed with a different [key1].
  *
- * The coroutine will be cancelled when the MapEffect leaves the composition.
+ * The coroutine will be cancelled when the [MapEffect] leaves the composition.
  *
  * This function should not be used to (re-)launch ongoing tasks in response to callback events by
  * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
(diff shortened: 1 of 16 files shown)


jush (approved):
LGTM with couple of nits

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -23,14 +45,23 @@ public fun MapboxMap(
   modifier: Modifier = Modifier,
   /**
    * Defines the initialisation configurations for a [MapboxMap].
+   *
    * It can only be set once and not mutable after the initialisation. Mutating the [MapInitOptions]
    * during recomposition will result in a [IllegalStateException].
    */
-  mapInitOptions: MapInitOptions = MapInitOptions(LocalContext.current.applicationContext),
+  mapInitOptions: MapInitOptions = MapInitOptions(LocalContext.current),
```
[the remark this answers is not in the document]
pengdev: reverted this change

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.mapbox.maps.MapView
+import com.mapbox.maps.MapboxExperimental
+import com.mapbox.maps.extension.compose.internal.MapApplier
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * When MapEffect enters the composition it will launch block into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when MapEffect is recomposed with a different key1.
+ *
+ * The coroutine will be cancelled when the MapEffect leaves the composition.
```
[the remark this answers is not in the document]
pengdev: updated

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -23,18 +45,27 @@ public fun MapboxMap(
...
    * Defines the initialisation configurations for a [MapboxMap].
+   *
    * It can only be set once and not mutable after the initialisation. Mutating the [MapInitOptions]
    * during recomposition will result in a [IllegalStateException].
    */
   mapInitOptions: MapInitOptions = MapInitOptions(LocalContext.current.applicationContext),
+  /**
+   * Callback to be invoked when the user clicks on the map view.
+   */
+  onMapClickListener: OnMapClickListener = OnMapClickListener { false },
```
jush, line 56: Is it clear for the developers that to ignore clicks they need to set a listener that returns `false`?

Would it be easier to understand if we allow nullable listeners?

jush: We would also save some microseconds from `GesturesPluginImpl.handleClickEvent` since the listener would be empty 😝

pengdev: yeah, makes sense, we can make it default to null.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -47,4 +101,65 @@ public fun MapboxMap(
...
+              this.longClickListener = listener
+            }
+            update(mapInitOptions) {
+              throw IllegalStateException(
+                """
+                Mutating MapInitOptions during composition is not allowed.
+                """.trimIndent()
+              )
+            }
+          }
```
pengdev, line 154: @yunikkk @jush continuing the discussion here, for now it just throws when the `MapInitOption` is updated.

Maybe we can also explore other ways to configure it, but allowing to recreate the MapView is too heavy and requires more architecture changes, wdyt?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param from original position
+   * @param to the target position
+   */
+  fun onMoved(from: Int, to: Int) {}
+
+  /**
+   * Invoked when the [MapNode] is removed from the MapboxMap node tree.
+   */
+  fun onRemoved() {}
```
alexander-kulikovskii, line 27:
```suggestion
  fun onDetached() {}
```
because we have `attached`

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
```diff
@@ -0,0 +1,83 @@
...
+  private val decorations = mutableListOf<MapNode>()
+
+  override fun onClear() {
+    decorations.forEach { it.onClear() }
+    decorations.clear()
+  }
+
+  override fun insertBottomUp(index: Int, instance: MapNode) {
+    decorations.add(index, instance)
+    instance.onAttached()
```
alexander-kulikovskii, line 59: should we notify another mapNodes about such kind of adding? maybe another mapNode will be fully hidden?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  override fun move(from: Int, to: Int, count: Int) {
+    // TODO: Notify the nodes that has been moved
+    decorations.move(from, to, count)
+  }
+
+  override fun remove(index: Int, count: Int) {
+    repeat(count) { offset ->
+      decorations[index + offset].onRemoved()
```
alexander-kulikovskii, line 79: should we check out of index?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -47,4 +101,65 @@ public fun MapboxMap(
...
+          factory = {
+            MapboxMapNode(
+              mapView,
+              onMapClickListener,
+              onMapLongClickListener
+            )
+          },
+          update = {
+            // input arguments updater
+            update(onMapClickListener) { listener ->
```
alexander-kulikovskii, line 123: nit:
maybe move to separate small functions? will be easy to read and understand code

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -10,6 +9,52 @@ import com.mapbox.maps.MapInitOptions
...
+import com.mapbox.maps.plugin.gestures.OnMapClickListener
+import com.mapbox.maps.plugin.gestures.OnMapLongClickListener
+import com.mapbox.maps.plugin.gestures.gestures
+import kotlinx.coroutines.awaitCancellation
+
+/**
+ * MapboxMapNode to observe/update the input arguments of MapboxMap.
+ */
+private class MapboxMapNode(
+  val controller: MapView,
```
alexander-kulikovskii, line 24:
```suggestion
  internal val controller: MapView,
```

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -10,6 +9,52 @@ import com.mapbox.maps.MapInitOptions
...
+  var clickListener: OnMapClickListener?,
```
alexander-kulikovskii, line 25:
```suggestion
  internal var clickListener: OnMapClickListener? = {},
```

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
```diff
@@ -10,6 +9,52 @@ import com.mapbox.maps.MapInitOptions
...
+
+  override fun onClear() {
+    cleanUp()
+  }
+
+  private fun cleanUp() {
+    controller.gestures.apply {
+      clickListener?.let {
+        removeOnMapClickListener(it)
+      }
```
alexander-kulikovskii, line 51: maybe set listener to `null`?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: The MapboxMapNode is a private class, so I think it's not necessary to add internal modifier to its properties.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: same here, as this is a private class, we don't need to add internal, and I intentionally didn't add default value, as we specifically pass in the click listener during construction.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: the MapboxMapNode#clickListener is set to listener below, whether it's null or not.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
pengdev: it's following the same name pattern as the abstract class, node can be removed from the node tree, and it's not guaranteed it could be reattached to the node tree.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
pengdev: other mapNode are independent on the newly added node
Newly added node will be placed on the top of existing nodes.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
pengdev: in theory we shouldn't get out of index, as the AbstractApplier will maintain a similar nodes structure.

alexander-kulikovskii: but potentially it's possible. maybe add some checks?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
alexander-kulikovskii: it's private for now. maybe in the future you will change it. so you already have proper modifiers
but it's up to you

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
alexander-kulikovskii: and what about default listener?

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * When MapEffect enters the composition it will launch [block] into the composition's CoroutineContext.
+ * The coroutine will be cancelled and re-launched when [MapEffect] is recomposed with a different [key1].
+ *
+ * The coroutine will be cancelled when the [MapEffect] leaves the composition.
+ *
+ * This function should not be used to (re-)launch ongoing tasks in response to callback events by
+ * way of storing callback data in MutableState passed to key. Instead, see rememberCoroutineScope
```
ank27, line 18: nit: maybe a link to `see rememberCoroutineScope`?

ank27 (approved):

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: by default there shouldn't be a listener with nothing done on a click event. Also it will introduce some overhead to pass through the event. It's changed to null by default with discussions in [this](https://github.com/mapbox/mapbox-maps-android/pull/1983#discussion_r1093232950) thread

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: refactored

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
pengdev: Imo composable nodes are internal details and shouldn't be public anyway, I've also moved this definition to internal package.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapEffect.kt
pengdev: rememberCoroutineScope is well defined in the coroutine package and is not used in this file, thus links won't be resolved properly, will leave it as is.

extension-compose/src/main/java/com/mapbox/maps/extension/compose/internal/MapApplier.kt
pengdev: on the other hand, I think we might need a crash here with the stack trace if there's such a case, I don't think we can safely ignore this error.

alexander-kulikovskii (approved):

extension-compose/src/main/java/com/mapbox/maps/extension/compose/MapboxMap.kt
yunikkk: Agree, lets iterate, should be fine for now

PR #1988, opened 2023-02-01
[script] treat develop as a target branch for revapi

tobrun:
Treat `develop` branch as another iteration of target branch next to main.

### DIFF
--- a/scripts/java-api-check-all.sh
+++ b/scripts/java-api-check-all.sh
@@ -40,7 +40,7 @@ echo "CURRENT_BRANCH: $CURRENT_BRANCH"
 echo "TARGET_BRANCH: $TARGET_BRANCH"
 readonly STABLE_RELEASE_TAG_PATTERN="v10\.[0-9]*\.0$"
 
-if [ "$TARGET_BRANCH" = "main" ]; then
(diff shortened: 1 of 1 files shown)


ank27 (approved):

pengdev (approved):

scripts/java-api-check-all.sh
```diff
@@ -40,7 +40,7 @@ echo "CURRENT_BRANCH: $CURRENT_BRANCH"
 echo "TARGET_BRANCH: $TARGET_BRANCH"
 readonly STABLE_RELEASE_TAG_PATTERN="v10\.[0-9]*\.0$"
 
-if [ "$TARGET_BRANCH" = "main" ]; then
+if [ "$TARGET_BRANCH" = "main" ] || [ "$TARGET_BRANCH" = "develop" ]; then
```
yunikkk, line 43: Not sure why both are needed, could we just replace it when migration will happen?

tobrun: putting together a plan that we would actually need 2 protected branches, so yes we need both long term

yunikkk (approved):

Issue #2344, opened 2024-04-12
[Compose] - AnnotatedLayerFeature not working in ViewAnnotation


shrikanth-zuper:
## Environment

- Android OS version: 14
- Devices affected: All devices
- Maps SDK Version: 11.3.0

## Observed behavior and steps to reproduce

I'm currently displaying a `ViewAnnotation` for a `PointAnnotation.` In order to hide the `ViewAnnotation` when that `PointAnnotation` is clustered, I'm using the `annotatedLayerFeature` to set the layer and feature id.

As seen from sample apps and other documentations, setting `annotatedLayerFeature` should ideally result in a behaviour where `ViewAnnotation` will be displayed only when that marker is visible. This is working as expected in iOS, but in Android it's not working properly. ViewAnnotation is not at all displayed when using annotatedLayerFeature. 

```Kotlin
MapboxMap {
    // Rendering markers (Point Annotations)
    PointAnnotationGroup(
        annotations = markers.invoke().map { marker ->
            (PointAnnotationOptions.fromFeature(
                Feature.fromGeometry(
                    marker.point,
                    null,
                    marker.id
                )
            ) ?: PointAnnotationOptions().withPoint(marker.point.toPoint()))
                .withIconImage(marker.icon)
        },
        annotationConfig = AnnotationConfig(
            layerId = "PointAnnotationLayer",
            annotationSourceOptions = AnnotationSourceOptions(
                clusterOptions = ClusterOptions(
                    textColor = Color.WHITE,
                    textSize = 16.0,
                    colorLevels = listOf(
                        Pair(0, Color.BLACK)
                    )
                )
            )
        )
    )

    // Rendering tool windows (View Annotations)
    markers.invoke().forEach { marker ->
        ViewAnnotation(
            options = viewAnnotationOptions {
                geometry(marker.point.toPoint())
                // This has no effect
                annotatedLayerFeature(
                    layerId = "PointAnnotationLayer"
                ) {
                    featureId(marker.id)
                }
                annotationAnchor {
                    anchor(ViewAnnotationAnchor.TOP)
                    offsetY(160.0)
                }
                allowOverlap(false)
            },
            content = marker.toolWindow
        )
    }
}
```

## Expected behavior

When using `annotatedLayerFeature` option in `ViewAnnotation`, that ViewAnnotation should be displayed when that marker is visible and it should be hidden when the marker is clustered. This is working properly in iOS.

## Notes / preliminary analysis

[...]

PR #2357, opened 2024-04-29
[compose] get annotationManager from PointAnnotationGroup

hiroaki404:
for get featureId to use ViewAnnotation

### Summary of changes

It doesn't work when using the combination of PointAnnotationGroup clustering and ViewAnotation.

When creating ViewAnnotationOptions, you can specify the featureId in annotatedLayerFeature to associate it with PointAnnotation. At this time, featureId is obtained from the Annotation as the return value when calling PointAnnotation.create().

However, when using the Copmose Extension, the PointAnnotationManager is hidden, so the annotation ID cannot be obtained.

As a workaround, you can remember PointAnnotationManager and access annotations property of PointAnnotationManager to get the annotation ID when needed.

Currently, annotation IDs are randomly generated, but another method is to specify an ID and create an annotation. However, I think this method requires a large scale of change.

Please consider the contents.

## Pull request checklist:
    
Fixes: #2344

### DIFF
--- a/extension-compose/src/main/java/com/mapbox/maps/extension/compose/annotation/generated/PointAnnotationGroup.kt
+++ b/extension-compose/src/main/java/com/mapbox/maps/extension/compose/annotation/generated/PointAnnotationGroup.kt
@@ -13,6 +13,7 @@ import com.mapbox.maps.extension.style.layers.properties.generated.*
 import com.mapbox.maps.plugin.annotation.AnnotationConfig
 import com.mapbox.maps.plugin.annotation.annotations
 import com.mapbox.maps.plugin.annotation.generated.PointAnnotation
+import com.mapbox.maps.plugin.annotation.generated.PointAnnotationManager
 import com.mapbox.maps.plugin.annotation.generated.PointAnnotationOptions
 import com.mapbox.maps.plugin.annotation.generated.createPointAnnotationManager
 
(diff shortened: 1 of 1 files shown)


extension-compose/src/main/java/com/mapbox/maps/extension/compose/annotation/generated/PointAnnotationGroup.kt
```diff
@@ -90,6 +92,7 @@ public fun PointAnnotationGroup(
   iconTranslateAnchor: IconTranslateAnchor? = null,
   textTranslate: List<Double>? = null,
   textTranslateAnchor: TextTranslateAnchor? = null,
+  onAnnotationManagerCreated: (PointAnnotationManager) -> Unit = {},
```
pengdev, line 95: Hey @hiroaki404 , thanks for contributing. With the proposed approach, I think you still cannot get the annotation id, as the annotation is supposed to be created by compose and you should not directly call `PointAnnotationManager.create()`. 
I think we'd go with exposing the feature id setter in the AnnotationOptions class, as it's more friendly in compose context and also align with iOS implementation. I have created a internal ticket for this feature.

hiroaki404: > I think we'd go with exposing the feature id setter in the AnnotationOptions class, as it's more friendly in compose context and also align with iOS implementation. I have created a internal ticket for this feature.

If that's possible, I think it's better too. Thank you!

PR #2558, opened 2025-01-10
Basic watchdog that triggers thread dumps on puck jank

jush:
### Summary of changes

An example on how to extend a custom location provider to track puck animations jank.

Every time the puck is animated it starts a watchdog that will create a thread dump if the animation has not run for some milliseconds.
Capturing multiple thread dumps allows to detect if the `main` thread is stuck.

The gist of it is to add listeners to the animator and signal the watchdog (see `LocationComponentAnimationActivity.listeners`). 
If enough time elapses (`Watchdog.TIME_TO_FIRST_TRIGGER`) without the watchdog being rescheduled (`Watchdog.reschedule`), then a thread dump will be written to disk (`root` permission is most likely needed to extract them). It continues to do so every 100 milliseconds (maximum 5 times) if `reschedule` is not yet called.
That way you could study the thread dumps and see what the `main` thread is doing.

NOTE: this is a simple implementation please study the code and adapt it better to your needs.

### DIFF
--- a/app/src/main/java/com/mapbox/maps/testapp/examples/LocationComponentAnimationActivity.kt
+++ b/app/src/main/java/com/mapbox/maps/testapp/examples/LocationComponentAnimationActivity.kt
@@ -1,8 +1,11 @@
 package com.mapbox.maps.testapp.examples
 
+import android.animation.Animator
+import android.animation.ValueAnimator
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
+import android.util.Log
 import androidx.appcompat.app.AppCompatActivity
 import androidx.interpolator.view.animation.FastOutSlowInInterpolator
 import com.mapbox.geojson.Point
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/mapbox/maps/testapp/examples/LocationComponentAnimationActivity.kt
```diff
@@ -88,11 +116,20 @@ class LocationComponentAnimationActivity : AppCompatActivity() {
     override fun registerLocationConsumer(locationConsumer: LocationConsumer) {
       this.locationConsumer = locationConsumer
       emitFakeLocations()
+      Watchdog.enabled = true
```
jush, line 119: Before using the `Watchdog` you must enable it.

app/src/main/java/com/mapbox/maps/testapp/examples/LocationComponentAnimationActivity.kt
```diff
@@ -88,11 +116,20 @@ class LocationComponentAnimationActivity : AppCompatActivity() {
...
+        // Simulate main thread busy for few milliseconds
+        Thread.sleep(150)
+        Log.d("TAG", "emitFakeLocations: Finished blocking main thread")
+      }, 15_000L)
     }
 
     override fun unRegisterLocationConsumer(locationConsumer: LocationConsumer) {
       this.locationConsumer = null
       handler.removeCallbacksAndMessages(null)
+      Watchdog.enabled = false
```
jush, line 132: Remember to disable it to stop its thread and free other resources.

